PR #11, opened 2021-03-04
Adds embed fields model ADR

SHession:
## What does this change?

This PR adds an ADR documenting the decision rationale behind our approach to modelling embed fields. It outlines the various options we considered and presents our currently decided approach (which may be amended later as we begin to implement it).

Future users of this repo will be better able to understand the design of the plugin and understand the decisions made.


twrichards (approved):
LGTM 👍 

Very clear IMO

jonathonherbert (approved):

PR #12, opened 2021-03-10
[housekeeping]: Add /src folder

jonathonherbert:
## What does this change?

Adds a src folder, and moves code into it.

## How to test

`yarn dev` should launch the application as before, and work as expected, displaying an editor you can interact with.

A general sense of tidiness.


SHession (approved):
LGTM. I experienced "A general sense of tidiness."

PR #13, opened 2021-03-10
Add webpack

jonathonherbert:
## What does this change?

Adds webpack, replacing parcel. Webpack is fast, and ... easier to use and maintain than it was. Most importantly, it typechecks Typescript code, and has first-party integration with Cypress, which we intend to add. Parcel [does not](https://github.com/cypress-io/cypress-parcel-preprocessor) 😢 

## How to test

Follow the setup instructions. Does everything behave as expected?

Webpack bundles our code, not parcel, enabling typechecking and Cypress integration.


README.md
```diff
@@ -7,4 +7,4 @@ This Prosemirror plugin adds the ability to add embed nodes to a document.
 This project can be tested locally using yarn: 
 
 `yarn` installs and sets up the project dependencies
-`yarn dev` builds the project locally, watches for file changes, and serves the project locally.
\ No newline at end of file
+`yarn start` builds the project locally, and watches for file changes. Open `dist/index.html` to view the test site.
```
SHession, line 10: This is not currently working, there is nothing in place bundle the HTML.

jonathonherbert: Are you certain? When I drag index.html into Chrome, things render as expected, and a bundle.js is generated in `/dist` – happy to pair, I may well be missing something.

(Aware this is not the usual way of serving these sites – until we have a need for a dev server, I thought just opening the file and refreshing the page when it changes would be enough)

jonathonherbert: Oops, discussed out of band and I've left index.html out of dist. Thanks for spotting! Revised w/ a dev server in any case.

SHession (approved):
Work well from a clean repo!

PR #14, opened 2021-03-10
Add cypress for testing

jonathonherbert:
## What does this change?

Adds Cypress for testing. Cypress will let us run unit and integration tests in a real DOM environment.

## How to test

- Run `yarn test`. Does Cypress appear, and run the tests when prompted?
- Run `yarn test-ci`. Does Cypress run all the tests, and produce the correct output?

There's **one problem** (arguably minor) that I can't figure out, yet – in interactive mode, you must manually refresh the tests, or change a file (the round arrow), to run them first time. I'm not sure why this is – any tips welcome!

It's easy to add unit and integration tests to the project.


.nvmrc
```diff
@@ -0,0 +1 @@
+14.15.4
```
SHession, line 1: Is this needed in a yarn project?

SHession (approved):
It looks like you have accidentally included one of the output mp4 files. 

I also noticed this issue you mentioned in the description, not sure how to fix that.

.nvmrc
jonathonherbert: .nvmrc isn't package-manager specific, but the "engines" field in package.json is, and we could set it, so I've added that! IIUC, it's probably useful to keep .nvmrc where it is, as it makes applying the right version easy w/ `nvm`, `fnm` etc.

jonathonherbert:
Thanks for catching, removed. Yes, me neither – probably best to merge and fix forward.

PR #15, opened 2021-03-11
Switch preact for react

jonathonherbert:
## What does this change?

When this project was created, it wasn't clear which direction we were going with frontend frameworks in tooling that would likely consume it.

Now it's clear that the consensus is React, it makes sense to switch, as the benefits of preact (bundle size) no longer apply.

This has the added benefit of improving compatibility with `emotion`, which may have implications for our integration with e.g. Source.

## How to test

This is a no-op – after revising dependencies with `yarn`, the application should run (`yarn start`)and test (`yarn test`, `yarn test-ci`) as before.

The application runs as before, but uses React.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,20 +9,19 @@
     "test-ci": "cypress run"
   },
   "dependencies": {
-    "babel-preset-preact": "^2.0.0",
+    "@emotion/react": "^11.1.5",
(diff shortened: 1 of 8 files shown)


tjsilver (approved):
Tested `main` and this branch and the application and tests ran as before. 👍

PR #16, opened 2021-03-16
Refactor `mount` types

jonathonherbert:
## What does this change?

A refactor to make the interface between embed rendering code and plugin code easier to understand.

Previously, our `createMount` function, which is plugin specific, knew about the way the mounting code rendered its embed. This meant threading functions and type attributes through plugin code and out into embed code again, resulting in a long, curried function that I found hard to understand.

By separating out what the embed cares about (managing state, rendering the embed, and issuing commands) from what the plugin cares about (mounting the embed, and providing those lifecycle methods), we can arrive at what I hope is a more comprehensible interface.

## How to test

- Is the code more comprehensible?
- Is the intent of the code clearer?
- Do the comments add value?

This is a no-op – the code should run and behave as before.

We're slightly more comfortable working with the plugin when we're working on it, or writing embeds.


tjsilver (approved):
Is the code more comprehensible? Yes!
    Is the intent of the code clearer? Yes!
    Do the comments add value? Yes!
Tested before and after and all good. 💯

PR #17, opened 2021-05-04
Add linting

jonathonherbert:
## What does this change?

Adds [Guardian-standard linting](https://www.npmjs.com/package/@guardian/eslint-config-typescript) to the project.

## How to test

The code should work as before. Running `yarn` and then `yarn lint` in the project directory should pass.

## NB

There is no check to enforce this yet. This should come via a CI step w/ Github Actions.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+node_modules
+dist
+coverage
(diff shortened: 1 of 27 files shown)


rhystmills (approved):
Looks good to me!

Is there a reason we've gone for types having a `TMyType` style naming convention in this repository (rather than just `MyType`)?

jonathonherbert:
Great question! IIRC it's a naming convention from a previous linter (the same pattern is used in prosemirror-typerighter) – but I think it's now inconsistent and no longer enforced, worth cleaning up, I think.

PR #18, opened 2021-05-05
Spruce developer experience

jonathonherbert:
## What does this change?

A quick and minimal spruce of the devX on this project – 

- fixes typings for external libraries, importing declarations where available and adding where they're not
- adds Prosemirror devtools.

## How to test

- CI should pass.
- The Prosemirror developer tools should appear in the bottom right hand side of the editor view, and accurately represent the document state.


src/declarations/prosemirror-example-setup.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module "prosemirror-keymap" {
```
SHession, line 1: As mentioned on my other PR review, this module name doesn't match the file.

SHession (approved):
Small issue, also mentioned on #19. Tested locally, LGTM!

src/declarations/prosemirror-example-setup.d.ts
jonathonherbert: 👍  that's fixed in #19 so I'll merge and fix forward.

PR #21, opened 2021-05-13
Add jest for unit tests

jonathonherbert:
## What does this change?

Cypress is great for integration tests, and I hoped to use it for unit tests, too. But it's slower, and produces a much less intelligible output when e.g. diffing objects.

Jest is quicker and more comprehensible when things go wrong programmatically, so devs are likely to have a better feedback loop writing unit tests in Jest.

This PR adds Jest to the project, and moves the single unit test we have out of Cypress and into Jest's remit. It adds Jest to the CI command, too.

## NB

I've had to reshuffle our linting in this PR as a result of this change, as Jest and Cypress' global type declarations don't play nicely together. As a result, we have nested `tsconfig` and `eslintrc` files in the Cypress folder, to ensure we're compiling and linting in the correct places in both `src` and `cypress`.

## How to test

- Does CI pass?
- Does running  `yarn test` run Jest locally, and run a single unit test?

Developers have an easier time writing unit tests for this project.


src/testHelpers.ts
```diff
@@ -0,0 +1,29 @@
+import { Plugin } from "prosemirror-state";
```
jonathonherbert, line 1: This file was pulling Cypress definitions into a Jest environment, so I've moved it to be within this project. The `index.ts` and its helpers should probably be organised more carefully in another PR.

package.json
```diff
@@ -8,8 +8,9 @@
   "license": "MIT",
   "scripts": {
     "start": "webpack serve --open",
-    "test": "cypress open",
-    "ci": "yarn lint && webpack serve & (wait-on http://localhost:7890 && cypress run)",
+    "test:integration": "cypress open",
+    "test": "jest",
+    "ci": "yarn lint && yarn test && webpack serve & (wait-on http://localhost:7890 && cypress run)",
```
SHession, line 13: Possibly this could be reworked slightly. `Test` could run all the test, and there could be specific integration and unit test scripts.

src/testHelpers.ts
SHession: I agree about the organising, happy to leave it to another PR.

SHession (approved):
LGTM

package.json
jonathonherbert: Yes, that's fair! I did think about `test:unit`.

PR #22, opened 2021-05-13
Generate embed schemas from embed props

jonathonherbert:
## What does this change?

#19 added node-based modelling for embed properties, but left those properties hardcoded in a custom schema. This PR adds an API for embeds to define their own properties. A prop's core spec is basic at the moment, with just two properties:

```ts
/**
 * A property of an embed, to be modelled as a Node in Prosemirror.
 */
interface Prop {
  // The data type of the property.
  type: string;
  // The name the property should have in the schema. Should be unique within the schema.
  name: string;
}
```

So, our super-basic `image` embed can express its modelling needs like so: 

```ts
export const imageProps = [
  {
    type: "richText",
    name: "caption",
  },
  {
    type: "richText",
    name: "altText",
  },
];
```

And later, when somebody is writing the embed, they'll find those properties have made the appropriate Prosemirror fields available in React in the types (difficult to show, I'm hovering over `nestedEditorProps` in this screenshot, which is where those properties are available)

Screenshot 2021-05-13 at 17 37 12

... so they can be accessed typesafely when people are writing embeds:

Screenshot 2021-05-14 at 12 14 12

At the moment, we have just a `richText` prop, but eventually we'll want to support the full gamut of types required by our elements – basic strings, numbers, booleans, etc. 

From these properties, we can do two things –

- dynamically generate the Prosemirror NodeSpec that will support these nodes, to be added to the schema by the plugin consumer
- parameterise the type of the passed Props, and use that type to ensure that embed creators have typesafe access to their NodeViews in e.g. React, as shown above (We can test this with the wonderful [`// @ts-expect-error`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments).

**NB:** It strikes me that `Prop` isn't a great name – something like Field would be better, as it relates to a presentation and not just a data type. Field is currently taken by our legacy Prosemirror field-based approach, but another refactor will likely change that soon – although we may need a field-based approach for some of these data types.

## How to test

- The automated tests should pass. The new unit tests should adequately cover the new functionality.


src/__tests__/mount.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      expect(nodeSpec.size).toBe(0);
+    });
+
+    it("should create an nodeSpec with a parent node for each embed", () => {
+      const testEmbed1 = createNoopEmbed("testEmbed1", []);
+      const testEmbed2 = createNoopEmbed("testEmbed2", []);
+      const { nodeSpec } = build([testEmbed1, testEmbed2]);
+      expect(nodeSpec.size).toBe(2);
+      expect(nodeSpec.get("testEmbed1")).toMatchObject({ content: "" });
+      expect(nodeSpec.get("testEmbed1")).toMatchObject({ content: "" });
```
SHession, line 57: Looks like a typo here, this second line should have "testEmbed2".

jonathonherbert, line 57:
```suggestion
      expect(nodeSpec.get("testEmbed2")).toMatchObject({ content: "" });
```

src/prosemirrorSetup.ts
```diff
@@ -1,28 +1,20 @@
...
-// create a schema with list support.
-export const mySchema = new Schema({
-  nodes: addEmbedNode(schema.spec.nodes as OrderedMap<NodeSpec>),
-  marks: schema.spec.marks,
-});
-
-const parser = DOMParser.fromSchema(mySchema);
-const serializer = DOMSerializer.fromSchema(mySchema);
+export const createParsers = <S extends Schema>(schema: S) => {
+  const parser = DOMParser.fromSchema(schema);
```
SHession, line 5: Tiny one, why do we need this Type Parameter?

src/nodeSpec.ts
```diff
@@ -0,0 +1,74 @@
+import OrderedMap from "orderedmap";
+import type { Node, NodeSpec } from "prosemirror-model";
+import type { ElementProp, ElementProps } from "./types/Embed";
+
+export const getNodeSpecFromProps = <Props extends ElementProps>(
+  embedName: string,
+  props: Props
+): OrderedMap<NodeSpec> => {
```
SHession, line 8: This is really cool!

SHession (approved):
This looks great!

src/types/Embed.ts
```diff
@@ -8,10 +8,43 @@ export type NestedEditorMap<LocalSchema extends Schema = Schema> = Record<
...
-  commands: ReturnType<TCommandCreator>
-) => (fields: TFields, commands: ReturnType<TCommandCreator>) => void;
+/**
+ * A property of an embed, to be modelled as a Node in Prosemirror.
+ */
+interface Prop {
+  // The data type of the property.
+  type: string;
+  // The name the property should have in the schema. Should be unique within the schema.
+  name: string;
```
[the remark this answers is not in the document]
jonathonherbert: Discussed out of band, there are a few ways we might do this. Probably the best way would be enforcing a namespace programmatically, which we might be able to do with [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) – for example, the fields `caption` and `altText` in an `imageEmbed` would become `imageEmbed_caption` and `imageEmbed_altText` or similar. One for a separate PR, but worth pursuing for sure.

src/prosemirrorSetup.ts
jonathonherbert: It guarantees that the parsers and serialisers that are produced by this function are typed to accept that schema – the type def for DOMParser.fromSchema will infer the schema type from its argument. If we didn't, it'd be typed by default to `Schema<any>`.

To be honest, this isn't necessarily all that important, as the Prosemirror types for `Schema` are a bit anemic anyway – it's more of a nice to have.

PR #23, opened 2021-05-17
Add checkbox `Prop`

jonathonherbert:
## What does this change?

Adds a checkbox field to our list of props. That makes two of 'em!

It's declared in the embed props like this:

Screenshot 2021-05-17 at 09 18 49

And instantiated in React like this:

Screenshot 2021-05-17 at 09 19 35

To appear like this:

Screenshot 2021-05-17 at 09 29 37

## Dev notes

The `checkbox` prop is different from the rich text prop, in that its data is managed by a `field` attribute on the node itself – the node contains no content, and always occupies a single character width in the schema: 

Screenshot 2021-05-17 at 09 24 09

I've abstracted this concept of a node that uses Fields for state to a `FieldNodeView`, that has two methods that implementations must fulfil – `createInnerView` and `updateInnerView`. 

So if we need new props – and we will! – the only thing we need to implement in the child class is the code that's responsible for maintaining the state of the DOM, which should make our field implementations quite thin. The code to implement the view for a checkbox is fifteen lines of DOM manipulation. (If we'd like to use something else to render things and manage state – like React – we can do that, but for simple things like inputs I thought it was best to avoid the overhead.)

## How to test

- The automated tests should pass, and be comprehensive enough that you've got confidence in the code.
- Have a play. Does everything behave as expected?
- Is the API reasonable? Is there anything we could change to make things clearer for the consumer?

We've a new type of field that's easy to use – and a useful abstraction that makes it easy to add more.


SHession:
There is an empty helpers file in here which should presumably be removed: src/nodeViews/helpers.ts

SHession (approved):
This looks great!

PR #24, opened 2021-05-17
Remove Fields (flat state) in favour of Props (nested state), which are now called ... Fields

jonathonherbert:
## What does this change?

This PR gets rid of the last vestiges of the previous sort of state management (a single, empty parent node, with lots of fields), in favour of the current sort of state management (one to many child nodes, which have content or fields).

We now display the current values of the fields, and their validation output, to show that everything's working as intended:

Screenshot 2021-05-17 at 17 38 05

In doing the above, I've tried to 
- keep things as typesafe as possible. As a result, the validation and field values passed to the consumer code are now typesafe, with tests
- tried to name things sensibly, and comment the more complicated type wrangling that's needed for the above.

As a result, this diff is larger than I intended, mainly because there's lots of renaming – v. happy to pair on reviewing.

There's one bug, which is longstanding but only visible now – initial values aren't reflected in the field values until the editor state changes. That'll have to wait for another PR.

## How to test

- The automated tests should pass. The new ones should make us feel confident we've got the types right
- The names for types should make at least _some_ sense. Lmk if things need renaming, or bear some explanation w/ comments.

### DIFF
--- a/src/__tests__/mount.spec.ts
+++ b/src/__tests__/mount.spec.ts
@@ -50,6 +50,10 @@ describe("mount", () => {
         prop1: {
           type: "richText",
         },
+        prop2: {
+          type: "checkbox",
+          defaultValue: true,
+        },
       } as const;
       mount(
         "testEmbed",
@@ -58,6 +62,8 @@ describe("mount", () => {
         (fields) => {
           // Prop1 is derived from the fieldSpec, and is a string b/c it's a richText field
           fields.prop1.toString();
+          // Prop2 is a boolean b/c it's a checkbox field
+          fields.prop2.value.valueOf();
         },
         () => null,
         { prop1: "text" }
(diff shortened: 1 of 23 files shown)


src/__tests__/mount.spec.ts
```diff
@@ -1,73 +1,115 @@
...
         "testEmbed",
-        [props],
+        fieldSpec,
         () => () => null,
-        (_, __, ___, nestedEditorProps) => {
-          // Prop1 is derived from the props
-          nestedEditorProps.prop1;
+        (fields) => {
+          // Prop1 is derived from the fieldSpec, and is a string b/c it's a richText field
+          fields.prop1.toString();
```
SHession, line 60: It would be good to see an equivalent test for a checkbox field. Probably should have mentioned that in #23...

src/__tests__/mount.spec.ts
```diff
@@ -1,73 +1,115 @@
...
     });
 
     it("should create child nodes for each embed prop, and the parent node should include them in its content expression", () => {
-      const testEmbed1 = createNoopEmbed("testEmbed1", [
-        {
+      const testEmbed1 = createNoopEmbed("testEmbed1", {
+        prop1: {
           type: "richText",
-          name: "prop1",
         },
```
SHession, line 107: As with above, it would be good to see a checkbox example.

SHession (approved):
Looks great! I've added a couple of minor comments.

jonathonherbert:
Thanks for the review! I've added those specs as you say, nice to have them. In future, it'd be nice to colocate the nodeSpec generation with the field themselves, in which case field-specific tests can be written that cover these cases, too – rather than having the `mount` need to know about checkboxes etc.

PR #25, opened 2021-05-20
Create node-based embeds with existing data

jonathonherbert:
## What does this change?

At the moment, creating embeds with field content doesn't fill embeds with that content, you just get a blank embed.

This PR fixes that, creating the necessary nodes according to their specification.

There's not a way to do this in the testing UI for now, but I've added unit tests that should convincingly show that this is possible and works as expected, including typesafety for the added attributes. It can also be tested by changing code in `index.ts`.

## How to test

- The automated tests should pass. They should be comprehensive enough that you're happy this works.
- Run the test server with `yarn start`, and experiment with the default values provided in `index.ts` on L70-72. Does altering these values and inserting embeds match the expected behaviour? (Don't forget that rich text fields expect HTML input!)


cypress/tests/ImageEmbed.spec.ts
```diff
@@ -66,7 +66,7 @@ describe("ImageEmbed", () => {
           typeIntoEmbedField("caption", "Caption text");
           typeIntoEmbedField("altText", "Alt text");
           assertDocHtml(
-            `<imageembed type="imageEmbed" has-errors="false"><div class="ProsemirrorEmbed__imageEmbed-caption"><p>Caption text</p></div><div class="ProsemirrorEmbed__imageEmbed-altText"><p>Alt text</p></div><embed-imageembed-usesrc class="ProsemirrorEmbed__imageEmbed-useSrc" fields="{&quot;value&quot;:false}"></embed-imageembed-usesrc></imageembed><p>First paragraph</p><p>Second paragraph</p>`
+            `<imageembed type="imageEmbed" has-errors="false"><div class="ProsemirrorEmbed__imageEmbed-altText"><p>Alt text</p></div><div class="ProsemirrorEmbed__imageEmbed-caption"><p>Caption text</p></div><embed-imageembed-usesrc class="ProsemirrorEmbed__imageEmbed-useSrc" fields="{&quot;value&quot;:false}"></embed-imageembed-usesrc></imageembed><p>First paragraph</p><p>Second paragraph</p>`
```
jonathonherbert, line 69: Sadly necessary, as ordering now matters for our fields – and so we've changed our markup, which will always contain them in alphabetical order . It shouldn't matter for our consumers, which will have other ways of getting hold of element data.

src/__tests__/embed.spec.ts
```diff
@@ -34,5 +35,111 @@ describe("buildEmbedPlugin", () => {
...
+      const testEmbed = createNoopEmbed("testEmbed", {
+        prop1: { type: "checkbox", defaultValue: { value: false } },
+      });
+      const { insertEmbed } = buildEmbedPlugin([testEmbed]);
+      insertEmbed("testEmbed", {
+        prop1: { value: true },
+      });
+      insertEmbed("testEmbed", {
+        // @ts-expect-error -- we should not be able to insert a non-existent field
+        prop1: "This shouldn't typecheck",
```
SHession, line 49: The explanation for this error is misleading. The field does exist, it's just the wrong type.

SHession (approved):
LGTM!

src/embed.ts
```diff
@@ -2,50 +2,72 @@ import OrderedMap from "orderedmap";
...
+export const buildEmbedPlugin = <
+  FSpec extends FieldSpec<string>,
+  Name extends string
+>(
+  embedSpecs: Array<TEmbed<FSpec, Name>>,
   predicate = defaultPredicate
 ) => {
-  const typeNames = embedSpec.map((_) => _.name);
+  const embedTypeMap = embedSpecs.reduce<
+    Partial<{ [name in Name]: TEmbed<FSpec, Name> }>
```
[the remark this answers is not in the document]
jonathonherbert: Yes, it's a bit vague – renamed to EmbedName(s).

PR #28, opened 2021-05-26
Remove T prefix for types

jonathonherbert:
## What does this change?

Removes the `T` prefix for many types. This convention was used inconsistently and was not enforced.

## How to test

This is a no-op – test should pass as before.


rhystmills (approved):
This PR makes me happy - looks good to me. Test all running as expected 👍

PR #29, opened 2021-05-26
Refactor: remove the consumer argument from `createEmbedSpec`

jonathonherbert:
## What does this change?

Removes the consumer function that's passed through `createEmbedSpec`. It's not used by `createEmbedSpec`, and is necessarily already in scope in the consumer context – so we don't need it as an argument. This simplifies the signature of `createEmbedSpec`.

# How to test

This is a no-op – test should pass as before.


SHession (approved):
LGTM!

PR #30, opened 2021-05-26
Embeds -> Elements

jonathonherbert:
## What does this change?

This repo talks a lot about embeds – but what we're really referring to are Elements. Within the flexible-content model, a block contains many Elements, some of which may be embeds from e.g. Youtube, Twitter etc. – but Elements also includes images, media atoms, pullquotes etc, which we want to model here, too.

This PR renames everything that refers to an Embed accordingly.

## How to test

This is a no-op – the tests should pass as usual.


SHession (approved):
Great change, and not a moment too soon!

PR #34, opened 2021-06-09
Renames "NodeView" to "FieldView"

SHession:
## What does this change?

This PR aims to improve the readability of the code by renaming "NodeView" -> "FieldViews" which provides both a clearer explanation of what it represents and avoid any conflation with the concept of a ProseMirror "NodeView".

This also renames "RTE" -> "RichText" as this is a more verbose name.

## How to test

This change should be a no-op

The classes representing the binding logic for a given field is easier to understand.

## Accessibility


rhystmills (approved):
Looks good to me - a sensible renaming effort.

It's running locally as expected and all tests are passing.

PR #35, opened 2021-06-14
Add text input field

SHession:
## What does this change?

This PR adds a text field option to the element API, allowing consumers to specify simple text inputs with no decorations.
Rich Text and Text fields are both handled as ProseMirror nodes, meaning collaborative editing should be possible to support within these fields. 

The Rich Text field view has been refactored into `ProseMirrorFieldView` with two derived classes `TextFieldView` and `RichTextFieldView`.  `ProseMirrorFieldView` handles the setup and integration with the parents PM instance. The derived class then simply specify how they would like the schema to look and what plugins should be added.

The rest of the setup is very similar to the existing field type options.

## How to test

The integration and unit tests should report healthy. Its also possible to run the project locally and test the functionality manually.

Consumer elements can specify text fields which support collaborative editing.


cypress/helpers/editor.ts
```diff
@@ -61,3 +61,18 @@ export const assertDocHtml = (expectedHtml: string) =>
...
+  altTextValue = "<p></p>",
+  captionValue = "<p></p>",
+  srcValue = "",
+  useSrcValue = "false",
+  mainImageValue = "",
+}: {
+  altTextValue?: string;
+  captionValue?: string;
+  srcValue?: string;
+  useSrcValue?: string;
```
jonathonherbert, line 75: Very sensible change – tests much more readable!

src/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -0,0 +1,253 @@
...
+    // The offset of this node relative to its parent FieldView.
+    private offset: number,
+    // The schema that the internal editor should use.
+    schema: LocalSchema,
+    // The initial decorations for the FieldView.
+    decorations: DecorationSet | Decoration[],
+    // The ProseMirror node type name
+    private readonly propName: string,
+    // Plugins that the editor should use
+    private plugins?: Plugin[]
```
jonathonherbert, line 48: Possible nit: if we don't use plugins outside of the construction of the editor, let's pass this into `createInnerEditorView` as an argument, instead, to keep them scoped to where they're used.

jonathonherbert (commented):
Looks great – the inheritors are nice and neat! One minor thing – hitting shift+enter causes an error:

Screenshot 2021-06-15 at 11 42 17

Looks similar to the issues caused by a bug in prosemirror-transform, and may be solved by upgrading that dependency. V. happy to pair.

jonathonherbert:
One other note – we shouldn't be able to style this text w/ keyboard shortcuts etc. 

Screenshot 2021-06-15 at 12 08 36

This might be because we're using `exampleSetup`, which seems to have a basic set of bindings that includes styling shortcuts.

jonathonherbert:
Hm, looks like you might be able to [explicitly disable marks for this node type](https://prosemirror.net/docs/ref/#model.NodeSpec.marks) in the schema, also.

SHession:
> Hm, looks like you might be able to [explicitly disable marks for this node type](https://prosemirror.net/docs/ref/#model.NodeSpec.marks) in the schema, also.

Explicitly setting marks to "" didn't seem to help, but removing the example setup plugin has worked and I don't think it's necessary to include.

jonathonherbert (approved):
Changes look great 👍

PR #36, opened 2021-06-15
Proper tags for content nodes

jonathonherbert:
## What does this change?

Previous to this PR, fields of type CONTENT were represented as `div` elements when serialised. This meant their order was important when parsing. It also meant we were vulnerable to Prosemirror making odd guesses when these nodes were empty.

This PR adds proper tags for these fields for serialisation.

It also adds a failing test for an odd edge-case we saw when content nodes were empty, and also adds a general test for parsing, round-tripping content through Prosemirror and back into HTML and ensuring everything matches up.

## How to test

The automated tests should pass, and you should be convinced they test for the problem at hand. You should be able to see they fix the odd parsing problem by checking out `3f91c7e` and watching the test fail (the tag names will be incorrect, but more importantly, the value of the `checkbox` field will also be incorrect).

You should be able to see nodes serialised correctly in localStorage when running the project locally.


rhystmills (approved):
Looking good to me and running as intended locally. Also happy to see those massive HTML strings broken up!

PR #41, opened 2021-06-21
Add dropdown field

rhystmills:
## What does this change?

This PR adds a dropdown field to the image element.

## How to test

This PR includes integration tests for the dropdown field.

These can be run by:
1. Running the application with `yarn start`
2. Running the integration tests with `yarn test:integration`

## Accessibility

The dropdown is usable with a screenreader but not especially user-friendly (it's not clear it's a dropdown with multiple options) - I'm leaving this for now as it's likely we will change how the rendering of the elements occurs - e.g. via a Source component in React. We may want to revisit this as part of the accessibility card that's currently open - and Source would likely cover this with its existing `Select` component if we move over to Source.

### DIFF
--- a/src/fieldViews/AttributeFieldView.ts
+++ b/src/fieldViews/AttributeFieldView.ts
@@ -1,4 +1,4 @@
-import type { Node } from "prosemirror-model";
+import type { Node, NodeType } from "prosemirror-model";
 import type { EditorView } from "prosemirror-view";
 import type { FieldView } from "./FieldView";
 import { FieldType } from "./FieldView";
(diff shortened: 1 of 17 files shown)


src/fieldViews/AttributeFieldView.ts
```diff
@@ -22,11 +22,8 @@ export abstract class AttributeFieldView<Fields extends unknown>
     // Returns the current position of the parent FieldView in the document.
     private getPos: () => number,
     // The offset of this node relative to its parent FieldView.
-    private offset: number,
-    defaultFields: Fields
-  ) {
-    this.createInnerView(node.attrs.fields || defaultFields);
```
jonathonherbert, line 28: Is it worth adding a note on the parent class that inheritors must make this call themselves, now we've removed it from the parent constructor? (Genuine question, maybe we feel it doesn't matter!)

rhystmills: Ah yes, definitely.

jonathonherbert (approved):
Tests look good, behaves as expected – great work 👍

PR #42, opened 2021-06-21
Fix incorrect fields in AttributeFieldView passed on update

jonathonherbert:
## What does this change?

Fixes a problem with incorrect fields being passed to inheritors of `AttributeFieldView` – a straightforward problem papered over by the loose typing of the attributes in Prosemirror's `Node` type.

This also uncovered another problem, where `sameMarkup` was inappropriately being used to figure out whether we should accept updates. This isn't right – this check includes a [deep comparison of attributes](https://github.com/ProseMirror/prosemirror-model/blob/eef20c8c6dbf841b1d70859df5d59c21b5108a4f/src/node.js#L131), which aren't relevant for an identity check in this case, and will always produce an incorrect result once field attributes are updated.

## How to test

The automated tests (unit and integration) should pass, and cover the cases discussed above.

### DIFF
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -120,6 +120,21 @@ describe("ImageElement", () => {
         getElementField("useSrc").find("input").click();
         assertDocHtml(getSerialisedHtml({ useSrcValue: "false" }));
       });
+
+      it(`should track the field offset, and update correctly after fields above have changed`, () => {
+        addElement();
+
+        getElementField("useSrc").find("input").click();
+        typeIntoElementField("altText", "Example text");
+        getElementField("useSrc").find("input").click();
+
+        assertDocHtml(
+          getSerialisedHtml({
+            altTextValue: "<p>Example text</p>",
+            useSrcValue: "false",
+          })
+        );
+      });
     });
 
     describe("Custom element – image src", () => {
--- a/src/fieldViews/__tests__/AttributeFieldView.spec.ts
+++ b/src/fieldViews/__tests__/AttributeFieldView.spec.ts
@@ -1,0 +1,77 @@
+import type { Node } from "prosemirror-model";
+import { Schema } from "prosemirror-model";
+import { schema } from "prosemirror-schema-basic";
+import { createEditorWithElements } from "../../__tests__/helpers";
+import { getNodeSpecForProp } from "../../nodeSpec";
+import { AttributeFieldView } from "../AttributeFieldView";
+
+const createInnerView = jest.fn();
+const updateInnerView = jest.fn();
+
+type TestAttributeFields = { value: boolean };
+
+class TestAttributeFieldView extends AttributeFieldView<TestAttributeFields> {
+  public static propName = "checkbox" as const;
+  public static defaultValue = "default";
+
+  public getNodeValue(node: Node): TestAttributeFields {
+    return node.attrs.fields as TestAttributeFields;
+  }
+
+  protected createInnerView(fields: TestAttributeFields) {
+    createInnerView(fields);
+  }
+
+  protected updateInnerView(fields: TestAttributeFields) {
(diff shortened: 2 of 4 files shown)


cypress/tests/ImageElement.spec.ts
```diff
@@ -120,6 +120,21 @@ describe("ImageElement", () => {
         getElementField("useSrc").find("input").click();
         assertDocHtml(getSerialisedHtml({ useSrcValue: "false" }));
       });
+
+      it(`should track the field offset, and update correctly after fields above have changed`, () => {
+        addElement();
+
+        getElementField("useSrc").find("input").click();
+        typeIntoElementField("altText", "Example text");
+        getElementField("useSrc").find("input").click();
```
SHession, line 129: I think this test could be improved, given the `useSrcValue` represents a boolean value it's possible to imagine this test passing if both the checkbox clicks fail.  We could possibly improve by using a dropdown field instead or asserting between checkbox clicks.

src/fieldViews/__tests__/AttributeFieldView.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+      value: false,
+    });
+
+    const newNode = testSchema.nodes.testField.create({
+      fields: { value: true },
+    });
+
+    fieldView.update(newNode, 0);
+
+    expect(updateInnerView.mock.calls[0]).toEqual([{ value: true }]);
```
SHession, line 75: Could we also use `getNodeValue` to confirm the update has worked?

SHession (approved):
Improvements look great and that tests should help prevent any future issues. I've added a couple of suggestions as to how we might improve the tests but otherwise LGTM.

cypress/tests/ImageElement.spec.ts
jonathonherbert: Great point. Could give this a tweak once #41 is merged, perhaps!

src/fieldViews/__tests__/AttributeFieldView.spec.ts
jonathonherbert: `getNodeValue` operates on a passed node, it doesn't reveal the value of the instance's current node – so I don't think this will help. It should probably be static. It's a bit of an appendix right now, and unless we refactor it soon it might be worth removing to avoid confusion!

SHession: Could we find the updated node and check its value?

jonathonherbert: It's a private member of this class – and we already have access to the node as we're passing it in, so we'd just be checking our own homework, if that makes sense. We could make the node property public, via a getter, perhaps? This'd be the only thing that used it for now, mind.

It's pretty interesting – if we did check this, we'd discover that our `node` value is indeed never updated! But this doesn't really matter for this class or its children – as only the values of the node are relevant. This is probably unexpected behaviour were we to rely on the node value elsewhere, mind. I'm tempted to rename `node` `initialNode` instead of adding a test, to make this clearer?

jonathonherbert: (The only reason we need to keep a node reference around at all is to match the incoming node to the initial node's `type`. With some creative thinking we could probably remove even this dependency.)

SHession: Sorry I'd slightly got the wrong end of the stick, I had meant we could check the ProseMirror version of that node's value but I was thinking of changes coming out of the field view. 

What you're suggesting make sense, from what I can see we're only concerned with the NodeType so we could store that instead.

jonathonherbert: Gotcha. Yeah, reckon the latter makes sense and is least surprising, I'll do that – really useful feedback.

PR #44, opened 2021-06-24
Add uuids to nodes representing fields

jonathonherbert:
## What does this change?

Adds uuids to nodes representing fields, for quick `update` checks within FieldViews.

Because the node we receive in `update` is [not necessarily the _right_ node](https://prosemirror.net/docs/ref/#view.NodeView.update), we must check to make sure that it is before continuing with updating our fieldView.

We are no longer able to rely on `sameMarkup` to run this check as a result of #43 – but this is a small optimisation regardless, as `sameMarkup` will run on every editor update, and comparing uuids is more efficient.

## How to test

To the user, this is a no-op – everything should work as expected. I've added automated tests to ensure that uuids are present on every node representing a fieldView, whether it was created programmatically (via `.create`) or via serialisation (via `parseDOM`).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,25105 @@
+{
+  "name": "prosemirror-elements",
+  "version": "1.0.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "prosemirror-elements",
+      "version": "1.0.0",
+      "license": "MIT",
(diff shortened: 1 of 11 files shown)


jonathonherbert:
This should no longer be necessary as a result of #43, and  is possibly not the right thing to do in any case (nodes can be duplicated by PM on copy/paste or split).

PR #45, opened 2021-06-30
Add image select field

SHession:
## What does this change?

This PR adds a new image selector field to the existing custom field. The image selector allows a user to pick an image from the Grid and retain the necessary information within the field node's attributes. 

This change adds a `Props` type parameter to the custom field view, allowing a consumer to specify additional properties that are required for rendering the field correctly. In this case, it's used to pass the field image select and crop callback methods which trigger an iframe and provide a function to call with the chosen image data.

The iframe behaviour is facsimile of the `flexible-content` image selection integration and provides broadly the same functionality although implemented a bit differently. This has also meant we need to use dex-nginx so the panda auth will work correctly. The new URL is https://prosemirror-elements.local.dev-gutools.co.uk/. 

## How to test

Load a new instance of the element, press the  "choose image" button, select an image and image crop, observe the image is now attach to the element.

We should functionally be able to use the image-selector field in composer. 

## Images

image_selector_2

### DIFF
--- a/src/elements/image/ImageElement.tsx
+++ b/src/elements/image/ImageElement.tsx
@@ -1,4 +1,3 @@
-import _ from "cypress/types/lodash";
 import React from "react";
 import type { FieldNameToValueMap } from "../../fieldViews/helpers";
 import { getPropViewTestId, PropView } from "../../renderers/react/PropView";
@@ -49,7 +48,7 @@ type ImageViewProps = {
     },
     {
       onSelectImage: (setMedia: SetMedia) => void;
-      onCropImage: (mediaId: string | undefined, setMedia: SetMedia) => void;
+      onCropImage: (mediaId: string, setMedia: SetMedia) => void;
     }
   >;
 };
@@ -75,12 +74,16 @@ const ImageView = ({ fieldViewProp }: ImageViewProps) => {
 
       {imageFields.mediaId ? (
         <button
-          onClick={() =>
-            fieldViewProp.fieldSpec.props.onCropImage(
-              imageFields.mediaId,
-              setMedia
-            )
-          }
+          onClick={() => {
+            if (imageFields.mediaId) {
+              fieldViewProp.fieldSpec.props.onCropImage(
+                imageFields.mediaId,
+                setMedia
+              );
+            } else {
+              fieldViewProp.fieldSpec.props.onSelectImage(setMedia);
+            }
+          }}
         >
           Crop Image
         </button>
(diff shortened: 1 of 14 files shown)


src/elements/image/ImageElement.tsx
```diff
@@ -40,22 +40,60 @@ export const ImageElement: React.FunctionComponent<Props> = ({
...
     <div data-cy={getPropViewTestId(fieldViewProp.name)}>
-      <input
-        value={imageFields.src || ""}
-        onChange={(e) => setImageFieldsRef.current?.({ src: e.target.value })}
-      ></input>
-      {JSON.stringify(imageFields)}
+      {imageFields.assets.length > 0 ? (
+        <img style={{ width: "25%" }} src={imageFields.assets[0]}></img>
+      ) : (
+        ""
```
jonathonherbert, line 72: `null` is IIRC the idiomatic way to express nothing in a React tree.

src/helpers.ts
```diff
@@ -204,4 +205,85 @@ const createDecorations = (name: string) => (state: EditorState) => {
   return DecorationSet.create(state.doc, decorations);
 };
 
-export { buildCommands, defaultPredicate, createDecorations };
+const onGridMessgae = (setMedia: SetMedia, modal: HTMLElement) => ({
```
jonathonherbert, line 208:
```suggestion
const onGridMessage = (setMedia: SetMedia, modal: HTMLElement) => ({
```

src/helpers.ts
```diff
@@ -204,4 +205,85 @@ const createDecorations = (name: string) => (state: EditorState) => {
...
+
+const onSelectImage = (setMedia: SetMedia) => {
+  const modal = document.querySelector(".modal") as HTMLElement;
+  modal.style.display = "Inherit";
+
+  (document.querySelector(
+    ".modal__body iframe"
+  ) as HTMLIFrameElement).src = `https://media.test.dev-gutools.co.uk/`;
+
+  const listner = onGridMessgae(setMedia, modal);
```
jonathonherbert, line 243:
```suggestion
  const listener = onGridMessage(setMedia, modal);
```

src/index.ts
```diff
@@ -16,7 +17,9 @@ const {
   insertElement,
   hasErrors,
   nodeSpec,
-} = buildElementPlugin([createImageElement("imageElement")]);
+} = buildElementPlugin([
+  createImageElement("imageElement", onSelectImage, onCropImage),
```
jonathonherbert, line 21: Yess, nice and neat ✨

jonathonherbert (approved):
Tested locally, works as expected for both image selection and cropping. Ace! A few very minor suggestions.

jonathonherbert (approved):

PR #46, opened 2021-06-30
Source-ify components

rhystmills:
## What does this change?

This adds some Source stylings to our current setup, bringing it closer to the style of UI seen in Composer and other Editorial Tools.

Our aim is to use Source in order to make styling faster and easier, and to take advantage of the work that's already been done (e.g. managing focus state, using Guardian fonts and colours) - but to modify it for our Ed Tools context (limited space, need to get lots of functionality into an interface).

## What does this not change?
- The styling and schema of the ProseMirror menu bars. This will already be done on the consuming application (i.e. Composer) and we would like to pass it in from there rather than recreating it (this should sort the odd focus state for the Rich Text Editor fields).
- Styling the Dropdown and Checkbox elements. These elements use DOM manipulation rather than React - which we are using in the Source-styled elements. While Source+React provides a good way to provide Guardian-styled elements, we want the DOM manipulation route to be viable for authors rather than enforcing one method of element rendering. Source Components like the Source Checkbox are designed to work with React so we can't use them for the existing CheckboxField.

## How to test

Run `yarn start` from this branch, or deploy this branch to CODE.


jonathonherbert (approved):
This looks great. Let's continue to pursue this approach as we build our first proper element.

PR #47, opened 2021-07-01
Correct CI script

jonathonherbert:
## What does this change?

At the moment, CI is hanging indefinitely when either linting or testing fails. IIUC, this is because (in `bash` at least) the & operator looks to have a higher associativity than &&, so if linting or tests fail, webpack serve is never called and we run wait-on in parallel, which will run indefinitely.

To test this, try running the following commands in your shell:

```bash
false && true && echo "This shouldn't appear" # no output
false && true & echo "This shouldn't appear" # echoes 'This shouldn't appear'
false && (true & echo "This shouldn't appear") # no output
```

## How to test

Introduce e.g. a linting mistake and run `yarn ci`. Does the script fail fast with an exit code of `1`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ This Prosemirror plugin adds the ability to add element nodes to a document.
 
 `yarn start` builds the project locally, spins up a webserver, and watches for file changes.
 
-
 ## Testing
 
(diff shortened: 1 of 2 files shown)


SHession (approved):
👍

PR #48, opened 2021-07-02
Restructure to better reflect project organisation

jonathonherbert:
## What does this change?

Moves our code about to better reflect the structure of the project. There are now four folders within `src`:

- `editorial-source-components` (this is an outlier likely to eventually disappear into the design system or elsewhere, and components here should be candidates for reuse elsewhere)
- `elements` (all of the example elements defined for the )
- `plugin`
- `renderers`

The `index` file within `src` is now designed for library exports. The demo code has been moved to `demo/index.ts`.

I've shuffled a few bits and pieces around in the `plugin` directory to put all of our helpers in their own directory, too.

The end result should be something that's easier to understand and contribute to.

## How to test

The diff's a big one – might be easiest to take a look at this in your IDE! Does this structure make more sense? Is there anything else we could tweak?


SHession (approved):
This change looks very sensible and should make the project easier to understand.

A couple of comments:

- ~~There are a few linter errors, mainly related to Guardian src (I assume these should be picked up in the CI)~~
- This would be a good opportunity to rename `mount.spec.ts` -> `elementSpec.spec.ts`

PR #49, opened 2021-07-02
Allow passing plugins into rich text editor fields

jonathonherbert:
## What does this change?

Creates a property in a RichTextField field declaration, `plugins: Plugin[]`, to allow element to pass plugins into fields directly. This replaces the current approach, where we just add an `exampleSetup` plugin arbitrarily.

This will eventually allow consumers to pass plugins to elements to govern RTE behaviour – e.g. at the Guardian we may do this to supply custom RTE menu bars in Composer, amongst other things.

I've added an exampleSetup plugin but ~turned menu bars off to show that the field is now governing plugin behaviour~ left menu bars in, as our integration tests rely on them 😅 

## How to test

This is a no-op – the automated tests should pass.

### DIFF
--- a/src/plugin/helpers/plugin.ts
+++ b/src/plugin/helpers/plugin.ts
@@ -29,7 +29,7 @@ export const getElementFieldViewFromType = (
         getPos,
         offset,
         innerDecos,
-        prop.plugins ?? []
+        field.plugins ?? []
       );
     case "checkbox":
       return new CheckboxFieldView(
--- a/src/plugin/helpers/plugin.ts
+++ b/src/plugin/helpers/plugin.ts
@@ -23,7 +23,14 @@ export const getElementFieldViewFromType = (
     case "text":
       return new TextFieldView(node, view, getPos, offset, innerDecos);
     case "richText":
-      return new RichTextFieldView(node, view, getPos, offset, innerDecos);
+      return new RichTextFieldView(
+        node,
+        view,
+        getPos,
+        offset,
+        innerDecos,
+        prop.plugins ?? []
+      );
     case "checkbox":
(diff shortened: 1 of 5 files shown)


src/plugin/helpers/plugin.ts
```diff
@@ -23,7 +23,14 @@ export const getElementFieldViewFromType = (
...
       return new TextFieldView(node, view, getPos, offset, innerDecos);
     case "richText":
-      return new RichTextFieldView(node, view, getPos, offset, innerDecos);
+      return new RichTextFieldView(
+        node,
+        view,
+        getPos,
+        offset,
+        innerDecos,
+        field.createPlugins ? field.createPlugins(node.type.schema) : []
```
SHession, line 32: We could move the example setup plugins here as a default?

SHession (approved):
Sensible change, LGTM!

src/plugin/helpers/plugin.ts
jonathonherbert: Coooould do! I'd be tempted to move `exampleSetup` into devDependencies and avoid bundling it in the library ... but it's probably quite little.

jonathonherbert: On reflection, I think if we decide we'd like to provide it, let's make it a field creator re: #54 (for example, `createDefaultRichTextField` or similar), to make sure plugin consumers that don't require it can tree shake that dependency out.

I do think that it makes sense to provide it down the line, b/c anyone playing around with this library will want a quick way to express an RTE, for sure.

SHession: That's a great idea, we could rebase the field functions once they're merged and add the default one.

PR #53, opened 2021-07-08
Add CustomDropdown

rhystmills:
## What does this change?
This adds the `CustomDropdown` - a Select element based on the [Source Select Box component](https://theguardian.design/2a1e5182b/p/663879-select-box/b/10875c) (rendered using React).

Changes have been made to the styling in order to fit in the smaller screen space available in an Editorial Tool compared to our reader-facing website.

## How to test
Run `yarn start` from this branch locally.

## Accessibility

- [x] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)


src/editorial-source-components/CustomDropdown.tsx
```diff
@@ -0,0 +1,63 @@
+import { css } from "@emotion/react";
+import { space } from "@guardian/src-foundations";
+import { Option, Select } from "@guardian/src-select";
+import type { Option as OptionValue } from "../plugin/types/Element";
+import { inputBorder } from "./inputBorder";
+import { labelStyles } from "./Label";
+
+const parentStyles = css`
+  width: initial;
+  div {
```
jonathonherbert, line 10: Is this to style the wrapper around Select – and if so, would it be worth documenting that coupling with a comment to explain what we're doing? If the shape of Select changes and breaks this, for example, a comment might make that link clearer.

src/editorial-source-components/CustomDropdown.tsx
```diff
@@ -0,0 +1,63 @@
...
+  :hover {
+    cursor: pointer;
+  }
+`;
+
+type CustomDropdownProps = {
+  options: Array<OptionValue<unknown>>;
+  selected: string;
+  label: string;
+  changeHandler: (event: React.ChangeEvent<HTMLSelectElement>) => void;
```
jonathonherbert, line 39: I'd argue this should be called `onChange`, as with other elements, as per convention.

src/plugin/fieldViews/SelectFieldView.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A FieldView representing a
+ * node that contains arbitrary fields that are updated atomically.
+ *
+ * Instead of rendering into a DOM node that's mounted by the consumer, this FieldView
+ * instead provides a `subscribe` method that allows consuming code to listen for
+ * state changes. In this way, consuming code can manage state and UI changes itself,
+ * perhaps in its own renderer format.
+ */
+export class SelectFieldView<Fields = unknown> extends CustomFieldView<Fields> {
```
jonathonherbert, line 15: Is this still used? I can't see a reference elsewhere in the PR.

src/elements/image/ImageElementForm.tsx
```diff
@@ -94,3 +97,26 @@ const ImageView = ({ fieldViewSpec }: ImageViewProps) => {
     </div>
   );
 };
+
+type CustomDropdownViewProps = {
+  fieldViewSpec: CustomFieldViewSpec<string, Array<Option<unknown>>>;
```
jonathonherbert, line 102: Could we replace this `unknown` type w/ e.g. `string` (I think that's the right type here?)

jonathonherbert (changes requested):
Looks great, and nice work on the accessibility checks! A few comments.

I've an odd bug – on a clean working directory on this branch, when I change the select option, the state updates correctly, but the dropdown display does not. Can you replicate? Be v. happy to pair to show/pair on a fix.

src/plugin/fieldViews/SelectFieldView.ts
rhystmills: Whoops - I must have left that in during early experimentation and didn't realise I committed it.

Will get rid of it now.

src/editorial-source-components/CustomDropdown.tsx
rhystmills: Yes that's right - applying CSS to the `.svg` and the existing wrapper `div` wasn't possible in the existing Source Select component so I've added these styles to a new parent `div`.

I hadn't considered that the Select element might change in the future, especially as it's currently flagged as being 'in development' in Source. A comment (at the very least) would definitely be sensible.

src/editorial-source-components/CustomDropdown.tsx
rhystmills: Good point - will do!

src/elements/image/ImageElementForm.tsx
rhystmills: I was in two minds about this - I wanted it to be clear that the value here can be anything, e.g. a number or boolean (we later `JSON.stringify` it before using it as a property in the DOM).

In this instance though, they're all strings - so your suggestion would make sense.

rhystmills:
> Looks great, and nice work on the accessibility checks! A few comments.
> 
> I've an odd bug – on a clean working directory on this branch, when I change the select option, the state updates correctly, but the dropdown display does not. Can you replicate? Be v. happy to pair to show/pair on a fix.

Ah - thanks for pointing that out - I realise I introduced this earlier today. Should be sorted now.

src/elements/image/ImageElementForm.tsx
jonathonherbert: Oh, that's interesting! We could always widen that type to a union of `string|number|boolean` to accommodate those types – or even better, do some type magic to infer this. V. happy to pair on this if you'd like – but happy to leave as a string for now, also.

jonathonherbert: Resolved w/ a string type for now – as with conversation on #54, we think it'd be better to defer these sorts of things to the consumer.

src/editorial-source-components/CustomDropdown.tsx
jonathonherbert: Yeah, there's a balance here – I'd almost be in favour of copy-pasting the element in its entirety, as at least then we absolutely know what we're getting – and I suspect it wouldn't be large! Comment if not sounds good though.

rhystmills: That makes sense - though I think it would make sense to restructure the element a bit so as not to have to rely on indirect css rules (particularly the `!important` I had to add in.

rhystmills: Having said that - the Select element from Source is in its own package so I think we'd only see problems with it if someone specifically bumped the version of `@guardian/src-select` and the interface changed.

I also checked Source and Select isn't marked as 'in development' after all, so I think it's unlikely the interface will change much.

jonathonherbert: Let's use it as is for now. Should be easy to change if we need.

jonathonherbert: @rhystmills to double check, did you agree that this still merited a comment?

jonathonherbert (approved):
Looks great, works as expected! One minor additional comment re: conversation on comments, not a blocker.

src/editorial-source-components/CustomDropdown.tsx
rhystmills: Yes - I thought I had added one already actually, will do so now!

PR #54, opened 2021-07-08
add field creator functions to imageelements

dskamiotis:
## What does this change?
When creating elements at the moment we have to declare fields manually with object literals. This can mean that field declaration can be unnecessarily verbose and sometimes require type cast.

```ts 
mainImage: {
      type: "custom",
      defaultValue: { mediaId: undefined, mediaApiUri: undefined, assets: [] },
      props: {
        onSelectImage,
        onCropImage,
      },
    } as CustomField<
      { mediaId?: string; mediaApiUri?: string; assets: string[] },
      {
        onSelectImage: (setSrc: SetMedia) => void;
        onCropImage: (mediaId: string, setSrc: SetMedia) => void;
      }
    >
```
 This PR introduces `creator` functions for each field type; this enables a more concise way of declaring fields:
```ts
mainImage: createCustomField<ImageField, ImageProps>(
      { mediaId: undefined, mediaApiUri: undefined, assets: [] },
      {
        onSelectImage,
        onCropImage,
      }
    )
```

## How to test
All the current test should pass. 
Take a look at the new field creators, does it make it easier read and or work with?

## Dev notes
We decided to move the field types into their fieldView class files. This puts most field related code in the same place.

### DIFF
--- a/.metals/metals.lock.db
+++ b/.metals/metals.lock.db
@@ -1,6 +1,0 @@
-#FileLock
-#Thu Jul 08 09:49:44 BST 2021
-hostName=localhost
-id=17a758b0407b7070ce8882977c80562f71a310241ab
-method=file
-server=localhost\:49752
--- a/.metals/metals.lock.db
+++ b/.metals/metals.lock.db
@@ -1,0 +1,6 @@
+#FileLock
+#Thu Jul 08 09:49:44 BST 2021
+hostName=localhost
+id=17a758b0407b7070ce8882977c80562f71a310241ab
+method=file
+server=localhost\:49752
--- a/src/plugin/fieldViews/CheckboxFieldView.ts
+++ b/src/plugin/fieldViews/CheckboxFieldView.ts
@@ -3,9 +3,9 @@ import type { EditorView } from "prosemirror-view";
 import { AttributeFieldView } from "./AttributeFieldView";
 import type { BaseFieldSpec } from "./FieldView";
 
-export type CheckboxFields = { value: boolean };
+export type CheckboxValue = { value: boolean };
 
-export interface CheckboxField extends BaseFieldSpec<CheckboxFields> {
+export interface CheckboxField extends BaseFieldSpec<CheckboxValue> {
   type: typeof CheckboxFieldView.fieldName;
 }
 
@@ -14,7 +14,7 @@ export const createCheckBox = (defaultValue: boolean): CheckboxField => ({
   defaultValue: { value: defaultValue },
 });
 
-export class CheckboxFieldView extends AttributeFieldView<CheckboxFields> {
+export class CheckboxFieldView extends AttributeFieldView<CheckboxValue> {
   public static fieldName = "checkbox" as const;
   public static defaultValue = { value: false };
   private checkboxElement: HTMLInputElement | undefined = undefined;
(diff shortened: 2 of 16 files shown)


src/plugin/fieldViews/CheckboxFieldView.ts
```diff
@@ -1,9 +1,19 @@
 import type { Node } from "prosemirror-model";
 import type { EditorView } from "prosemirror-view";
 import { AttributeFieldView } from "./AttributeFieldView";
+import type { BaseFieldSpec } from "./FieldView";
 
 export type CheckboxFields = { value: boolean };
 
+export interface CheckboxField extends BaseFieldSpec<CheckboxFields> {
```
SHession, line 8: We have a type `CheckboxFields` and interface `CheckboxField` which is a bit confusing. I think `CheckboxFields` could be renamed to something more appropriate. Possibly how Fields is used in `AttributeFieldView` more accurately represents values?

src/plugin/fieldViews/DropdownFieldView.ts
```diff
@@ -1,13 +1,28 @@
 import type { Node } from "prosemirror-model";
 import type { EditorView } from "prosemirror-view";
-import type { Option } from "../types/Element";
 import { AttributeFieldView } from "./AttributeFieldView";
+import type { BaseFieldSpec } from "./FieldView";
+
+type Option = { text: string; value: string };
```
SHession, line 6: We've lost the type parameter here which allowed value to represent a variety of different types. Was this intentional?

SHession (commented):
A few comments here to think about. A couple of naming improvements that could be made and a loss of functionality in regards to dropdowns. 

I'm a fan of colocating the types with the field views 👍

src/plugin/fieldViews/DropdownFieldView.ts
jonathonherbert: Yes, and we should have added a note – IIUC an option selected in a dropdown will always we of type `string`, I don't think anything else was possible in practice.

jonathonherbert: I'll update the dev notes in the PR.

src/plugin/fieldViews/CheckboxFieldView.ts
jonathonherbert: Yes, let's rename `CheckboxField`, it should really be `CheckboxValue`. We could do with doing that across the board as you say, the original class names that type parameter `NodeValue`! Shall we include that in this PR?

src/plugin/fieldViews/DropdownFieldView.ts
SHession: It was serialising the value as JSON, then parsing it out to support this. Admittedly it might be unnecessary but should work.

src/plugin/fieldViews/CheckboxFieldView.ts
SHession: I think we could change `CheckboxFields` -> `CheckboxValue` in this PR as we are adding a new interface called `CheckboxField`. Other changes could reasonably be deferred to a separate PR.

src/plugin/fieldViews/DropdownFieldView.ts
jonathonherbert: Oh, wow, I missed this! As an aside, surely that sort of behaviour should be up to the consumer? It seems odd to bake that assumption into the core library, as a string type is IIUC by far the most likely choice for a consumer. (I don't recall ever seeing a dropdown that encoded JSON into its values. Perhaps I've just ... never truly lived 😁 )

We'll restore this behaviour in the meantime, as if we do want to make a change that should be a separate PR.

SHession: That's fair to say. It seemed like a reasonable idea at the time, incase you wanted the dropdown to represent different values. Possibly I didn't stop to question it enough.

.metals/metals.lock.db
```diff
@@ -0,0 +1,6 @@
+#FileLock
+#Thu Jul 08 09:49:44 BST 2021
+hostName=localhost
```
SHession, line 3: We don't usually version these files and I suspect they should be included in the `.gitignore`. It should probably be removed from this PR.

SHession (commented):
Name changes look great, it should be easier to follow now. I've added a comment about a metals file presumably added accidentally.

.metals/metals.lock.db
jonathonherbert: Ta, just done that.

SHession (approved):

dskamiotis:
> Name changes look great, it should be easier to follow now. I've added a comment about a metals file presumably added accidentally.

Yes @SHession that may well have been as a result of my push as I have used vs code for Scala before

PR #55, opened 2021-07-12
Save state correctly

jonathonherbert:
## What does this change?

#52 introduced a bug – it stopped us saving content locally! oh no :(

This fixes that, and adds a test for it.

## How to test

The automated tests should pass.

Whilst running the demo, make a change to the editor state in any editor, and refresh the browser page. Does the state persist after a refresh?


SHession (approved):
Tested locally, it was saving content as expected

PR #56, opened 2021-07-13
Fix intermittent cypress failures

jonathonherbert:
## What does this change?

#55 introduced an intermittent failure in our integration tests, caused by localStorage remaining populated between tests. This [shouldn't happen by default](https://docs.cypress.io/api/commands/clearlocalstorage).

This PR resets localStorage before each test manually.

It also bumps Cypress. This didn't fix the problem, but it's good to keep up to date.

## How to test

The integration tests should pass, reliably.

### DIFF
--- a/cypress/helpers/editor.ts
+++ b/cypress/helpers/editor.ts
@@ -6,6 +6,16 @@ import {
 import { elementWrapperTestId } from "../../src/renderers/react/ElementWrapper";
 import { getFieldViewTestId } from "../../src/renderers/react/FieldView";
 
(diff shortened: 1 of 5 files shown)


dskamiotis (approved):
This seems to solved the  9 intermittent failures I was getting. Ran local test via Cypress; all seem to pass successfully for imageElement, elementWrapper and editor.

SHession (approved):
Ran Cypress tests locally a few time, looks good. Also nice to see the Cypress bump, might make Cypress a little nicer to work with(?)

PR #57, opened 2021-07-15
Add command to bundle library for distribution

jonathonherbert:
_co-authored-by: @rhystmills_ 

## What does this change?

Adds a new script, `build`, which creates CommonJS and ES6 bundles and Typescript declarations in the `dist` folder, and amends `package.json` to point library consumers to the right files.

This should make the project fit for a 0.1.0 release 🎉 

## How to test

Run `yarn` to ensure deps are up-to-date, and then `yarn build`. You should see files in the `dist` folder corresponding to CommonJS, ES6 and declaration files.

(We also generate declarations for `demo`. This is redundant, but b/c we only refer to the declaration file in `src` in our package, it shouldn't matter, and it'd take a reshuffle of our tsconfigs to sort, so I think we're OK.)

Run `yarn link` in this repository, and then `yarn link prosemirror-elements` in another project in which you're importing it. Attempt to import a function from the library, e.g. 

```import { buildElementPlugin } from 'prosemirror-elements'```

Does your IDE recognise the library typings? Does your consumer build successfully bundle and run the library code?

Our library can be successfully released and imported by consumers.

### CODE REVIEW

**Comment by jonathonherbert** [comment]
@SHession thanks for checking this out out-of-band – it looks like webpack has trouble with module output at present (see this [long issue](https://github.com/webpack/webpack/issues/2933) for WIP), so I've reverted to tsc for both sorts of builds for now. Arguably this approach is preferable, as it's consistent with the way we [package things elsewhere](https://github.com/guardian/pkgu).

This should now be ready for another test/review.

**Comment by jonathonherbert** [comment]
@SHession I've updated the readme to cover our problems with `yarn link`.

File: src/editorial-source-components/TextInput.tsx [hunk 1]
```diff
@@ -1,10 +1,32 @@
...
 import { space } from "@guardian/src-foundations";
 import { textSans } from "@guardian/src-foundations/typography";
+import type { Props } from "@guardian/src-helpers";
+import type { Width } from "@guardian/src-text-input";
 import { TextInput as OriginalTextInput } from "@guardian/src-text-input";
+import type { InputHTMLAttributes } from "react";
 import { inputBorder } from "./inputBorder";
 
-export const TextInput = styled(OriginalTextInput)`
+interface TextInputProps extends InputHTMLAttributes<HTMLInputElement>, Props {
```
SHession, line 13: A bit unclear on how the changes in this file relate to the PR.

SHession (approved):
Changes look good and successfully tested linking the project to PM-Typerighter as an experiment. 

I'm not sure if the demo folder should be part of src, it doesn't seem like something we would want to include in dist unless there is a good reason for it?

src/editorial-source-components/TextInput.tsx
jonathonherbert: That's fair, I'll add a comment. Because `TextInputProps` aren't directly exported by Source, we can't generate a definition file with them. Instead, we have to define them manually. It's a bit rubbish.

jonathonherbert:
Good point re: demo, I was shuffling things about to test and neglected to put it back.

PR #58, opened 2021-07-15
chore: add semantic release

jonathonherbert:
## What does this change?

Adds [semantic-release](https://github.com/semantic-release/semantic-release) via a workflow and [github action](https://github.com/marketplace/actions/action-for-semantic-release). This means automated, semantically versioned releases with nicely formatted release histories – see [prosemirror-typerighter](https://github.com/guardian/prosemirror-typerighter/releases) for an example of a release history generated with this approach.

In order for semantic release to determine the correct version number, the project requires pull request names to be compliant with the [commitizen](https://github.com/commitizen-tools/commitizen) format: for example, `fix: a bug that was squashed` or `feat: an exciting new thing`.

This PR includes a github action that will validate the name of the PR in CI to ensure it's correctly formatted.

Comments on this approach really welcome.

## How to test

Does the github action run the correct code, and pass without errors?

Does the readme accurately reflect the new process? Is there anything we should add/remove?

### DIFF
--- a/.github/workflows/node-js.yml
+++ b/.github/workflows/node-js.yml
@@ -16,3 +16,18 @@ jobs:
           node-version: ${{ steps.nvm.outputs.NODE_VERSION }}
       - run: yarn install
       - run: yarn ci
+      - run: yarn build
+      - name: Semantic Release
+        uses: cycjimmy/semantic-release-action@v2
+        with:
+          branches: |
+            [
+              'main',
+              {
+                name: 'beta',
+                prerelease: true
+              }
+            ]
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/semantic-pr.yml
+++ b/.github/workflows/semantic-pr.yml
@@ -1,0 +1,16 @@
+
+name: 'Lint PR'
+on:
+    pull_request:
+        types:
+            - opened
+            - edited
+            - synchronize
+
+jobs:
+    main:
+        runs-on: ubuntu-latest
+        steps:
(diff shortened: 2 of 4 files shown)


SHession (approved):
This PR looks sensible. 

One small nit is that, I'm not a big fan of enforcing squash commits but AFAIK there isn't currently a way to avoid using them.

akash1810:
> One small nit is that, I'm not a big fan of enforcing squash commits but AFAIK there isn't currently a way to avoid using them.

FWIW @guardian/cdk uses semantic-release and does not enforce squashing. From our experience, so long as each commit in the change follows the [Angular](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) or [Karma](http://karma-runner.github.io/6.1/dev/git-commit-msg.html) format, semantic-release is happy. (Our docs are a little out of date too! https://github.com/guardian/cdk/pull/705)

jonathonherbert:
@akash1810 re:  _From our experience, so long as each commit in the change follows the Angular or Karma format, semantic-release is happy. (Our docs are a little out of date too! guardian/cdk#705)_

I think semantic-release will just [ignore poorly-formed commits](https://github.com/semantic-release/semantic-release/blob/master/docs/support/troubleshooting.md#squashed-commits-are-ignored-by-semantic-release) – so IIUC it shouldn't matter if it's just the merge commit that is well-formed.

jonathonherbert:
@akash1810 @SHession – just FYI, re: the above: _> I think semantic-release will just ignore poorly-formed commits – so IIUC it shouldn't matter if it's just the merge commit that is well-formed._

This is true, but the merge commit will never be well formed! For example, the commit for this PR was

```
Merge pull request #58 from guardian/jsh/add-sem-rel
    
chore: add semantic release
```

... which doesn't parse.

So it looks like squash commits are the only way to do this sensibly at the PR stage at present.

PR #59, opened 2021-07-19
docs: Adds vision document

SHession:
## What does this change?

This PR adds an initial version of our ProseMirror-Elements vision document. It provides a clear set of challenges, ambition and guiding principals which can be used to help orient the direction of the project.

When key decisions need to be made, we can refer to this document to see how the different options compare to our guiding principals. It should also provide a clear definition of the problem we are aiming to solve and why.

### DIFF
--- a/docs/vision.md
+++ b/docs/vision.md
@@ -1,0 +1,59 @@
+# Challenges
+
+1. Nesting arbitrary content within ProseMirror 
+2. Seamlessly Integrating nested ProseMirror rich-text fields with the parent instance
+3. Providing a way to create any possible custom nested element
+4. Abstracting ProseMirror core functionality and  offering a simple interface
+
+# Vision
+
+The goal of the ProseMirror-Elements plugin is to provide a logical extension to ProseMirror's core functionality that makes modeling arbitrary nested content simple. It should be easy to understand, easy to implement and agnostic about the elements that use it.
+
+## Easy to understand
+
+Creating new elements with the plugin should require minimal understanding of the ProseMirror ecosystem and the plugin itself. It should abstract the complexity of adding nested content and provide developers with a clear intuitive interface.
+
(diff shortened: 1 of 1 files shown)


docs/vision.md
```diff
@@ -0,0 +1,59 @@
+# Challenges
+
+1. Nesting arbitrary content within ProseMirror 
+2. Seamlessly Integrating nested ProseMirror rich-text fields with the parent instance
```
paperboyo, line 4:
```suggestion
2. Seamlessly integrating nested ProseMirror rich-text fields with the parent instance
```

docs/vision.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+## Easy to implement
+
+Adding the plugin to a new or existing project should be seamless and quickly show results. Once users have set the plugin up, they should be able to easily develop and add new elements.
+
+#### KPIs
+
+- Possible to implement into a ProseMirror instance quickly
+- Creating new elements can be
```
paperboyo, line 33: can be… great?

SHession: Looks like I was distracted mid-

docs/vision.md
```diff
@@ -0,0 +1,58 @@
...
+
+#### KPIs
+
+- Possible to implement into a ProseMirror instance quickly
+- Creating new elements can be 
+- Support different requirements
+
+## Agnostic
+
+The plugin should have no opinion about the elements that use it. Any complexity necessary to implement an element should be possible without increasing the surface area of the API.
```
paperboyo, line 38: > Any complexity necessary (…) should be possible
No complexity should be possible! 😜

jonathonherbert (approved):
Reviewed as a mob w/ @dskamiotis, @SHession and @rhystmills 👍

PR #61, opened 2021-07-23
Change element creation signature from array to object

jonathonherbert:
## What does this change?

Changes the element creation from array to object.

Before: 

```ts
buildElementPlugin([
  createImageElement("otherElement", onSelectImage, onCropImage),
  createOtherElement("otherElement")
]);
```

After:

```ts
buildElementPlugin({
  imageElement: createImageElement(onSelectImage, onCropImage),
  otherElement: createOtherElement()
});
```

### Why?

Two reasons – the main one, to fix a bug in our typesafety, and a side-effect, ergonomics.

#### Typesafety

At the moment, it's either difficult or impossible to correctly maintain type information about individual elements when they're passed in via an array type.

As a result, when more than one element is used via `buildElementPlugin`, the compiler assumes that each element is identical to the first. If it's not, there's a TypeError. This didn't show up until @SHession and @dskamiotis attempted to add a second element to an elements plugin.

As I understand it, the difficulty stems from the fact that Typescript cannot map into arbitrary arrays in the way that [Mapped Types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) map into objects. Mapped tuple types can't help us here – we not only need to infer the correct type for the array as a whole, we need to index into individual properties via their `name` typesafely.

Passing our fields as an object, and preserving the type information from the object keys via a mapped type, allows us to index into that object typesafely, and fixes the bug.

#### Ergonomics

This may be a post-hoc justification for a needed change, but I think this is a better API in any case:
- Names aren't really a necessary concept when creating elements themselves – they only make sense in the context of a mapping that's specific to their use in a plugin.
- This lets us reuse elements when the same element is required for different fields:
```ts
const rteElement = createRTEElement(commonSettings)

buildElementPlugin({
  field1: rteElement,
  field2: rteElement
});
```

### Other notes

This removes the `name` property, previously parameterised, from our `ElementSpec`, as it's only used when generating our `NodeSpec` output (which is generic, anyhow). This simplifies our type signature for `ElementSpec` a bit, which is nice.

## How to test

- Does the new method make sense, ergonomically?
- Do the automated tests pass? Do they satisfy you that we're typesafe when building our plugins and instantiating elements?


dskamiotis:
> ## What does this change?
> Changes the element creation from array to object.
> 
> Before:
> 
> ```ts
> buildElementPlugin([
>   createImageElement("otherElement", onSelectImage, onCropImage),
>   createOtherElement("otherElement")
> ]);
> ```
> 
> After:
> 
> ```ts
> buildElementPlugin({
>   imageElement: createImageElement(onSelectImage, onCropImage),
>   otherElement: createOtherElement()
> });
> ```
> 
> ### Why?
> Two reasons – the main one, to fix a bug in our typesafety, and a side-effect, ergonomics.
> 
> #### Typesafety
> At the moment, it's either difficult or impossible to correctly maintain type information about individual elements when they're passed in via an array type.
> 
> As a result, when more than one element is used via `buildElementPlugin`, the compiler assumes that each element is identical to the first. If it's not, there's a TypeError. This didn't show up until @SHession and @dskamiotis attempted to add a second element to an elements plugin.
> 
> As I understand it, the difficulty stems from the fact that Typescript cannot map into arbitrary arrays in the way that [Mapped Types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) map into objects. Mapped tuple types can't help us here – we not only need to infer the correct type for the array as a whole, we need to index into individual properties via their `name` typesafely.
> 
> Passing our fields as an object, and preserving the type information from the object keys via a mapped type, allows us to index into that object typesafely, and fixes the bug.
> 
> #### Ergonomics
> This may be a post-hoc justification for a needed change, but I think this is a better API in any case:
> 
> * Names aren't really a necessary concept when creating elements themselves – they only make sense in the context of a mapping that's specific to their use in a plugin.
> * We shouldn't permit elements with duplicate names, which is no longer possible with this API
> * This lets us reuse elements when the same element is required for different fields:
> 
> ```ts
> const rteElement = createRTEElement(commonSettings)
> 
> buildElementPlugin({
>   field1: rteElement,
>   field2: rteElement
> });
> ```
> 
> ### Other notes
> This removes the `name` property, previously parameterised, from our `ElementSpec`, as it's only used when generating our `NodeSpec` output (which is generic, anyhow). This simplifies our type signature for `ElementSpec` a bit, which is nice.
> 
> ## How to test
> * Does the new method make sense, ergonomically?
> * Do the automated tests pass? Do they satisfy you that we're typesafe when building our plugins and instantiating elements?

Looks really good @jonathonherbert thanks for doing that on 10% day!
@jonathonherbert @SHession @rhystmills Just wondering if this merge will affect the Pullquotes and other elements created?

I am assuming this elegantly solves the issue of being able to build multiple elements  without the type restrictions/persistence. 
Just wondering if it's ok to pull and merge this into Pullquote branch? (also any recommendation re merging vs re-basing?)

jonathonherbert:
@dskamiotis good stuff! As we discussed out of band, I think it's probably best to hold fire this afternoon.

If you did want to get this (or the other PR that may be relevant, #62) merged next week, and were worried about merge conflicts with other, parallel work, you could always attempt to rebase this branch into your current work before approving and merging this PR. It might be a good task to pair on with another member of @guardian/digital-cms.

PR #62, opened 2021-07-23
feat: Add an `isMultiline` attribute to TextField

jonathonherbert:
## What does this change?

Adds an `isMultiline` attribute to the `TextField` field. This adds two things to that field:

- the ability to add line breaks to the text w/ the Enter key. The input will expand to accommodate new lines, represented as `<br>` elements in the output.
- the ability to define the initial height of the element via a `rows` option, analogous to the same attribute in a textarea.

## How to test

I've altered the altText field to demonstrate this new format, and for our integration tests to assert against it.

- Have a play. Does it behave as expected?
- Do the integration tests give you confidence this behaviour is well tested?


rhystmills (approved):
Having tested locally this looks to be working well 👍 

I did notice an existing bug - throwing an error when using the `enter` key as the last character in a RichTextField. It predates this PR, so no need to fix it here.

I also wonder if we could make it clearer to a user whether an input is single line plain text, multiline plain text, or a rich text editor - perhaps worth getting some UI advice on that.

PR #63, opened 2021-07-27
feat: Only show focus when the user is keyboard navigating, not clicking

rhystmills:
## What does this change?
This PR makes some small changes to focus style, so that the wide focus ring only shows to users who are keyboard navigating, *not* selecting fields with the mouse.

This is accomplished by adding the Source `FocusStyleManager`, and running `FocusStyleManager.onlyShowFocusOnTabs();`. this only needs to be done once per application - and may also need to be done in applications using our Source-styled Elements.

## How to test

Run `yarn start` locally and use the `Element`.

## Accessibility

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -1,3 +1,4 @@
+import { FocusStyleManager } from "@guardian/src-foundations/utils";
 import type OrderedMap from "orderedmap";
 import { collab } from "prosemirror-collab";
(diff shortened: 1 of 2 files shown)


dskamiotis (approved):
This working great @rhystmills 👍 
Working even with reverse tab going back up the stack

PR #65, opened 2021-07-30
feat: Create Embed Element

rhystmills:
_Co-authored with @dskamiotis_ 

## What does this change?
This PR adds an Embed Element replicating the existing Embed element in Composer.

| Existing element | New Prosemirror Embed Element |
| --- | --- |
| image | image |

This PR also:
- adds a CustomCheckbox which renders a modified [Source Checkbox](https://theguardian.design/2a1e5182b/p/466fad-checkbox/b/33fc2f).
- adds a monospace-styling option (`isCode`) for code fields, used here in the `embedCode` field.
- includes validation (as found on the original Element)

#### What does this _not_ do?
This does not yet add:
- Text field placeholders

## How to test

Run `yarn start`.

### DIFF
--- a/src/elements/embed/EmbedForm.tsx
+++ b/src/elements/embed/EmbedForm.tsx
@@ -23,7 +23,11 @@ export const EmbedElementForm: React.FunctionComponent<Props> = ({
   fieldViewSpecMap: fieldViewSpecs,
 }) => (
   <div data-cy={EmbedElementTestId}>
-    <CustomDropdownView fieldViewSpec={fieldViewSpecs.weighting} />
+    <CustomDropdownView
+      fieldViewSpec={fieldViewSpecs.weighting}
+      label="Weighting"
+      errors={errors.weighting}
+    />
     <FieldView fieldViewSpec={fieldViewSpecs.sourceUrl} />
     <FieldView fieldViewSpec={fieldViewSpecs.embedCode} />
     <FieldView fieldViewSpec={fieldViewSpecs.caption} />
--- a/src/elements/embed/EmbedForm.tsx
+++ b/src/elements/embed/EmbedForm.tsx
@@ -1,9 +1,9 @@
 import React from "react";
 import { Label } from "../../editorial-source-components/Label";
-import { CustomCheckboxView } from "../../plugin/fieldViews/CustomCheckboxView";
-import { CustomDropdownView } from "../../plugin/fieldViews/CustomDropdownView";
 import type { FieldNameToValueMap } from "../../plugin/fieldViews/helpers";
 import type { FieldNameToFieldViewSpec } from "../../plugin/types/Element";
+import { CustomCheckboxView } from "../../renderers/react/customFieldViewComponents/CustomCheckboxView";
+import { CustomDropdownView } from "../../renderers/react/customFieldViewComponents/CustomDropdownView";
 import { FieldView } from "../../renderers/react/FieldView";
 import type { createEmbedFields } from "./EmbedSpec";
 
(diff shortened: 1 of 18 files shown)


src/plugin/fieldViews/TextFieldView.ts
```diff
@@ -73,19 +79,25 @@ export class TextFieldView extends ProseMirrorFieldView {
...
+          );
 
-        (this.innerEditorView.dom as HTMLDivElement).style.minHeight = `${
-          parseInt(lineHeight, 10) * rows + parseInt(paddingTop) * 2
-        }px`;
+          (this.innerEditorView.dom as HTMLDivElement).style.minHeight = `${
+            parseInt(lineHeight, 10) * rows + parseInt(paddingTop) * 2
+          }px`;
+        }
+        if (isCode) {
```
jonathonherbert, line 97: We shouldn't have to wait a tick to apply the code styles – this is necessary in the case of `minHeight` because we must measure the element height to calculate the value.

As a result, it might be best to move the monospace styling out of this timeout, just to avoid confusion as to its purpose.

src/elements/embed/EmbedForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+}) => (
+  <div data-cy={EmbedElementTestId}>
+    <CustomDropdownView fieldViewSpec={fieldViewSpecs.weighting} />
+    <FieldView fieldViewSpec={fieldViewSpecs.sourceUrl} />
+    <FieldView fieldViewSpec={fieldViewSpecs.embedCode} />
+    <FieldView fieldViewSpec={fieldViewSpecs.caption} />
+    <FieldView fieldViewSpec={fieldViewSpecs.altText} />
+    <CustomCheckboxView fieldViewSpec={fieldViewSpecs.required} />
+
+    <hr />
```
jonathonherbert, line 33: We shouldn't need these json readouts for non-test elements, although I appreciate they might have been useful for debugging!

jonathonherbert (commented):
Couple of comments worth addressing before we merge, but this is great work.

src/elements/embed/EmbedForm.tsx
rhystmills: Good point!

src/plugin/fieldViews/TextFieldView.ts
rhystmills: Good point, I'm on it.

src/elements/embed/EmbedSpec.tsx
```diff
@@ -0,0 +1,56 @@
...
+import {
+  createValidator,
+  htmlMaxLength,
+  htmlRequired,
+  maxLength,
+} from "../../plugin/helpers/validation";
+import { createReactElementSpec } from "../../renderers/react/createReactElementSpec";
+import { EmbedElementForm } from "./EmbedForm";
+
+export const createEmbedFields = () => {
```
jonathonherbert, line 14: This shouldn't need to exist w/in a function for an embed element – the reason we do it for the imageElement is that the image select/crop functions are unique to the consumer.

So we should be fine to just export an object called `embedFields` and use that directly in the element spec.

jonathonherbert (approved):
Taken a look locally and validation etc. looks good – ace.

One minor change since those revisions that might be worth addressing, but this is now good to 🚢 !

jonathonherbert (approved):

rhystmills:
> Taken a look locally and validation etc. looks good – ace.
> 
> One minor change since those revisions that might be worth addressing, but this is now good to 🚢 !

Ah yes - implementing that suggestion now.

PR #67, opened 2021-07-30
feat: Create Code Element

dskamiotis:
co-authored with @rhystmills, @jonathonherbert 

## What does this change?
This PR adds a Code Element replicating the existing Embed element in Composer.

### Images
Before: 

 
After:

Screenshot 2021-08-05 at 16 34 36

This PR also:
- adds a `CustomDropDown` which renders a modified Source Dropdown (allowing for clearer displayed text and improved - arrow button).
- includes the use of `createTextField` as a text area allowing for multi line 
- adds a createTextField with multiline capabilities set with a default of 11 lines. for `codeText` to replicate Composer.
- The naming of the field `code` had to be amended to `codeText` due to browser error in rendering

#### What does this _not_ do?
This does not yet add:
- Deal with data output in flexible content: 
```json
{
  elementType: "code",
  fields: {
    html: "Example code",
    language: "javascript"
  },
  assets: [ ]
},
```

## How to test
Run `yarn start`


src/elements/code/CodeElementSpec.tsx
```diff
@@ -0,0 +1,32 @@
...
+    { text: "Ruby", value: "ruby" },
+    { text: "Scala", value: "scala" },
+    { text: "YAML", value: "yaml" },
+  ]),
+};
+
+export const codeElement = createReactElementSpec(
+  codeFields,
+  (_, errors, __, fieldViewSpecs) => {
+    return <CodeElementForm errors={errors} fieldViewSpecs={fieldViewSpecs} />;
```
dskamiotis, line 25: Much more efficient and robust way to pipe validation through our function. This can be easily transformed in other elelments.

src/elements/code/CodeElementForm.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const CodeElementForm: React.FunctionComponent<Props> = ({
+  errors,
+  fieldViewSpecs,
+}) => (
+  <div data-cy={CodeElementTestId}>
+    <Field
+      label="Code"
+      fieldViewSpec={fieldViewSpecs.codeText}
+      errors={errors.codeText}
```
dskamiotis, line 22: Makes a lot more sense to pass errors as a prop 👍

demo/index.ts
```diff
@@ -7,7 +7,8 @@ import { Schema } from "prosemirror-model";
 import { schema as basicSchema } from "prosemirror-schema-basic";
 import { EditorState } from "prosemirror-state";
 import { EditorView } from "prosemirror-view";
-import { createImageElement } from "../src/elements/demo-image/DemoImageElement";
+import { createImageElement } from "../src";
+import { codeElement } from "../src/elements/code/CodeElementSpec";
```
SHession, line 11: We'll need to include `codeElement` as an export of src when the time comes to add the element to Composer.

src/elements/code/CodeElementSpec.tsx
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import { createCustomField } from "../../plugin/fieldViews/CustomFieldView";
+import { createTextField } from "../../plugin/fieldViews/TextFieldView";
+import { createValidator, required } from "../../plugin/helpers/validation";
+import { createReactElementSpec } from "../../renderers/react/createReactElementSpec";
+import { CodeElementForm } from "./CodeElementForm";
+
+export const codeFields = {
+  codeText: createTextField({ isMultiline: true, rows: 11 }, true),
+  language: createCustomField("Plain text", [
```
SHession, line 10: The default here should likely be "text" rather than "Plain text".

src/elements/code/CodeElementSpec.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const codeElement = createReactElementSpec(
+  codeFields,
+  (_, errors, __, fieldViewSpecs) => {
+    return <CodeElementForm errors={errors} fieldViewSpecs={fieldViewSpecs} />;
+  },
+  createValidator({ codeText: [required()] }),
+  {
+    codeText: "",
+    language: "Plain text",
```
SHession, line 30: Same here as the above comment.

SHession (approved):
Spec and element display look good.

I've noticed a small difference between this and the current code element. It looks like the new one is outputting `<br>`s  and the old one is outputting `\n`s. This is probably something to be handled by ingress and egress rules but may be worth thinking about before we consider this ready for Composer.

src/elements/demo-image/DemoImageElementForm.tsx
```diff
@@ -60,7 +60,13 @@ export const ImageElementForm: React.FunctionComponent<Props> = ({
       fieldViewSpec={fieldViewSpecs.optionDropdown}
       errors={errors.optionDropdown}
     />
-    <ImageView fieldViewSpec={fieldViewSpecs.mainImage} />
+    <ImageView
+      fieldViewSpec={fieldViewSpecs.mainImage}
+      onChange={(_, __, ___, description) => {
+        fieldViewSpecs.altText.update(description);
+        fieldViewSpecs.caption.update(description);
```
dskamiotis: Does this alter the way in which errors are handled? As my branch is rebased from this one.

jonathonherbert: I don't think we have validation for the `ImageView` component at the moment – is that what you're referring to here? I may have misunderstood, but I don't think that this should affect your work.

SHession (approved):
Solution to the whitespace issue looks good and the element behaves well locally. 

I've noticed the required field validation reports an issue when you refresh the page even if there is content, this may be an existing issue.

PR #68, opened 2021-08-03
Namespace field nodes with their element type, and add tests

jonathonherbert:
## What does this change?

Namespaces nodes representing fields with their element type. This prevents different elements with identical field names conflicting, which is currently blocking #65. 

## How to test

- Launch the demo and examine the Prosemirror schema in the devtools. Each node type should be clearly namespaced with the field name:

Screenshot 2021-08-03 at 13 20 00


rhystmills (approved):
Tested locally and all looks to be working well to me 👍 

Also tested as a base for the [EmbedElement](https://github.com/guardian/prosemirror-elements/pull/65) branch and the namespace clash is no longer occurring there.

PR #69, opened 2021-08-04
feat: add validation helpers

jonathonherbert:
_co-authored-with: @dskamiotis, @rhystmills_  

## What does this change?

Adds a validation helper function that allows the user to compose validators against fields. This should make common validation tasks easier to express.

## How to test

- The unit tests should pass. The new unit tests should adequately demonstrate the helper's behaviour.
- The validation applied to the demo element should work as expected – take a look at the `DemoImageElement` declaration, and compare it to the runtime behaviour.

We have an easier time expressing validation with new elements.

### DIFF
--- a/src/plugin/helpers/validation.ts
+++ b/src/plugin/helpers/validation.ts
@@ -3,7 +3,7 @@ import type { FieldSpec } from "../types/Element";
 
 type Validator = (fieldValue: unknown) => string[];
 
-export const buildValidator = (
+export const createValidator = (
   fieldValidationMap: Record<string, Validator[]>
 ) => <FSpec extends FieldSpec<string>>(
   fieldValues: FieldNameToValueMap<FSpec>
--- a/src/plugin/helpers/validation.ts
+++ b/src/plugin/helpers/validation.ts
@@ -1,0 +1,46 @@
+import type { FieldNameToValueMap } from "../fieldViews/helpers";
+import type { FieldSpec } from "../types/Element";
+
+type Validator = (fieldValue: string) => string[];
+
+export const buildValidator = (
+  fieldValidationMap: Record<string, Validator[]>
+) => <FSpec extends FieldSpec<string>>(
+  fieldValues: FieldNameToValueMap<FSpec>
+): Record<string, string[]> => {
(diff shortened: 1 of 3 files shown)


dskamiotis:
Integration and unit tests along with yarn start working successfully 👍

src/plugin/helpers/validation.ts
```diff
@@ -0,0 +1,64 @@
...
+  for (const fieldName in fieldValidationMap) {
+    const validators = fieldValidationMap[fieldName];
+    const value = fieldValues[fieldName];
+    const fieldErrors = validators.flatMap((validator) => validator(value));
+    errors[fieldName] = fieldErrors;
+  }
+  return errors;
+};
+
+export const htmlMaxLength = (maxLength: number): Validator => (value) => {
```
dskamiotis, line 22: It would be good to update the HtmlRequired but as you said given the type issues this is a really good middle ground for us to get the validation ball rolling and working well.

dskamiotis (approved):

src/plugin/helpers/validation.ts
```diff
@@ -0,0 +1,64 @@
+import type { FieldNameToValueMap } from "../fieldViews/helpers";
+import type { FieldSpec } from "../types/Element";
+
+type Validator = (fieldValue: unknown) => string[];
+
+export const buildValidator = (
```
rhystmills, line 6: This is a _very_ minor naming suggestion, but maybe `createValidator` would be a better name here - I always find I have to parse 'build' mentally as a noun vs. verb - does this _validate_ a _build_, or does it _build_ a _validator_? Whereas 'create' is unambiguously a verb.

rhystmills (approved):
This looks like a great and intuitive interface that encapsulates validation nicely. I had a very pedantic naming suggestion but this all looks good to me.

src/plugin/helpers/validation.ts
jonathonherbert: Completely agree, great suggestion – done. I think we use 'build' elsewhere in this project – can I suggest raising another PR to change those names, too?

PR #70, opened 2021-08-04
feat: spruce validation UI

jonathonherbert:
## What does this change?

Gives us some common components to express error messages in the UI.

I haven't altered our custom components, as I'm aware that there may be changes needed downstream, but hopefully the `Field` component contains a template to work from.

Before:

Screenshot 2021-08-04 at 17 45 54

After (note more readable labels, validation message and highlighting, consistent margins):

Screenshot 2021-08-04 at 17 47 05

## How to test

Do the changes look OK? Do the components look reusable for other elements?

## Accessibility

- [x] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398) – this wasn't easy, are there ways to make error-reporting clearer to screen readers?
- [x] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f) – **interesting, can we express 'error-ness' another way here?**

### DIFF
--- a/src/elements/demo-image/DemoImageElementForm.tsx
+++ b/src/elements/demo-image/DemoImageElementForm.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import { CustomDropdown } from "../../editorial-source-components/CustomDropdown";
+import { Field } from "../../editorial-source-components/Field";
 import { Label } from "../../editorial-source-components/Label";
 import type { Option } from "../../plugin/fieldViews/DropdownFieldView";
 import type { FieldNameToValueMap } from "../../plugin/fieldViews/helpers";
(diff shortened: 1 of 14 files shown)


dskamiotis:
@jonathonherbert, this looks great Im really pleased with the non over complicated CSS as well allowing for simple  future addition/amendments 🎆

src/elements/demo-image/DemoImageElementForm.tsx
```diff
@@ -24,14 +25,30 @@ export const ImageElementTestId = "ImageElement";
...
-  fieldViewSpecMap: fieldViewSpecs,
+  fieldViewSpecs,
 }) => (
   <div data-cy={ImageElementTestId}>
-    <FieldView fieldViewSpec={fieldViewSpecs.caption} />
-    <FieldView fieldViewSpec={fieldViewSpecs.altText} />
-    <FieldView fieldViewSpec={fieldViewSpecs.src} />
-    <FieldView fieldViewSpec={fieldViewSpecs.useSrc} />
-    <FieldView fieldViewSpec={fieldViewSpecs.optionDropdown} />
+    <Field
```
dskamiotis, line 31: This makes it supper easy to implement across the in task elements once we re base. Great re usable components

dskamiotis (approved):

rhystmills (approved):
Styling looks great to me, and code all structured intuitively 👍 

I would say we are good on the 'colour to convey meaning' accessibility question as we are also including a text prompt (i.e. 'Required'). 

I think using colour as an additional layer of communication is always a good thing - it's just best not to use it in isolation to convey meaning.

PR #71, opened 2021-08-05
feat: Add API to programmatically update fields

jonathonherbert:
_co-authored-by: @dskamiotis_ 

## What does this change?

Adds an function to allow us to programmatically update fields to the `FieldViewSpec` type:

```ts
<button
  data-cy={UpdateAltTextButtonId}
  onClick={() => fieldViewSpecs.altText.update("Default alt text")}
>
```

The function is typesafe, and should only allow us to pass values that match the type handled by the field.

See `DemoImageElement.tsx`, where it's used in the example above, and to programmatically update the `altText` and `caption` fields when we receive a new image.

## How to test

- The new integration test should pass, and you should be convinced it's testing a programmatic update.
- Add a new image to the demo image element. Does that operation also update the altText and caption fields with the description of that image?

## Dev notes

In order to keep our types as simple as possible, I've removed the type parameter from the `DropdownFieldView` fieldview – it meant we had to use conditional types in our `Field` component, which meant that writing the type signatures was more complex.

At the moment, we don't have an example of a dropdown field that doesn't have a `string` value – we actually don't use this component at all, preferring to model dropdowns via a `CustomFieldView` and a React component which uses Source. Me and @SHession have discussed this tradeoff before.

My current take is that this isn't worth maintaining or adding complexity for whilst there's not a use case for it, hence the removal, but if there's a strong argument for its continued inclusion we can definitely revisit this decision!

### DIFF
--- a/src/plugin/types/Element.ts
+++ b/src/plugin/types/Element.ts
@@ -11,6 +11,7 @@ import type {
   DropdownField,
   DropdownFieldView,
 } from "../fieldViews/DropdownFieldView";
+import type { FieldView } from "../fieldViews/FieldView";
 import type {
   FieldNameToValueMap,
   FieldTypeToViewMap,
@@ -45,20 +46,19 @@ export type FieldViews =
 export type NonCustomFieldViews =
   | TextFieldView
   | RichTextFieldView
-  | CheckboxFieldView
-  | DropdownFieldView;
+  | CheckboxFieldView;
 
-export type FieldViewSpec<FieldView extends FieldViews> = {
-  fieldView: FieldView;
+export interface FieldViewSpec<F> {
+  fieldView: F;
   fieldSpec: Field;
   name: string;
-};
+  update: (value: F extends FieldView<infer Value> ? Value : never) => void;
+}
 
-export type CustomFieldViewSpec<Data = unknown, Props = unknown> = {
-  fieldView: CustomFieldView<Data>;
+export interface CustomFieldViewSpec<Data = unknown, Props = unknown>
+  extends FieldViewSpec<CustomFieldView<Data>> {
   fieldSpec: CustomField<Data, Props>;
-  name: string;
-};
+}
 
 export type FieldNameToFieldViewSpec<FSpec extends FieldSpec<string>> = {
   [name in Extract<keyof FSpec, string>]: FSpec[name] extends CustomField<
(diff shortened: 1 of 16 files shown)


src/plugin/types/Element.ts
```diff
@@ -45,20 +46,19 @@ export type FieldViews =
 export type NonCustomFieldViews =
   | TextFieldView
   | RichTextFieldView
-  | CheckboxFieldView
-  | DropdownFieldView;
```
rhystmills, line 49: I'm not totally sure why the DropdownFieldView has been removed here - was this intentional?

rhystmills (approved):
This PR looks good to me. The updating interface is simple and intuitive, and the simplifications to the `DropdownFieldView` also make sense 👍  (the developer is always free to encode non-string data types as JSON and use that as the value).

I spotted one change which I didn't fully understand - removing `DropdownFieldView` from the `NonCustomFieldViews` type. Otherwise everything looks great!

src/plugin/types/Element.ts
jonathonherbert: Good spot. I think this was removed when DropdownFieldView still had a generic, but as I've removed the generic, it can now return. Thanks!

jonathonherbert: In fact, it's no longer needed at all as a result of that change – even better.

SHession:
> At the moment, we don't have an example of a dropdown field that doesn't have a string value

This isn't strictly true, in Composer weightings are currently modeled as `string | undefined` in a way I'm a bit unfamiliar with.

image

jonathonherbert:
> > At the moment, we don't have an example of a dropdown field that doesn't have a string value
> 
> This isn't strictly true, in Composer weightings are currently modeled as `string | undefined` in a way I'm a bit unfamiliar with.
> 
> image

Discussed out of band – at the moment we model dropdown fields w/ `CustomField`s to let us use Source. @SHession mentioned some concerns over ergonomics here, which are definitely justified – we think we can provide abstractions to make this a very similar experience to the 'native' dropdown fieldviews (for example, by providing a `createDropdownField` function that produces a `CustomField` with the right data shape).

Ultimately we may be a bit stuck with these sorts of solutions, as we think consumers will almost always want their own ways to render these sorts of fields – but if there's a clear need to a simpler solution, perhaps there's another way to discover further down the line.

PR #72, opened 2021-08-05
chore: Consistent type parameters for FieldViews

jonathonherbert:
## What does this change?

A quick PR to make the type parameter names for FieldViews more consistent (`Fields`|`CustomFieldValue`|`Value` -> `Value`).

## How to test

Did I catch 'em all?

### DIFF
--- a/src/plugin/fieldViews/AttributeFieldView.ts
+++ b/src/plugin/fieldViews/AttributeFieldView.ts
@@ -6,8 +6,8 @@ import { FieldType } from "./FieldView";
 /**
  * A FieldView representing a node that contains fields that are updated atomically.
  */
-export abstract class AttributeFieldView<Fields extends unknown>
-  implements FieldView<Fields> {
+export abstract class AttributeFieldView<Value extends unknown>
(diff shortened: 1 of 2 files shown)


src/plugin/fieldViews/AttributeFieldView.ts
```diff
@@ -6,8 +6,8 @@ import { FieldType } from "./FieldView";
 /**
  * A FieldView representing a node that contains fields that are updated atomically.
  */
-export abstract class AttributeFieldView<Fields extends unknown>
-  implements FieldView<Fields> {
+export abstract class AttributeFieldView<Value extends unknown>
+  implements FieldView<Value> {
```
dskamiotis, line 10: I think this makes it much clearer to devs working on this. The use of field is used widely and specifying its `value` makes a lot of sense.

dskamiotis:
> Did I catch 'em all?
It seems on my inspection you have indeed, all tests and localhost working fine too 👍

dskamiotis (approved):

PR #73, opened 2021-08-05
feat: Add an easy way to get data in and out of nodes

jonathonherbert:
## What does this change?

Adds two methods created by `buildElementPlugin`, `getNodeFromElementData` and `getElementDataFromNode` .

These functions work as you'd expect:

- `getElementDataFromNode`, given a `Node` of the right shape,  returns its data representation, or `undefined`
- `getNodeFromElementData` returns a `Node` or `undefined` from element data.

This will be handy in Composer, as we translate between our JSON element model and a Prosemirror node when we create a document, and vice-versa when we save.

## How to test

Examine the unit tests written for this case. Do they convince you that these functions work? Are there any other tests you think we could write to ensure this is working correctly?

## Notes

In order to make this change as typesafe as possible, I've altered the signature for `insertElement` to match the signature for element data in these functions, which looks like 

```ts
{
  elementName: ElementName,
  values: ElementValues
}
```

This makes our job typing these arguments a little easier, as our element data can now be represented as a union of these objects, where each member is specific to an element. This is done with a new type helper, `ExtractDataTypeFromElementSpec`, which uses a [distributed conditional type to produce the right shape](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#distributive-conditional-types).

For example, with two elements, Element1 and Element2, we end up with an type that looks like

```ts
{
  elementName: Element1Name,
  values: Element1Values
} | {
  elementName: Element2Name,
  values: Element2Values
}
```

Without this change, it's difficult to avoid a type that looks like 

```ts
{
  elementName: Element1Name | Element2Name,
  values: Element1Values | Element2Values
}
```

... which means that consumers could mix up Element1's name with Element2's values, and vice versa.


rhystmills (approved):
This all looks good to me, and tests look sensible 👍  

Tests are all passing as expected, and application is also running as intended on Firefox and Chrome.

I've added a couple of comments where there are minor typos in test descriptions.

PR #75, opened 2021-08-11
feat: Expand rich text options to accommodate a modified NodeSpec 

jonathonherbert:
_> co-authored-by: @rhystmills_ 

## What does this change?

Expands the options for rich text options to accommodate a modified `NodeSpec`. This can be useful when e.g. restricting marks, or specifying different kinds of content.

The `createRichTextField` field creator signature has been modified to allow these options to be passed.

We think there's room for a new field creator, alongside `createDefaultRichTextField`, to create a field that does not include paragraphs – perhaps `createFlatRichTextField`. This would reduce some repetition when declaring fields – for example, all of our captions across different elements are likely to use these defaults.

This would also need to include a binding for the `enter` key that inserts a hard_break (as with the multiline text input)  – but that can be added in a separate PR.

## How to test

-  Have a play with the 'restricted text'. It should only permit italic marks, and only shift+enter should work (to insert line breaks).

### DIFF
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -88,6 +88,28 @@ describe("ImageElement", () => {
         });
       });
 
+      it(`restrictedTextField – can toggle italic style of an input in an element`, () => {
+        addImageElement();
+        getElementMenuButton("restrictedTextField", "Toggle emphasis").click();
+        typeIntoElementField("restrictedTextField", "Example text");
+        getElementRichTextField("restrictedTextField")
+          .find("em")
+          .should("have.text", "Example text");
+      });
+
+      it(`restrictedTextField – can't toggle strong style of an input in an element`, () => {
+        addImageElement();
+        getElementMenuButton(
+          "restrictedTextField",
+          "Toggle strong style"
+        ).click();
+        typeIntoElementField("restrictedTextField", "Example text");
+        getElementRichTextField("restrictedTextField").should(
+          "not.contain.html",
+          "<strong>"
+        );
+      });
+
       it("should serialise content as HTML within the appropriate nodes in the document", () => {
         addImageElement();
         typeIntoElementField("caption", "Caption text");
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -2,7 +2,7 @@ import { FocusStyleManager } from "@guardian/src-foundations/utils";
 import type OrderedMap from "orderedmap";
 import { collab } from "prosemirror-collab";
 import { exampleSetup } from "prosemirror-example-setup";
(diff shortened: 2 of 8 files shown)


cypress/tests/ImageElement.spec.ts
```diff
@@ -88,6 +88,28 @@ describe("ImageElement", () => {
         });
       });
 
+      it(`restrictedTextField – can toggle italic style of an input in an element`, () => {
```
dskamiotis, line 91: I think we will eventually change the file name to DemoImageElement.

demo/index.ts
```diff
@@ -51,8 +51,20 @@ const {
   pullquoteElement,
 });
 
+const restrictedParagraph = {
+  content: "text*",
+  marks: "em",
```
SHession, line 56: What is this additional node spec being used for? It looks superfluous to me,  unless I've missed something.

src/elements/demo-image/DemoImageElement.tsx
```diff
@@ -37,6 +41,13 @@ export const createImageFields = (
 ) => {
   return {
     caption: createDefaultRichTextField(),
+    restrictedTextField: createRichTextField({
+      createPlugins: (schema) => exampleSetup({ schema }),
+      nodeSpec: {
+        content: "(text|hard_break)*",
+        marks: "em",
+      },
```
SHession, line 49: This a great concise API, must be inspired by the vision document 👀

SHession (approved):
Tested locally and reviewed the code, looks great. Slight unclear on why the additional node spec is needed, I'm possibly missing something.

demo/index.ts
jonathonherbert: Great catch – this was used when me and @rhystmills were proving the concept. It's no longer needed.

jonathonherbert:
Thanks @SHession!

Re: _>  if we ever require something that will need to be part of the sub-RTE but not the parent_ 

The parent schema must include all possible node types, including all types used by the child – but it should be possible to add node types that are _only_ used by the sub-RTE, by ensuring that only the node representing the sub-RTE has a content expression that includes that node type, if that makes sense?

SHession:
> it should be possible to add node types that are _only_ used by the sub-RTE, by ensuring that only the node representing the sub-RTE has a content expression that includes that node type, if that makes sense?

That makes sense, I think I was more envisioning issues with the menu items display, which might have to be adjusted to account for this. Either way, should all be solvable if we need to.

jonathonherbert:
Agreed, and the menu question is really a question for the consumer (via a plugin) – we only include a menu plugin here for demo purposes.

PR #76, opened 2021-08-11
feat: Add `createFlatRichTextField` field creator

jonathonherbert:
## What does this change?

Adds a `createFlatRichTextField` field creator. (It's not a great name, eh. Can we think of a better one? `createNoParaRichTextField`?)

It creates a rich text field that doesn't use paragraphs – instead, it produces flat text, with `br` tags for line breaks.

To do this, the field adds two pieces of configuration –
- adds a keymapping from `Enter` that adds a `br` tag
- specifies the content for the node to be `(text|hard_break)*`

The user is free to add plugins or alter other parts of the `NodeSpec` themselves.

## How to test

This effectively duplicates the testable functionality of #75 in a helper function. I've added a test to check that the enter key works.

Try the field out in the demo element. Does it work as expected? Does the output match what we describe above?

### DIFF
--- a/src/plugin/fieldViews/RichTextFieldView.ts
+++ b/src/plugin/fieldViews/RichTextFieldView.ts
@@ -2,7 +2,7 @@ import { exampleSetup } from "prosemirror-example-setup";
 import { redo, undo } from "prosemirror-history";
 import { keymap } from "prosemirror-keymap";
 import type { Node, NodeSpec, Schema } from "prosemirror-model";
-import type { Plugin } from "prosemirror-state";
+import type { EditorState, Plugin, Transaction } from "prosemirror-state";
 import type { Decoration, DecorationSet, EditorView } from "prosemirror-view";
 import type { BaseFieldSpec } from "./FieldView";
 import { ProseMirrorFieldView } from "./ProseMirrorFieldView";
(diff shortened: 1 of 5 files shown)


src/plugin/fieldViews/RichTextFieldView.ts
```diff
@@ -27,6 +27,50 @@ export const createRichTextField = ({
...
+          return true;
+        },
+      };
+
+      const plugin = keymap(keymapping);
+      return [plugin, ...(createPlugins?.(schema) ?? [])];
+    },
+    nodeSpec: {
+      ...nodeSpec,
+      content: "(text|hard_break)*",
```
SHession, line 66: Could this have issue if we ever wanted other kinds of content in a flat field, i.e. lists? Perhaps this never happens but I'm not sure if that is made explicit from the name i.e. the name should make clear it can only have text and breaks. Naming is difficult!

Also the fact you can specify a content property that will then be ignored could perhaps be a bit confusing.

SHession (approved):
Change looks good. I found a slight "bug" when copying content from another field with <p> tag soft breaks(?), having all the breaks removed.

src/plugin/fieldViews/RichTextFieldView.ts
jonathonherbert: It's flat! No-one is allowed a list 😁  I dislike the name also, but can't think of anything else at present – suggestions welcome, and we can easily change it down the line if we can't think of anything ATM.

_> Also the fact you can specify a content property that will then be ignored could perhaps be a bit confusing._

The type for the `RichTextOptions`, `Partial<Omit<NodeSpec, "content">>`, removes the `content` type from the nodeSpec, so you shouldn't be able to specify a content property when calling `createFlatRichTextField` (for the reason you mention.)

jonathonherbert:
Re: _> "bug" when copying content from another field_

This coincides with the current behaviour of the (Scribe-based) image caption field – I'm guessing that in both cases, the parser in question is seeing tags it doesn't understand (`p` tags) and simply discarding them, which has the effect of concatenating all the content.

We could write a parser to translate `p` to `br` tags on paste if this is problematic in the longer term.

src/plugin/fieldViews/RichTextFieldView.ts
SHession: > removes the content type from the nodeSpec

Oh I missed that, looks good.

> we can easily change it down the line if we can't think of anything ATM.

Rationale on name makes sense, let's go with that 👍

PR #77, opened 2021-08-11
fix: Remove element defaults

jonathonherbert:
## What does this change?

Removes our element-level defaults. We ... didn't really use them, except (incorrectly) on initialisation! They were a hangover from a previous design, and there weren't any tests to verify they had an effect.

This fixes an incidental bug where validation errors were not appearing when nodes were initialised on doc serialisation (card [here](https://trello.com/c/ygmh8P87/718-resolve-validation-issue)).

## How to test

- The automated tests should pass, and they should satisfy you that defaults are respected.
- Add content to a required field w/ the DemoImageElement, and refresh the page. The validation message should not appear.


SHession (approved):
This looks like a great improvement which reduces a great amount of boilerplate 👏

PR #78, opened 2021-08-12
feat: Add transformer functions when getting element data from nodes, and vice versa

jonathonherbert:
_co-authored-by: @rhystmills_ 

## What does this change?

Adds the ability to add options transformer functions to an element spec. These transformer functions let consumers transform data between an external representation of element data, and the `prosemirror-elements` representation.

As an example, at the Guardian, flexible-content expresses our elements as 

```ts
{
  elementType: string,
  fields: { ... element fields },
  assets: Asset[]
}
```

Additionally, some elements require additional, static field values, e.g. `isMandatory`, which aren't present in the element fields.

By allowing the consumer to supply a transformer function on a per-element basis, and applying them when `getElementDataFromNode` and `getNodeFromElementData` are called, we can ensure that we keep the internal and external models are kept separate, whilst making getting data in and out of prosemirror documents as simple as possible at the call site.

The signature of the transform object is

```ts
  {
    transformElementDataIn: (external: ExternalData) => {
      // return internal representation
    },
    transformElementDataOut: (internal): ExternalData => {
      // return external representation
    },
  }
```

We can add these transformers to our current elements in subsequent PRs.

## How to test

The automated tests should pass. We've added tests for runtime and type behaviour. It should be clear that:

- The code works!
- Typescript knows about the shape of the data concern, on input and output, and won't tolerate badly-formed data.

## Things to ponder

This makes our types, already fairly dense, even more complex. There's a legitimate question as to whether these transforms belong here – or should be left to consumer code. This approach makes calling our input/output functions really need, but does make declaring an element more complicated.

As an alternative, we could defer this to Composer, complicating the code in Composer but avoiding additional complexity in this library's interface. Thoughts welcome.

One way of reducing this complexity would be to create a helper function, e.g. `createGuElementSpec`, which supplies the default Composer transformer functions so Guardian devs don't have to. (We could add an argument to allow them to pass additional fields, like `isMandatory`, if necessary.)

### DIFF
--- a/src/plugin/__tests__/element.spec.ts
+++ b/src/plugin/__tests__/element.spec.ts
@@ -1,5 +1,6 @@
 import type { Node } from "prosemirror-model";
 import { buildElementPlugin } from "../element";
+import { createElementSpec } from "../elementSpec";
 import type { CustomField } from "../fieldViews/CustomFieldView";
 import {
   createEditorWithElements,
@@ -329,6 +330,12 @@ describe("buildElementPlugin", () => {
         </testelement2>
       `;
 
+    const testElementTransformHTML = `
+      <testelementwithtransform type="testElementWithTransform" has-errors="false">
+      <element-testelementwithtransform-field1 class="ProsemirrorElement__testelementwithtransform-field1"><p></p></element-testelementwithtransform-field1>
+      </testelement2>
+    `;
+
     const testElement = createNoopElement({
       field1: { type: "richText" },
       field2: {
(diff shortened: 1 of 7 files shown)


src/plugin/__tests__/element.spec.ts
```diff
@@ -350,6 +357,45 @@ describe("buildElementPlugin", () => {
...
+      () => null,
+      () => null,
+      {
+        transformElementDataIn: (external: ExternalData) => {
+          return external.nestedElementValues;
+        },
+        transformElementDataOut: (internal): ExternalData => {
+          return {
+            nestedElementValues: internal,
+          };
```
dskamiotis, line 376: Im assuming this will reshape/redefine the JSON response for CAPI?
- Deal with data output in flexible content: 
```json
  {
    elementType: "pullquote",
    fields: {
      html: "Example pullquote",
      role: "supporting",
      attribution: "Attribution"
    },
    assets: [ ]
  },
```

This is a great idea to centralise and standardise it.

dskamiotis (approved):
All test (integration + unit) passed and compiles and runs successfully. Looks really good, and definitely a good question re
`As an alternative, we could defer this to Composer, complicating the code in Composer but avoiding additional complexity in this library's interface. Thoughts welcome.`
We could look at how complex creating an element will be for clients using the pugin accompanied by making it explicit to clients that this is indeed possible?

Implementing a `createGuElementSpec` helper with decent documentation seems sensible.

src/plugin/__tests__/element.spec.ts
rhystmills: Exactly, this would be a means to transform the data structure used by ProseMirror to one expected by CAPI.

PR #79, opened 2021-08-18
style: Ergonomic type renaming

SHession:
## What does this change?

This PR aims to improve the readability of the library by making a few helpful renames to important types:

Field -> FieldDescription
FieldSpecs -> FieldDescriptions

FieldViewSpec -> Field
 Properties: 
 - fieldSpec -> description
 - fieldView -> view

Field (React component) -> FieldWrapper

This should make setting up new elements and their views easier to understand intuitively and reasoning with the type system more natural. 

## How to test

This change should be a no-op and make no functional changes. Consider how this renamed types and arguments affect how easy it is to understand the api and create new elements. 

Comparing our API and type system to our [vision document](https://github.com/guardian/prosemirror-elements/blob/main/docs/vision.md), the plugin should  "abstract the complexity of adding nested content and provide developers with a clear intuitive interface".

### DIFF
--- a/src/elements/embed/EmbedForm.tsx
+++ b/src/elements/embed/EmbedForm.tsx
@@ -1,51 +1,51 @@
 import React from "react";
-import { Field } from "../../editorial-source-components/Field";
+import { FieldWrapper } from "../../e../../editorial-source-components/FieldWrapper"
 import type { FieldNameToValueMap } from "../../plugin/fieldViews/helpers";
-import type { FieldNameToFieldViewSpec } from "../../plugin/types/Element";
+import type { FieldNameToField } from "../../plugin/types/Element";
 import { CustomCheckboxView } from "../../renderers/react/customFieldViewComponents/CustomCheckboxView";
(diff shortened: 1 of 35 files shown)


src/elements/embed/EmbedForm.tsx
```diff
@@ -1,51 +1,51 @@
 import React from "react";
-import { Field } from "../../editorial-source-components/Field";
+import { FieldWrapper } from "../../e../../editorial-source-components/FieldWrapper";
```
rhystmills, line 2: Stray 'e' in the relative path here?

rhystmills (approved):
These renames look much better to me - they are a lot easier to understand conceptually.

All unit and integration tests are running as expected for me.

Only issue I've noticed is a stray 'e' in the path for an import in `EmbedForm.tsx`.

src/elements/embed/EmbedForm.tsx
SHession: Great spot, I'll sort that

PR #80, opened 2021-08-19
fix: Fixes element wrapper arrows not updating correctly

SHession:
## What does this change?

This PR fixes a small issue with the element wrapper arrows not updating correctly. This can mean it's possible for the element to become stuck in a single position. 

The fix works by building on an existing solution which applies an invisible decoration to all elements, ensuring any updates are caught and process. This decoration was only previously applied to the image element as the decoration was looking for an element with a certain name.

There may be nice and more performant ways of ensuring these arrows update, but this will resolve the issue for now.

## How to test

Using this branch, do the wrapper arrows function correctly?

## Images

before: 
ELEM_Wrapper_ISSUE_before

after: 
ELEM_Wrapper_ISSUE

## Accessibility

### DIFF
--- a/src/plugin/nodeSpec.ts
+++ b/src/plugin/nodeSpec.ts
@@ -35,6 +35,10 @@ const getNodeSpecForElement = (
     ).join(" "),
     attrs: {
       type: elementName,
+      //Used to determine which nodes show have a forced update decoartion applied. See prosemirror.ts
+      addUpdateDecoration: {
+        default: true,
+      },
       hasErrors: {
         default: false,
       },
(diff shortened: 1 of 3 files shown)


src/plugin/nodeSpec.ts
```diff
@@ -35,6 +35,10 @@ const getNodeSpecForElement = (
     ).join(" "),
     attrs: {
       type: elementName,
+      //Used to determine which nodes show have a forced update decoartion applied. See prosemirror.ts
```
jonathonherbert, line 38:
```suggestion
      // Used to determine which nodes should receive update decorations, which force them to update when the document changes. See `createUpdateDecorations` in prosemirror.ts.
```

jonathonherbert (approved):
Ah, nice! Minor suggestion for phrasing in the comment.

PR #82, opened 2021-08-23
chore: Remove `ref` from `useCustomFieldState`

jonathonherbert:
## What does this change?

At the moment, the `useCustomFieldState` hook has this return type: 

```ts
[
  Data,
  MutableRefObject<((fields: Data) => void) | undefined>
]
```

This isn't great for two reasons:
- the consumer must check for the presence of `current` on the returned ref before using the update function
- this isn't how `useState` hooks usually work!

Using the update API provided in #71, we can now avoid needing to defer the assignment of this function, letting us simplify this type to the more idiomatic 

```ts 
[Data, (fields: Data) => void]
``` 

## How to test
This is a no-op – the tests should pass as usual.

### DIFF
--- a/src/plugin/fieldViews/AttributeFieldView.ts
+++ b/src/plugin/fieldViews/AttributeFieldView.ts
@@ -53,9 +53,9 @@ export abstract class AttributeFieldView<Value extends unknown>
     return true;
   }
 
-  public update = (value: Value) => {
+  public update(value: Value) {
     this.updateOuterEditor(value);
-  };
+  }
 
   public destroy() {
     // Nothing to do – the DOM element is garbage collected.
--- a/src/plugin/fieldViews/AttributeFieldView.ts
+++ b/src/plugin/fieldViews/AttributeFieldView.ts
@@ -53,9 +53,9 @@ export abstract class AttributeFieldView<Value extends unknown>
     return true;
   }
 
-  public update(value: Value) {
+  public update = (value: Value) => {
     this.updateOuterEditor(value);
-  }
+  };
 
   public destroy() {
     // Nothing to do – the DOM element is garbage collected.
--- a/src/renderers/react/useCustomFieldViewState.ts
+++ b/src/renderers/react/useCustomFieldViewState.ts
@@ -1,18 +1,12 @@
-import type { MutableRefObject } from "react";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useState } from "react";
 import { CustomFieldView } from "../../plugin/fieldViews/CustomFieldView";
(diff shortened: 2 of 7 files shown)


src/plugin/fieldViews/AttributeFieldView.ts
```diff
@@ -53,9 +53,9 @@ export abstract class AttributeFieldView<Value extends unknown>
     return true;
   }
 
-  public update(value: Value) {
+  public update = (value: Value) => {
```
SHession, line 56: Why have these been changed to arrow functions?

src/renderers/react/useCustomFieldViewState.ts
```diff
@@ -1,30 +1,24 @@
-import type { MutableRefObject } from "react";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useState } from "react";
 import { CustomFieldView } from "../../plugin/fieldViews/CustomFieldView";
 import type { CustomField } from "../../plugin/types/Element";
 
 export const useCustomFieldViewState = <Data extends unknown>({
```
SHession, line 5: Probably should have got this in the types PR but maybe we could rename it to `useCustomFieldViewState` -> `useCustomFieldState`?

SHession (approved):
I've tested locally and the changes look good. It makes a big difference to the API. I've left a couple of small comments.

src/plugin/fieldViews/AttributeFieldView.ts
jonathonherbert: Having this function lexically scoped means we can pass it around without worrying about it being bound to the wrong thing – e.g. L23 of src/renderers/react/useCustomFieldViewState.ts.  

Having said that, this is a bit obscure. Torn as to whether to wrap the invocation on L23 in another arrow function and restore the old version. Not much in it.

src/renderers/react/useCustomFieldViewState.ts
jonathonherbert: 👍

src/plugin/fieldViews/AttributeFieldView.ts
SHession: Of course, I missed that. I think it would be preferable to keep our functions consistent in the class and wrap it as you described, it would mean you only have use that syntax in one place rather than in each class. However, I'm not too fussy.

jonathonherbert: Agreed, I'll do that – thanks for the feedback.

SHession (approved):

PR #83, opened 2021-08-23
feat: Remove typesafety for getNodeFromElementData

jonathonherbert:
## What does this change?

Removes some typesafety for `getNodeFromElementData` in favour of a runtime check. In our CMS, we often have operations that iterate through a series of elements to conditionally produce nodes. Not all of those elements will be managed by this plugin. As a result, a better API is to conditionally return a node _if_ this node is handled by prosemirror-elements.

For example, this code, if strictly typed, will not be possible: 

```ts
const elements = documentElements; // Some of these aren't handled by prosemirror-elements
const nodes = elements.map(getNodeFromElement) // as a result, this will not typecheck
```

We'd have to write something like this, instead:

```ts
const elements = documentElements; // Some of these aren't handled by prosemirror-elements
const nodes = elements.map(element => {
  if (elementIsHandledByProsemirrorElements()) { // This function doesn't exist yet!
    getNodeFromElement(element)
  } else {
    // Do something else with the element
  }
)
```

This results in less work for the consuming code, which would otherwise have to figure out if an element was handled by prosemirror-elements (through a mechanism that doesn't yet exist) before passing it to `getNodeFromElementData`.

## How to test

The automated tests should pass.

### DIFF
--- a/src/plugin/__tests__/element.spec.ts
+++ b/src/plugin/__tests__/element.spec.ts
@@ -467,29 +467,20 @@ describe("buildElementPlugin", () => {
           expect(node?.eq(view.state.doc.firstChild as Node)).toBe(true);
         });
 
-        it("should not permit data that does not match an element", () => {
+        it("should return undefined if the element does not exist in the element map", () => {
           const { getNodeFromElementData, view } = createEditorWithElements(
             { testElement, testElement2 },
             testElementHTML
           );
 
-          getNodeFromElementData(
-            {
-              elementName: "testElement",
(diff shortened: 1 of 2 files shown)


rhystmills (approved):
All makes sense to me!

dskamiotis:
Sensible API design, even if seems counter intuitive with less type safety, on closer inspection far more effective and productive.

dskamiotis (approved):
All integration and unit tests running and passing successfully

PR #84, opened 2021-08-23
feat: Add custom groupName to elements

jonathonherbert:
## What does this change?

At the moment, we hardcode our group name into the element nodeSpec. This doesn't work in our CMS, which expects an 'element' group. Other consumers may expect other groups – so, this PR allows the consumer to specify, defaulting to 'block', as it's already present in the default schema and will make it easy to experiment with this library.

## How to test

- The new automated test should pass.
- You should be able to see the new groupName in the schema.


rhystmills (approved):
This all looks good to me. All tests passing locally and application running as expected 👍

PR #85, opened 2021-08-24
feat: Add gu element creator

jonathonherbert:
## What does this change?

Adds a new function, `createGuElement`, to take care of boilerplate when creating Guardian-specific Elements. This function makes two assumptions:

- you want to use React to render the element, with the standard `ElementWrapper`
- you need to transform the flat fields generated by prosemirror-elements to fit the [`flexible-model` format.](https://github.com/guardian/flexible-model/blob/main/src/main/thrift/content.thrift)

Eventually this function is intended to live in a specific gu-elements package, alongside the Guardian-specific elements that currently live alongside it in the `elements` folder.

## How to test

The automated test should pass. It roundtrips element data, expressed in the flexible-model format, into and out of a prosemirror node.

## Dev notes

This PR (incorrectly!) assumes that the assets will always be available in the root fields object, or undefined. My presumption is that we're not going to know where those assets live, generically – so we'll likely need to specify that in some way in this function.

Not quite sure how to do that yet, but in the absence of an element with assets, this PR gets us off the ground.

### DIFF
--- a/src/elements/__tests__/createGuElementSpec.spec.tsx
+++ b/src/elements/__tests__/createGuElementSpec.spec.tsx
@@ -1,0 +1,45 @@
+import type { Node } from "prosemirror-model";
+import { createTextField } from "../../plugin/fieldViews/TextFieldView";
+import { createEditorWithElements } from "../../plugin/helpers/test";
+import { createGuElementSpec } from "../createGuElementSpec";
+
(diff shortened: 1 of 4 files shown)


rhystmills (approved):
The interface makes sense to me, all tests are passing and the application is running as expected 👍

PR #86, opened 2021-08-24
feat: Handle passing non-element nodes to getElementDataFromNode

jonathonherbert:
## What does this change?

At the moment, `getElementDataFromNode` always expects to be passed a node representing an element. But how do we know this before we pass one?

This PR alters its signature and implementation to accept any node, and return `undefined` if it receives something it doesn't recognise.

This makes it easy to iterate through a document in consumer code, passing nodes blindly into `getElementDataFromNode` for transformation, and handling them in some other way if we receive `undefined`.

## How to test

The automated tests should pass, and adequately test for this condition.

### DIFF
--- a/src/plugin/__tests__/element.spec.ts
+++ b/src/plugin/__tests__/element.spec.ts
@@ -573,7 +573,7 @@ describe("buildElementPlugin", () => {
           );
 
           // Type refinement should work on each element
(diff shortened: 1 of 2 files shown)


rhystmills (approved):
Code looks sensible to me, all tests passing and running locally 👍

PR #88, opened 2021-08-24
refactor: Un-nests checkbox field values

SHession:
## What does this change?

This PR un-nests the checkbox field values, which should make them a little easier to work.

Before:
`CheckboxValue = { value: boolean };`

After
`CheckboxValue = boolean;`

## How to test

This should be a no-op in terms of functionality but make checkbox fields easier to work with.

## Accessibility

### DIFF
--- a/src/plugin/__tests__/elementSpec.spec.ts
+++ b/src/plugin/__tests__/elementSpec.spec.ts
@@ -48,7 +48,7 @@ describe("mount", () => {
         },
         field2: {
           type: "checkbox",
-          defaultValue: { value: true },
+          defaultValue: true,
         },
       } as const;
       createElementSpec(
@@ -58,7 +58,7 @@ describe("mount", () => {
           // field1 is derived from the fieldDescriptions, and is a string b/c it's a richText field
           fields.field1.toString();
           // field2 is a boolean b/c it's a checkbox field
-          fields.field2.value.valueOf();
+          fields.field2;
           return null;
         }
       );
@@ -182,7 +182,7 @@ describe("mount", () => {
           const fieldDescriptions = {
             field1: {
               type: "checkbox" as const,
-              defaultValue: { value: true },
+              defaultValue: true,
             },
           };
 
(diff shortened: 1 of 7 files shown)


src/plugin/plugin.ts
```diff
@@ -125,17 +126,31 @@ const createNodeView = <
...
-      name,
-      view: fieldView,
-      // We coerce types here: it's difficult to prove we've the right shape here
-      // to the compiler, and we're already beholden to runtime behaviour as there's
-      // no guarantee that the node's `name` matches our spec. The errors above should
-      // help to defend when something's wrong.
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any -- as above
-      update: (value: unknown) => fieldView.update(value as any),
-    } as unknown) as FieldNameToField<FDesc>[typeof name];
+    if (fieldView instanceof CheckboxFieldView) {
```
SHession, line 129: This is a bit nasty as the intersectional type of `boolean` and `string` is `never`. If there is a nicer way of doing this I would be keen to explore it.

src/plugin/__tests__/elementSpec.spec.ts
```diff
@@ -58,7 +58,7 @@ describe("mount", () => {
           // field1 is derived from the fieldDescriptions, and is a string b/c it's a richText field
           fields.field1.toString();
           // field2 is a boolean b/c it's a checkbox field
-          fields.field2.value.valueOf();
+          fields.field2;
```
jonathonherbert, line 61: This no longer tests the boolean-ness of the field – we want something that'll exist on a Boolean, but no other type. IIRC valueOf was such a method.

src/plugin/plugin.ts
jonathonherbert: I'd be tempted to hand-wave types here, as it feels a bit odd introducing runtime behaviour to get around them – v. happy to pair on this!

jonathonherbert (commented):
Great to see this! _Why_ did I nest these values in the first instance 😅 

A few things to mention before merging.

jonathonherbert (approved):

PR #89, opened 2021-08-25
feat: Remove custom tags and classes

jonathonherbert:
_co-authored-by: @rhystmills_

## What does this change?

Removes custom tags and classes from serialised elements, in favour of attributes prefixed by `pm-elements`.

Using custom tags isn't recommend by the [W3C spec](https://html.spec.whatwg.org/multipage/dom.html#elements), and has caused problems where element names have coincided with actual element names (e.g. `code`), causing odd behaviour. We could namespace element tags to solve the latter, but making them `divs` also solves the former.

## How to test

The tests have been modified to reflect the new HTML output, but this is a no-op – everything should work as before.

### DIFF
--- a/src/plugin/nodeSpec.ts
+++ b/src/plugin/nodeSpec.ts
@@ -5,8 +5,8 @@ import type { FieldNameToValueMap } from "./fieldViews/helpers";
 import { fieldTypeToViewMap } from "./fieldViews/helpers";
 import type { FieldDescription, FieldDescriptions } from "./types/Element";
 
-export const elementTypeAttr = "pm-elements-element-type";
-export const fieldNameAttr = "pm-elements-field-name";
+export const elementTypeAttr = "pme-element-type";
+export const fieldNameAttr = "pme-field-name";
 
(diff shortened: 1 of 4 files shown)


src/plugin/nodeSpec.ts
```diff
@@ -5,6 +5,9 @@ import type { FieldNameToValueMap } from "./fieldViews/helpers";
 import { fieldTypeToViewMap } from "./fieldViews/helpers";
 import type { FieldDescription, FieldDescriptions } from "./types/Element";
 
+export const elementTypeAttr = "pm-elements-element-type";
+export const fieldNameAttr = "pm-elements-field-name";
```
SHession, line 9: I wonder if it might be nice to call these `pme-field-name` and `pme-element-type`?

SHession (approved):
Tested locally, LGTM. Small suggestion about the name of `pm-elements-element-type`.

src/plugin/nodeSpec.ts
jonathonherbert: haha, I was having exactly the same thought!

PR #90, opened 2021-08-25
fix: Restrict marks in text elements

SHession:
## What does this change?

This PR restrict all marks in text fields to prevent them from being added by other means i.e. copy and paste.

## How to test

Trying copying some text from an RTE field with marks into a text field. 
There is currently no Cypress test that checks this path, but I  have added some shortcut tests for good measure.

Text fields never contain marks.

## Accessibility


jonathonherbert (approved):
LGTM, ace – quick solve on the platform-specific shortcuts 😁

PR #91, opened 2021-08-27
feat: Add validation results to `getElementDataFromNode`

jonathonherbert:
## What does this change?

Adds the validation results to the output of `getElementDataFromNode`. This makes it easy for consumers to get hold of results when they're processing the output of the plugin.

## How to test

The new automated test should pass, and it should be clear what it's doing.

## Dev notes

I've changed the output of the `Validator` function in this PR from `null` to `undefined` – we don't use `null` anywhere else in this library, and I'm not aware of a good reason to use it over `undefined` in this instance, but I'd be interested if there was a rationale!


SHession (approved):
Tested locally and reviewed the code changes, LGTM.

PR #92, opened 2021-08-31
feat: Re-style code element

rhystmills:
Co-authored by @jonathonherbert 

## What does this change?
 
This makes a few changes to the element layouts, especially the Code element:
- The close button is moved to the left side of the element, to prevent accidental clicks
- The Code element's dropdown is moved inline with its label (the option is also available for other dropdowns via the `display` property
- The Code element now uses less space by default (4 lines instead of 11).

## How to test
Run `yarn start` from the root of the repository.

## Images
| Before | After |
|--|--|
| image | image | 

## Accessibility

(N.B - tested with screen reader - found that our buttons are not very accessible - they need a label)

### DIFF
--- a/src/renderers/react/ElementWrapper.tsx
+++ b/src/renderers/react/ElementWrapper.tsx
@@ -2,7 +2,7 @@ import { css } from "@emotion/react";
 import styled from "@emotion/styled";
 import { space } from "@guardian/src-foundations";
 import { focusHalo } from "@guardian/src-foundations/accessibility";
-import { neutral } from "@guardian/src-foundations/palette";
+import { border, neutral } from "@guardian/src-foundations/palette";
 import {
   SvgArrowDownStraight,
   SvgArrowUpStraight,
@@ -13,8 +13,11 @@ import type { ReactElement } from "react";
 import React from "react";
 import type { CommandCreator } from "../../plugin/types/Commands";
 
+const buttonWidth = 32;
+
 const Container = styled("div")`
   margin: ${space[3]}px 0;
+  position: relative;
 `;
 
 const Body = styled("div")`
@@ -39,26 +42,19 @@ const Panel = styled("div")`
   padding-right: ${space[3]}px;
 `;
 
-const Actions = styled("div")`
-  display: flex;
-  flex-direction: column;
-  opacity: 0;
-  transition: opacity 0.2s;
-`;
-
-const Button = styled("button")`
+const Button = styled("button")<{ expanded?: boolean }>`
   appearance: none;
   background: ${neutral[93]};
   border: none;
   border-top: 1px solid ${neutral[100]};
   color: ${neutral[100]};
   cursor: pointer;
-  flex-grow: ${({ expanded }: { expanded?: boolean }) =>
-    expanded ? "1" : "0"};
+  flex-grow: ${({ expanded }) => (expanded ? "1" : "0")};
+  ${({ expanded }) => !expanded && "height: 32px;"};
   font-size: 16px;
   line-height: 1;
   padding: ${space[1]}px;
-  width: 32px;
+  width: ${buttonWidth}px;
   transition: background-color 0.1s;
   :focus {
     ${focusHalo}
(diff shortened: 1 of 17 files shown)


SHession:
This looks like a sensible set of a changes. A couple of initial thoughts from me, firstly, I think I prefer the close button on the right-hand side as having it on the left creates a mostly unnecessary margin. Secondly, the composer element requires two clicks to delete the element, might this instead be a good way to prevent accidental deletes?

jonathonherbert:
Both good points. We thought a bit about these decisions, be interested to hear thoughts.

Re: close on lhs:
- It's there in the current element designs, so this change will be mean editorial muscle memory will still work (bear in mind this element will coexist with others, so this could get quite confusing if one element out of 15 had different controls)
- It prevents a misclick on 'send to bottom' removing the element
- It falls outside the boundaries of the element in the same way that the elements in Composer do – although confusingly, in the demo, we must introduce a gutter to the editor to ensure everything's displayed, the grey boundary of the element does line up with the editor text.

Re: confirm on close:
- my suspicion is that this is always hastily double clicked (no other editor behaviour has 'confirm' semantics IIRC)
- It's styled poorly (brokenly) in Composer as it stands, and getting this right will be additional effort
- We have an easy way of getting things back in RTEs – Ctrl-Z! (Legitimate question – would users give this a try? My hunch is yes, but others may reasonably disagree)

If we do want confirm on close, let's add it in another PR.

rhystmills:
> This looks like a sensible set of a changes. A couple of initial thoughts from me, firstly, I think I prefer the close button on the right-hand side as having it on the left creates a mostly unnecessary margin. Secondly, the composer element requires two clicks to delete the element, might this instead be a good way to prevent accidental deletes?

Cheers Sam.

Our thoughts were that:
- Composer has the margin in place so we wouldn't be creating a new margin, only using the existing one
- Users expect the close button to be there based on the previous designs
- Also including the double click functionality was overkill and a slightly odd design pattern anyway, especially given that users can undo deleting an element

On the other hand I don't particularly like the close button sitting over there myself, and we could free up that left margin  if all the elements conformed to a right-side close button. Perhaps it's worth getting UX input or finding out why that decision was made initially?

paperboyo:
> no other editor behaviour has 'confirm' semantics IIRC

Publish haz it. And it’s all over eg. Delete in Grid. Whether it’s a good pattern (or promotes mindless doubleclick and causes accidental publish actions) is another matter.

jonathonherbert:
> > no other editor behaviour has 'confirm' semantics IIRC
> 
> Publish haz it. And it’s all over eg. Delete in Grid. Whether it’s a good pattern (or promotes mindless doubleclick and causes accidental publish actions) is another matter.

... specifically in the RTE (although, I bet there's _something_ 😅)

SHession:
Thanks both very useful thoughts. I think there is an interesting questions here, namely, how closely should we try to preserve existing behaviour? 

In regards to moving the button to the left we are arguing in favour of persevering muscle memory and consistency between elements. In the case of the double click we are arguing the opposite, saying this is a bug and favouring its removal. The reasons behind this are valid, but I thought it was worth pointing out the juxtaposition. 

I think maintaining consistency in the position makes sense for now, as you say, it will be confusing if it's different between elements. Regarding the double click, I think the undo functionality is a great point and not one I had considered. Testing this in Composer and PME both have the same issue of the editor losing focus, meaning cmd+Z doesn't initially work. This intuitively makes it feel like the element is gone and at best makes it'a clunky to restore. Perhaps this is an improvement we could make in PME. Until then, I think including the double click might be a good to include for now, under the same consistency arguments as the positioning.

Finally, I still prefer the positioning on the RHS, but would be happy to defer this (and other usability improvements) to a time where all elements are displayed through PME.

rhystmills:
That makes perfect sense to me Sam - I'll work to add the double-click functionality.

jonathonherbert:
Great points both. Agreed that consistency is the sticking point.

Let's implement the delay in another piece of work, as it's not yet clear what the design will be, and this is already a fairly busy PR.

src/renderers/react/ElementWrapper.tsx
```diff
@@ -39,26 +42,19 @@ const Panel = styled("div")`
...
+const Button = styled("button")<{ expanded?: boolean }>`
   appearance: none;
   background: ${neutral[93]};
   border: none;
   border-top: 1px solid ${neutral[100]};
   color: ${neutral[100]};
   cursor: pointer;
-  flex-grow: ${({ expanded }: { expanded?: boolean }) =>
-    expanded ? "1" : "0"};
+  flex-grow: ${({ expanded }) => (expanded ? "1" : "0")};
```
SHession, line 52: I may be missing something obvious, but what is this expanded used for / come from? When is it ever defined?

SHession (approved):
Tested changes locally, LGTM!

src/renderers/react/ElementWrapper.tsx
rhystmills: If the expanded prop exists on the styled component, e.g:
</img>

PR #93, opened 2021-08-31
fix: Alter fields of code element to correspond with flexible-content

jonathonherbert:
## What does this change?

Alter the field names of code element to correspond with flexible-content. In doing this, we also need to strip `isMandatory` from elements that don't explicitly need it – `isMandatory` is ... not required.

## How to test

- The fields for the Code element should be `html` and `language` in the prosemirror dev tools.

Tested locally in Composer via `yarn link`.


rhystmills (approved):
All makes sense to me 👍 

Running locally and all tests passing.

PR #95, opened 2021-08-31
fix: Add aria-labels to move and delete element buttons

rhystmills:
## What does this change?
This adds an `aria-label` attribute to our move and delete element buttons.

## Why?
The buttons do not contain text so currently their purpose is not clear when using a screen reader.

## How to test
Run `yarn start` and test via a screenreader, e.g. VoiceOver on Mac.

## Accessibility

### DIFF
--- a/src/renderers/react/ElementWrapper.tsx
+++ b/src/renderers/react/ElementWrapper.tsx
@@ -140,7 +140,7 @@ export const ElementWrapper: React.FunctionComponent<Props> = ({
           data-cy={removeTestId}
           disabled={!remove(false)}
           onClick={() => remove(true)}
-          aria-label="Delete element."
+          aria-label="Delete element"
         >
           <SvgCross />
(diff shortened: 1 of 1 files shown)


src/renderers/react/ElementWrapper.tsx
```diff
@@ -172,13 +175,15 @@ export const ElementWrapper: React.FunctionComponent<Props> = ({
           expanded
           disabled={!moveDown(false)}
           onClick={() => moveDown(true)}
+          aria-label="Move element down."
```
jonathonherbert, line 178: The nittiest of nits – do aria elements need a full stop?

jonathonherbert (approved):
👍

src/renderers/react/ElementWrapper.tsx
rhystmills: Wise words - will remove!

PR #96, opened 2021-08-31
feat: Support dropdown undefined values

SHession:
## What does this change?

This PR adds the ability to have a dropdown field with an undefined value. This is necessary to support dropdowns of this nature in Composer.

## How to test

This PR adds a dropdown with a default value of undefined into the `DemoImageElement`, this can be tested using the Cypress tests or manually.

We are able to support dropdowns with undefined values options as per our requirements in Composer.

This change effectively treats options with an empty string value as undefined. This may cause issues if consumers ever want an option to represent an empty string. The other way we might approach this issue is to add a custom ingress and egress rule which handles option values we know represent undefined. Interested to hear opinions on this.

## Images

image

## Accessibility

### DIFF
--- a/src/plugin/nodeSpec.ts
+++ b/src/plugin/nodeSpec.ts
@@ -148,7 +148,7 @@ export const getNodeSpecForField = (
           parseDOM: getDefaultParseDOMForLeafNode(elementName, fieldName),
           attrs: {
             fields: {
-              default: field.defaultValue,
+              default: undefined,
             },
           },
         },
@@ -204,8 +204,10 @@ const getDefaultParseDOMForLeafNode = (
         return false;
       }
 
+      const fields = dom.getAttribute("fields");
+
       const attrs = {
-        fields: JSON.parse(dom.getAttribute("fields") ?? "{}") as unknown,
+        fields: (fields ? JSON.parse(fields) : undefined) as unknown,
       };
 
       return attrs;
--- a/src/plugin/nodeSpec.ts
+++ b/src/plugin/nodeSpec.ts
@@ -207,7 +207,7 @@ const getDefaultParseDOMForLeafNode = (
       const fields = dom.getAttribute("fields");
 
       const attrs = {
-        fields: (fields ? JSON.parse(fields) : undefined) as unknown,
+        fields: (fields !== null ? JSON.parse(fields) : undefined) as unknown,
       };
 
       return attrs;
--- a/cypress/helpers/editor.ts
+++ b/cypress/helpers/editor.ts
@@ -99,6 +99,7 @@ export const getSerialisedHtml = ({
   codeValue = "",
   useSrcValue = "false",
   optionValue = "opt1",
+  otherOptionValue = undefined,
   restrictedTextValue = "",
   customDropdownValue = "opt1",
   mainImageValue = {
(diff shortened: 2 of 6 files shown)


src/plugin/nodeSpec.ts
```diff
@@ -204,8 +204,10 @@ const getDefaultParseDOMForLeafNode = (
         return false;
       }
 
+      const fields = dom.getAttribute("fields");
+
       const attrs = {
-        fields: JSON.parse(dom.getAttribute("fields") ?? "{}") as unknown,
+        fields: (fields ? JSON.parse(fields) : undefined) as unknown,
```
SHession, line 210: I had to change the default parsing rule for a leaf node, from am empty object to undefined. This was due to issues with the dropdown being able to find the selected value on a refresh. This may have unintended consequences for other fields, I can control this with a function argument if that would be useful?

jonathonherbert: Difficult to unravel what this might mean, as we'd be very surprised if fields weren't present on this node! Arguably this makes more sense, as an empty object doesn't make sense for many sorts of attributes (e.g. string, number); undefined is probably less surprising. Whether we account for this possibility in the types for that field is another matter. The non-presence of the `field` attribute is arguably a good spot to throw an error.

However, we should continue to use the null coalescing operator (`??`) when we're deciding whether to fall through to `undefined`, because values like `""` or `0` will be coerced to `undefined` by this change.

jonathonherbert:
This is an interesting change. IIUC, as we use `CustomDropdownField` for our own components, in the absence of an actual use case I'm wondering if it needs to be made. The question I'm asking myself is – what do we gain from having an opinion about this sort of thing, for a component that as far as I can see, we don't use?

The code looks good though! There's one blocker I've mentioned re `??`. Once addressed, happy to approve.

src/plugin/nodeSpec.ts
SHession: Great point, I will fix that.

SHession:
> This is an interesting change. IIUC, as we use `CustomDropdownField` for our own components, in the absence of an actual use case I'm wondering if it needs to be made. The question I'm asking myself is – what do we gain from having an opinion about this sort of thing, for a component that as far as I can see, we don't use?

It's a valid question, I was making this change to the dropdown partly as a spike and partly for completeness. The branch off of this one, will make the same changes to the custom dropdown too. This work has resolved some of the issues that would also be present for that implementation. Unless you have specific concerns about supporting undefined values in the standard dropdown I see no harm in making the change for the standard and custom dropdowns.

src/plugin/nodeSpec.ts
jonathonherbert: Ah, specifically in the case when the attribute does not exist (not when it does exist, but contains `undefined`) – but arguably we shouldn't be throwing at this point anyhow 👍

SHession: Yeah it does make sense, but I'm not sure how we could check for the case where it exists but is undefined. `dom.getAttribute("fields")` returns a `string | null`.

SHession:
Closing this PR due to a concern about the complexity it adds in order to support non-serializable types i.e. undefined. See #100 which attempts to salvage the useful parts out.

PR #98, opened 2021-09-01
feat: Add classes to nested Prosemirror instances for consumer styling

jonathonherbert:
## What does this change?

At the moment, in the demo page, we inherit lots of styles from the Prosemirror css directly, imported from the Prosemirror website. This PR moves many these styles in-house, removing the dependency on external stylesheets and making it explicit what assumptions our demo page is making re: instance styling.

It also adds classes to the Prosemirror DOM nodes created by both the Text and RichText FieldViews. This allows consumers to target them directly.

## How to test

- Is the demo page still styled correctly?
- Do classes appear on the root nodes of nested Prosemirror instances according to their FieldView? You should see e.g. 
Screenshot 2021-09-01 at 12 14 49

### DIFF
--- a/src/plugin/fieldViews/ProseMirrorFieldView.ts
+++ b/src/plugin/fieldViews/ProseMirrorFieldView.ts
@@ -51,7 +51,7 @@ export abstract class ProseMirrorFieldView implements FieldView<string> {
     // The ProseMirror node type name
     private readonly fieldName: string,
     // Plugins that the editor should use
-    private plugins?: Plugin[]
+    plugins?: Plugin[]
   ) {
     this.applyDecorationsFromOuterEditor(decorations);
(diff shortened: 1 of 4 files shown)


src/plugin/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -51,7 +51,7 @@ export abstract class ProseMirrorFieldView implements FieldView<string> {
     // The ProseMirror node type name
     private readonly fieldName: string,
     // Plugins that the editor should use
-    private plugins?: Plugin[]
+    plugins?: Plugin[]
```
jonathonherbert, line 54: Whilst exploring, I noticed this isn't used outside of the constructor, so I tidied it up.

SHession (approved):

PR #99, opened 2021-09-01
feat: Support custom dropdown undefined values

SHession:
## What does this change?

This PR build on the work of #96  and adds support for undefined option values for custom dropdowns.

## How to test

This PR adds a custom dropdown with a default value of undefined into the DemoImageElement, this can be tested using the Cypress tests or manually.

We are able to support dropdowns with undefined values options as per our requirements in Composer.

## Accessibility

### DIFF
--- a/src/plugin/nodeSpec.ts
+++ b/src/plugin/nodeSpec.ts
@@ -161,7 +161,7 @@ export const getNodeSpecForField = (
           parseDOM: getDefaultParseDOMForLeafNode(elementName, fieldName),
           attrs: {
             fields: {
-              default: { value: field.defaultValue },
+              default: field.defaultValue,
             },
           },
         },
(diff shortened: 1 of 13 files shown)


src/plugin/nodeSpec.ts
```diff
@@ -161,7 +161,7 @@ export const getNodeSpecForField = (
           parseDOM: getDefaultParseDOMForLeafNode(elementName, fieldName),
           attrs: {
             fields: {
-              default: { value: field.defaultValue },
+              default: field.defaultValue,
```
SHession, line 164: Another potentially contentious change here. This PR changes how default custom field attributes are modelled. This is due to how newNodes are created and how `setNodeMarkup` treats `undefined` as a missing argument. The default needs to be undefined, otherwise we can't set the attribute value to it.

Not the best explanation, happy to go into more detail if that's helpful.

jonathonherbert: That does make sense, and we should support `undefined` as a possible attribute value – 👍

SHession:
Closing this PR due to a concern about the complexity it adds in order to support non-serializable types i.e. undefined. See #100 which attempts to salvage the useful parts out.

PR #100, opened 2021-09-02
refactor: Refactor dropdown types

SHession:
## What does this change?

After a discussion of how to handle undefined values in dropdown, this PR attempts to save some of the little tweaks that were made to enable it. #96 and #99 will be closed due to a concern about adding too much additional complexity to support non-serializable types i.e. undefined.

## How to test

This change should be a no-op and all tests should pass as before.

It's easier to work with custom dropdowns.

## Accessibility


jonathonherbert (approved):
Looks good, types are neater and handy to have a helper for custom dropdowns.

PR #101, opened 2021-09-02
fix: Remove custom tags for rich text elements

jonathonherbert:
## What does this change?

#89 was incomplete – it didn't remove custom tags from the deserialisation step (`getAttrs`) in the nodeSpec declaration for rich text elements.

As a result, we were seeing odd results when deserialising from HTML:

This PR makes that change, adding a test beforehand to verify that this behaviour is fixed, as it was previously a blind spot.

## More details

Our `getAttrs` function was still using old code that expected a div with a custom element tag (e.g. `<element-image-element-caption>`, rather than `<div pme-field-name="imageElement_caption">`.

So, we serialised the element correctly, but when Prosemirror attempted to deserialise the element (e.g. on a page refresh, as we store the document as html in localStorage), Prosemirror didn't understand the shape of the rich text data – it was expecting `<element-image-element-caption>`, but found `<div pme-field-name="imageElement_caption">`.

As a result, it made creative guesses about where the data should go, which were often incorrect!

## How to test

- The automated test should pass.
- Attempting to replicate the behaviour in the GIF should no longer be possible.

### DIFF
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -10,6 +10,7 @@ import {
   getSerialisedHtml,
   italicShortcut,
   selectDataCy,
+  setDocFromHtml,
   typeIntoElementField,
   visitRoot,
 } from "../helpers/editor";
@@ -143,6 +144,12 @@ describe("ImageElement", () => {
           })
         );
       });
+
+      it("should deserialise content from HTML into the appropriate node in the document", () => {
+        const values = { altTextValue: "Alt text" };
+        setDocFromHtml(values);
+        assertDocHtml(getSerialisedHtml(values));
+      });
     });
 
     describe("Text field", () => {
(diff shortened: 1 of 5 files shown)


cypress/tests/ImageElement.spec.ts
```diff
@@ -211,6 +218,12 @@ describe("ImageElement", () => {
         typeIntoElementField("src", "Src text");
         assertDocHtml(getSerialisedHtml({ srcValue: "Src text" }));
       });
+
+      it("should deserialise content from HTML into the appropriate node in the document", () => {
```
SHession, line 222: Do we want tests like these for other field types i.e. checkbox, dropdown?

SHession (approved):
Tested the issue with the example of typing text into the demo element `src` field and refreshing to see it in the `restricted text field`. The content now no longer parses into the wrong field.

cypress/tests/ImageElement.spec.ts
jonathonherbert: I think it'd be good to have them in a separate PR – I've added a card to our backlog.

PR #102, opened 2021-09-06
feat: Click to confirm element close

rhystmills:
## What does this change?
Elements in Composer ([flexible-content](https://github.com/guardian/flexible-content)) currently require two clicks to close, but the implementation is not working very well. 

In order to be consistent with this existing behaviour (especially during migration where some, but not all, elements may be migrated), this PR adds 'click to confirm close' behaviour to our ProseMirror Elements. However, we've tidied the implementation with a new hover tooltip. As before, there is only a limited timeframe in which a user can confirm the close, after which a timeout occurs, and the button returns to its initial state.

**Once all elements are migrated to ProseMirror, we can revisit whether we want this behaviour at all.**

(N.B - due to a lack of bin icon in Source, we've borrowed one from the material design icon set).

## How to test

Run `yarn start` from the repository root and try closing an element.

## Images

| Close timeout | Close the element |
| --- | --- |
|  |  |

#### It's also worth noting that introducing a timeout introduces additional WCAG criteria to consider, e.g. https://www.w3.org/TR/WCAG21/#timing-adjustable

### DIFF
--- a/cypress/tests/ElementWrapper.spec.ts
+++ b/cypress/tests/ElementWrapper.spec.ts
@@ -59,6 +59,7 @@ describe("ElementWrapper", () => {
     it("should remove an element", async () => {
       addImageElement();
       cy.get(selectDataCy(removeTestId)).click();
+      cy.get(selectDataCy(removeTestId)).click();
       const elementTypes = await getArrayOfBlockElementTypes();
       expect(elementTypes).to.deep.equal(["paragraph", "paragraph"]);
     });
--- a/src/editorial-source-components/SvgBin.tsx
+++ b/src/editorial-source-components/SvgBin.tsx
@@ -1,0 +1,14 @@
+export const SvgBin = () => {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      height="24px"
+      viewBox="0 0 24 24"
+      width="24px"
+      fill="#000000"
+    >
+      <path d="M0 0h24v24H0z" fill="none" />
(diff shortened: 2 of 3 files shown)


jonathonherbert:
Code looks great! A few minor design things:

- It'd be good to keep the tooltip text the same size as the element text
- Same w/ aligning its left edge w/ the left edge of the button
- At the moment, the colour sequence is a bit odd – hover: red -> click once: black feels like you're de-escalating once you click. Would swapping these colours (hover: black -> click once: red) make things clearer?
- Nit, feel free to ignore: I'd recommend either using box-shadow of the same style that composer uses (e.g. `0 2px 4px rgb(0 0 0 / 30%)`, from the prepublish checks), or keeping the design flat and using a darker bg colour, which I think looks cleaner (in part b/c it's difficult to shadow triangles 😅 ). The latter is personal taste in the absence of a consistent approach from Composer, _do not_ let me yuck your yum 😄 

Screenshot 2021-09-06 at 16 05 46

jonathonherbert:
(... great point re: accessibility considerations, too!)

rhystmills:
Good advice Jon I've had a pop at including it and will update the images in the PR.

I went for a shadow in the end as I found a method which will also shade the triangle.

jonathonherbert:
@rhystmills great work! One final nit (I'm really sorry) – could we keep the 'black' -> 'red' styling (rather than 'red' -> 'darker red')? It'll match what's currently in Composer – unless there's a good reason to change.

jonathonherbert (approved):

rhystmills:
> @rhystmills great work! One final nit (I'm really sorry) – could we keep the 'black' -> 'red' styling (rather than 'red' -> 'darker red')? It'll match what's currently in Composer – unless there's a good reason to change.

Done!

PR #103, opened 2021-09-07
docs: Extended accessibility checklist

rhystmills:
## What does this change?

This PR adds a PR template which extends our accessibility checklist to match the [Editorial Tools Accessibility guidelines](https://github.com/guardian/accessibility/blob/main/editorial-tools-guidelines.md).

These will look like:
## Accessibility

- [ ] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
- [ ] [Navigable with keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
- [ ] [Colour contrast passed](https://webaim.org/resources/contrastchecker/)
- [ ] [The change doesn't use only colour to convey meaning](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html)
- [ ] [Interactive elements show a focus ring when focused by keyboard](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html)
- [ ] [Semantic elements have been used where possible](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html)
<s>[No content flashes more than three times in any one second period](https://www.w3.org/WAI/WCAG21/Understanding/three-flashes-or-below-threshold.html)</s>

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -19,9 +19,9 @@
 This repository follows the Editorial Tools accessibility guidelines: https://github.com/guardian/accessibility/blob/main/editorial-tools-guidelines.md
 -->
 
-- [ ] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
-- [ ] [Navigable with keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
-- [ ] [Colour contrast passed](https://webaim.org/resources/contrastchecker/)
-- [ ] [The change doesn't use only colour to convey meaning](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html)
-- [ ] [Interactive elements show a focus ring when focused by keyboard](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html)
-- [ ] [Semantic elements have been used where possible](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html)
+- [ ] [Tested with screen reader](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#screen-readers)
+- [ ] [Navigable with keyboard](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#keyboard-navigation)
+- [ ] [Colour contrast passed](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#colour-contrast)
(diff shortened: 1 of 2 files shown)


rhystmills:
In hindsight - I wonder if `No content flashes more than three times in any one second period` is perhaps better left out of the checklist - though we can keep an extended list of requirements in the editorial accessibility guidelines.

It will rarely come up in UX changes, and I think developers would challenge it when it did.

rhystmills:
I also wonder if it would be helpful to expand on points like `Semantic elements have been used where possible` in the guidelines - not all developers will know the 'semantic' jargon, or know why they're helpful (e.g. as cues for screenreaders).

jonathonherbert:
@rhystmills I'd agree on both points. The links we've got here are useful in that they link to a canonical spec, but they're probably not the quickest ways for devs to understand what's expected of them.

Perhaps we could add a (guide) or (examples) link where appropriate to external resources that we think will be useful in this way?

rhystmills:
> @rhystmills I'd agree on both points. The links we've got here are useful in that they link to a canonical spec, but they're probably not the quickest ways for devs to understand what's expected of them.
> 
> Perhaps we could add a (guide) or (examples) link where appropriate to external resources that we think will be useful in this way?

Cheers Jon.

Would you say it's worth writing the examples first before finishing this PR? (They could live in the accessibility guidelines alongside the Ed Tools Accessibility Guidelines.

rhystmills:
> @rhystmills I'd agree on both points. The links we've got here are useful in that they link to a canonical spec, but they're probably not the quickest ways for devs to understand what's expected of them.
> 
> Perhaps we could add a (guide) or (examples) link where appropriate to external resources that we think will be useful in this way?

Following up, have set up a second PR with detailed explanations to link to here: https://github.com/guardian/accessibility/pull/7

jonathonherbert (approved):
Looks good, let's merge 👍 

For the future: my feeling is that this list is necessary but probably not sufficient to really get us working towards best practice, and that some kind of automation to sense check at CI would be another good step in that direction. I'd be interested to know what's available in that space – and whether, if and when we have a palette of useful automated checks, whether these can become part of the org's default configuration for new projects etc.

rhystmills:
> Looks good, let's merge 👍
> 
> For the future: my feeling is that this list is necessary but probably not sufficient to really get us working towards best practice, and that some kind of automation to sense check at CI would be another good step in that direction. I'd be interested to know what's available in that space – and whether, if and when we have a palette of useful automated checks, whether these can become part of the org's default configuration for new projects etc.

I agree - Simon pointed us in the direction of a few automated options which we can look into (Pa11y, React Axe, eslint-plugin-jsx-a11y).

PR #104, opened 2021-09-07
feat: refactor validation api

SHession:
## What does this change?

This PR changes how the validation logic is surfaced to the consumer. With this PR it is no longer returned from`getElementFromNode` but instead a function is provided called `elementDataValidator` which can be used to validate element values.

## How to test

I have added some additional unit tests to ensure the functionality is working correctly, but should be a no-op for the demo.

Consumers have an easy way of validating element values, whilst validation logic is nicely encapsulated within the element spec.

## Accessibility

### DIFF
--- a/src/plugin/__tests__/element.spec.ts
+++ b/src/plugin/__tests__/element.spec.ts
@@ -665,6 +665,50 @@ describe("buildElementPlugin", () => {
 
         describe("validateElementData", () => {
           it("should output found errors", () => {
+            const { validateElementData } = createEditorWithElements({
+              testElementWithValidation,
+              testElementWithDifferentValidation,
+            });
+
+            const errors = validateElementData("testElementWithValidation", {
+              field1: "Some text",
+            });
+
+            expect(errors).toEqual({ field1: ["Some error"] });
+
+            const otherErrors = validateElementData(
+              "testElementWithDifferentValidation",
+              { checkbox: true }
+            );
+
+            expect(otherErrors).toEqual({ checkbox: ["Some other error"] });
+          });
+
+          it("should output undefined if there are no errors", () => {
+            const { validateElementData } = createEditorWithElements({
+              testElement,
+            });
+
+            const errors = validateElementData(
+              "testElement",
+              testElementValues.values
+            );
+
+            expect(errors).toEqual(undefined);
+          });
+
+          it("should not allow non-existent elements", () => {
+            const { validateElementData } = createEditorWithElements({
+              testElement,
+            });
+            validateElementData(
+              // @ts-expect-error -- we should not be able to check a non-existent element
+              "non-existing-element",
+              testElementValues.values
+            );
+          });
+
+          it("should accept the getElementDataFromNode output", () => {
             const {
               insertElement,
               getElementDataFromNode,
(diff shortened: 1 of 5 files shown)


src/plugin/__tests__/element.spec.ts
```diff
@@ -654,28 +662,107 @@ describe("buildElementPlugin", () => {
...
+              getElementDataFromNode,
+              elementDataValidator,
+              view,
+              serializer,
+            } = createEditorWithElements({
+              testElementWithValidation,
+              testElementWithDifferentValidation,
+            });
+
+            insertElement({
```
jonathonherbert, line 679: Super minor: do we need to round trip this data through a node and out again? Could be a smaller test if we just pass object values directly. (We could also make this change in the test below.)

src/plugin/helpers/test.ts
```diff
@@ -81,6 +81,7 @@ export const createEditorWithElements = <
     nodeSpec,
     getNodeFromElementData,
     getElementDataFromNode,
+    elementDataValidator,
```
jonathonherbert, line 84: One thought on naming – functions are verbs, not nouns – what do you think about `validateElementData`, which describes what this function does, rather than what it is?

src/plugin/helpers/element.ts
```diff
@@ -122,3 +121,24 @@ const getValuesFromContentNode = (node: Node, serializer: DOMSerializer) => {
...
+export const createElementDataValidator = <
+  FDesc extends FieldDescriptions<keyof FDesc>,
+  ElementNames extends keyof ESpecMap,
+  ExternalData,
+  ESpecMap extends ElementSpecMap<FDesc, ElementNames, ExternalData>
+>(
+  elementTypeMap: ESpecMap
+) => (
+  elementName: keyof ESpecMap,
+  values: unknown
```
jonathonherbert, line 134: Torn on typesafety here. I think we do expect our data to be a valid shape for the element, even if it doesn't pass validation – that is to say, our validators expect an object that matches the type provided by the Element, even if the data itself is invalid.

Suspect that if keys are missing, for example, our validators will throw errors. So a lack of typesafety here could lead to runtime problems.

We could use a signature of the kind used in `insertElement` to make our `values` typesafe rather than `unknown`. I'd value your thoughts!

jonathonherbert (commented):
Two non-blocking comments and one thought about typesafety, but this is looking good! I'll wait for some feedback re: the latter before approving.

src/plugin/__tests__/element.spec.ts
SHession: I think I might add a specific test case for this, to ensure that we can always pass the output from `getElementDataFromNode` directly to the validator. Unless you see this as redundant?

jonathonherbert: It just expands the surface area of what we're testing, but I see the intent – no strong opinions 👍

src/plugin/helpers/element.ts
SHession: Worked through this offline, it's now ready for re-review.

jonathonherbert (approved):
Ace, tests look good too. There's some repetition in a test name that's worth addressing before we merge, otherwise, 👍

src/plugin/__tests__/element.spec.ts
```diff
@@ -654,28 +662,96 @@ describe("buildElementPlugin", () => {
...
+
+            const errors = validateElementData({
+              elementName: "testElement",
+              values: testElementValues.values,
+            });
+
+            expect(errors).toEqual(undefined);
+          });
+
+          it("should output undefined if there are no errors", () => {
```
jonathonherbert, line 703: Guessing this should be renamed.

PR #105, opened 2021-09-07
feat: Button font update for image elelemt

dskamiotis:
co-authored @rhystmills 

## What does this change?
We are currently using Source button for the `copy from caption` and `re crop image` on the image element.
In order to keep the styling in line with the rest of Composer we would like the font for the text to be Agate Sans.

This PR aims to amend the default font for the source button to Agate Sans.

## How to test

Run `yarn start` from the repository root and try closing an element.
Also using dev tools; inspect the font source (as screenshots below)

## Additional information
In order to implement this the following additions were made to this feature:
* New Button component created in its own file
* The button component is based off of Source button
* Additional styling added through props to override the default font
* Re factor the default props for Source button to allow for our custom button

## Images

| Before: Default font   | Styling |
| --- | --- |
| image) |image |

| After: Customised font  | Styling |
| --- | --- |
| image|image |

### DIFF
--- a/src/editorial-source-components/Button.tsx
+++ b/src/editorial-source-components/Button.tsx
@@ -1,0 +1,36 @@
+import { css } from "@emotion/react";
+import { Button as SourceButton } from "@guardian/src-button";
+import type { ReactElement } from "react";
+import React from "react";
+
+export const buttonStyles = css`
+  font-family: "Guardian Agate Sans";
+`;
+
+type Size = "default" | "small" | "xsmall";
+type ButtonPriority = "primary" | "secondary" | "tertiary" | "subdued";
+type IconSide = "left" | "right";
+
+type Props = {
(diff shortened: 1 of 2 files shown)


jonathonherbert:
Great work! One thing to mention before merging – it looks like `ButtonProps` is exported Source, so instead of redeclaring the Props, you should be able to do something like:

```ts
import type { ButtonProps } from "@guardian/src-button";

// ... styling code

export const Button: React.FunctionComponent<ButtonProps> = ({
  children,
  size,
  ...props
}) => {

// ... rest of file
```

Do we need to do the same thing re: fonts for the dropdown component, too?

dskamiotis:
> mention before mer

Thanks @jonathonherbert that makes a lot more sense, should have checked the source of source! 👍

jonathonherbert (approved):

PR #106, opened 2021-09-08
feat: Adds ValidationError type into validator

SHession:
## What does this change?

This PR adds a `ValidationError` type which provides more error information than previously by including an additional `message` property and using the existing string as the `error` property. It also adds a `FieldValidationErrors` type for convenience and readability.
```
export type ValidationError = {
  error: string;
  message: string;
};
export type FieldValidationErrors = Record<string, ValidationError[]>
```

## How to test

This should be a no-op from a user perspective but provide consumers and the view with more information about validation problems.

We have enough data to display clear validation messages to consumers. 

## Accessibility

### DIFF
--- a/src/plugin/helpers/validation.ts
+++ b/src/plugin/helpers/validation.ts
@@ -2,7 +2,7 @@ import type { FieldValidationErrors, ValidationError } from "../elementSpec";
 import type { FieldNameToValueMap } from "../fieldViews/helpers";
 import type { FieldDescriptions } from "../types/Element";
 
-type Validator = (fieldValue: unknown) => ValidationError[];
+type Validator = (fieldValue: unknown, fieldName: string) => ValidationError[];
 
 export const createValidator = (
   fieldValidationMap: Record<string, Validator[]>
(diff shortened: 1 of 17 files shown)


src/plugin/helpers/validation.ts
```diff
@@ -19,46 +20,72 @@ export const createValidator = (
...
   if (typeof value !== "string") {
     throw new Error(`[maxLength]: value is not of type string`);
   }
   if (value.length > maxLength) {
-    return [`Too long: ${value.length}/${maxLength}`];
+    return [
+      {
+        error: `Too long: ${value.length}/${maxLength}`,
+        message:
+          customMessage ?? `Property is too long: ${value.length}/${maxLength}`,
```
jonathonherbert, line 56: Would it be worth us passing the field name in when validating, so we could  write `${fieldName} is too long` and make the default more usable?

jonathonherbert (approved):
LGTM! One suggestion.

src/plugin/helpers/validation.ts
SHession: Haha I had exactly that and took it out. I think it's a good idea, happy to add it back.

SHession: @jonathonherbert I've added this back in. I think I removed it as the field name may be both internal and none human friendly i.e. camel case, but I see little harm in having it as a default. Let me know what you think.

PR #107, opened 2021-09-09
feat: Image size warning

rhystmills:
## What does this change?

This replicates the image size warning for the Image element, which is present in the existing element in Composer. This warning occurs when the largest asset available has at least one dimension smaller than 460px.

There's quite a bit of code here to allow us remain type-safe while validating the image fields. Is this valuable enough to justify not simply asserting the type using `as`?

## How to test

Run `yarn start` locally and try using a crop with a dimension smaller than 460px.

## Images

| Too Small | Correct Size |
| --- | --- |
|  |  |

## Accessibility

### DIFF
--- a/src/elements/image/ImageElement.tsx
+++ b/src/elements/image/ImageElement.tsx
@@ -7,6 +7,7 @@ import {
   createValidator,
   htmlMaxLength,
   htmlRequired,
+  largestAssetMinDimension,
 } from "../../plugin/helpers/validation";
 import { createReactElementSpec } from "../../renderers/react/createReactElementSpec";
 import { ImageElementForm } from "./ImageElementForm";
@@ -102,5 +103,6 @@ export const createImageElement = (
       altText: [htmlMaxLength(1000), htmlRequired()],
       source: [htmlMaxLength(250), htmlRequired()],
       photographer: [htmlMaxLength(250)],
+      mainImage: [largestAssetMinDimension(460)],
     })
   );
--- a/src/elements/image/ImageElement.tsx
+++ b/src/elements/image/ImageElement.tsx
@@ -3,11 +3,11 @@ import React from "react";
 import { createCustomField } from "../../plugin/fieldViews/CustomFieldView";
 import { createFlatRichTextField } from "../../plugin/fieldViews/RichTextFieldView";
 import { createTextField } from "../../plugin/fieldViews/TextFieldView";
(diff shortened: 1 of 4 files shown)


src/plugin/helpers/validation.ts
```diff
@@ -19,6 +19,82 @@ export const createValidator = (
...
+  prop: Y
+): obj is X & Record<Y, unknown> => {
+  return Object.hasOwnProperty.call(obj, prop);
+};
+
+const isRecord = (value: unknown): value is Record<string, unknown> => {
+  return typeof value === "object" && value !== null;
+};
+
+const isValidImageAsset = (
```
SHession, line 40: I would recommend moving this logic into a file located within the image element folder, as it is only likely to be useful for that particular element. When the elements are split out, this divide should become clearer.

rhystmills: Sounds very sensible - will do so.

src/elements/image/ImageElement.tsx
```diff
@@ -83,6 +84,88 @@ export const createImageFields = (
...
+    hasOwnProperty(maybeImage, "fields") &&
+    isRecord(maybeImage.fields) &&
+    hasOwnProperty(maybeImage.fields, "width") &&
+    hasOwnProperty(maybeImage.fields, "height") &&
+    typeof maybeImage.fields.width === "number" &&
+    typeof maybeImage.fields.height === "number"
+  );
+};
+
+const validateAssets = (maybeAssets: unknown[]) => {
```
SHession, line 118: I've come up with a slightly cleaner way of doing this without this function but it isn't much of an improvement. Let me know if you're interested.

SHession (commented):
A few things to think about on this. I see your concerns about the code necessary to add typesafety, I think moving it into a separate file should help a lot. Functionally the check behaves well and displays sensibly! Let me know what you think of the comments before I +1

src/elements/image/ImageElement.tsx
rhystmills: Cheers for the review Sam - have implemented your other points, and definitely interested in looking into you suggestion.

SHession: Had a good discussion about this on a call. We have come up with a solution which is cleaner and more succinct which has now been commited.

SHession (approved):

PR #108, opened 2021-09-09
fix: Allow error messages to wrap beneath labels 

jonathonherbert:
## What does this change?

Allows error messages to wrap beneath labels when they exceed a single line.

Before:

Screenshot 2021-09-09 at 11 18 24

After:

Screenshot 2021-09-09 at 11 18 51

## How to test

Have a play with our elements. Do our error messages behave as expected? Is everything else as it should be?

## Dev notes

It'd be nice to specify elements programmatically (`${label} {` rather than `label {` in InputHeading), but we can't do that without adding a babel plugin. That's probably another PR.

### DIFF
--- a/src/editorial-source-components/CustomCheckbox.tsx
+++ b/src/editorial-source-components/CustomCheckbox.tsx
@@ -10,10 +10,8 @@ const parentStyles = css`
   white-space: nowrap;
   div {
     ${labelStyles}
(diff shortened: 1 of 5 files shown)


rhystmills (approved):
Tested locally and works very nicely, with all tests passing 👍

PR #109, opened 2021-09-09
fix: Image element layout changes

jonathonherbert:
## What does this change?

A few layout changes for the image element:

1) Moves the 'copy from altText' button inline with its label. 

To do this, I've changed the `label` prop to accept a`ReactNode`,  allowing us to pass arbitrary content in as a label. Because strings are `ReactNode`s, we can pass strings as usual.

2) Adjusts the photographer columns, which were not full width.

3) Changes buttons from 'primary' to 'secondary' to reflect [docs](https://theguardian.design/2a1e5182b/p/435225-button/b/86f344): 'use for actions that need to be clear, but aren’t the primary focus of a page.'

There are two more things that I think are worth changing:
- the 'weighting' field is not inline
- the spacing between 'alt text' and 'photographer' fields is not consistent (because flex/grid  elements [do not permit margin collapse](https://css-tricks.com/the-rules-of-margin-collapse/))

Both of these are difficult to solve without completely separating layout concerns from our elements (otherwise we're left struggling w/ 'first-child' style solutions, which don't play well w/ responsive layouts). This will be done in a separate PR, with an explanation.

Before:

Screenshot 2021-09-09 at 11 44 57

After:

Screenshot 2021-09-09 at 12 07 41

## How to test

Take a look at the layout. Does it reflect the screenshots above? Is there anything that's regressed in other elements?

### DIFF
--- a/src/editorial-source-components/InputHeading.tsx
+++ b/src/editorial-source-components/InputHeading.tsx
@@ -20,7 +20,7 @@ const Errors = ({ errors }: { errors: string[] }) =>
   !errors.length ? null : <Error>{errors.join(", ")}</Error>;
 
 type Props = {
-  label: string;
+  label: React.ReactNode;
   errors: string[];
 };
 
--- a/src/elements/image/ImageElementForm.tsx
+++ b/src/elements/image/ImageElementForm.tsx
@@ -1,8 +1,10 @@
 import { css } from "@emotion/react";
+import styled from "@emotion/styled";
+import { Button } from "@guardian/src-button";
+import { space } from "@guardian/src-foundations";
 import { SvgCamera } from "@guardian/src-icons";
 import { Column, Columns, Inline } from "@guardian/src-layout";
 import React from "react";
-import { Button } from "../../editorial-source-components/Button";
 import { Error } from "../../editorial-source-components/Error";
 import { FieldWrapper } from "../../editorial-source-components/FieldWrapper";
 import type {
@@ -38,6 +40,10 @@ type ImageViewProps = {
   errors: ValidationError[];
   field: CustomField<MainImageData, MainImageProps>;
 };
+
+const AltText = styled.span`
+  margin-right: ${space[2]}px;
+`;
 
 export const ImageElementTestId = "ImageElement";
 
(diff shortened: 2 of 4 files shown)


src/editorial-source-components/InputHeading.tsx
```diff
@@ -20,7 +20,7 @@ const Errors = ({ errors }: { errors: string[] }) =>
   !errors.length ? null : <Error>{errors.join(", ")}</Error>;
 
 type Props = {
-  label: string;
+  label: React.ReactNode;
```
dskamiotis, line 23: question: we are changing to `React.ReactNode` whilst still passing in a string from the various forms?

Im assuming this is being used to allow for a potential `null` or `undefined` prop value?
Just hoping Im on the right lines https://reactnative.dev/docs/react-node

src/elements/image/ImageElementForm.tsx
```diff
@@ -72,33 +74,37 @@ export const ImageElementForm: React.FunctionComponent<Props> = ({
           <FieldWrapper
             field={fields.altText}
             errors={errors.altText}
-            label="Alt text"
+            label={
+              <span>
+                <AltText>Alt text</AltText>
```
dskamiotis, line 79: Was this wrapper used to allow for customised styling?

dskamiotis (approved):
Really elelgant changes clear to see how these improve the overall look of the element 👍

src/editorial-source-components/InputHeading.tsx
jonathonherbert: It's true that this allows a null value, a string value, or a React element! Ultimately, I think it's just useful to provide a label 'slot' that the user can fill with whatever they want – text, additional content, etc.

It's required in this case to allow us to pass our `Copy from alt text` button in as part of the label content.

src/elements/image/ImageElementForm.tsx
jonathonherbert: The `span`? No, it's not used to do anything other than wrap the content in a single parent – perhaps a fragment (`<> </>`) would be a better choice here?

dskamiotis: I think as long as we have a consistent pattern throughout all components? Im assuming both are inline tags in their behaviour.

jonathonherbert: They are – but it's worth noting that `Fragment` in React doesn't actually resolve to a DOM node! A fragment is just a way of wrapping lists of DOM nodes to stop React complaining – it disappears in the final HTML:

```jsx
// This is what you're write in React
<>
  <div>Some</div>
  <div>list</div>
  <div>of</div>
  <div>things</div>
</>
```

```html
// This is what ends up on the page
<div>Some</div>
<div>list</div>
<div>of</div>
<div>things</div>
```

I'll change my implementation to use the fragment, the span isn't doing any useful work.

PR #110, opened 2021-09-09
fix: Add FieldVerticalLayout component and remove margins from inputs/labels

jonathonherbert:
## What does this change?

Adds a layout component to take care of spacing our fields vertically.

This was previously done with margins, which mostly worked – but because `flex` doesn't permit [margin-collapse](https://css-tricks.com/the-rules-of-margin-collapse/), we had a problem with fields meeting flex containers.

A single component in charge of vertical spacing makes easier to make global adjustments to layout.

Here's a fun spot-the-difference that I've ruined for everyone! 

Before:

Screenshot 2021-09-09 at 12 07 41

After: 

Screenshot 2021-09-09 at 12 46 16

## How to test

Besides the small change in the before/after, this should be a no-op. Do all the elements look as they should?

## Accessibility

### DIFF
--- a/src/renderers/react/customFieldViewComponents/CustomDropdownView.tsx
+++ b/src/renderers/react/customFieldViewComponents/CustomDropdownView.tsx
@@ -1,5 +1,4 @@
 import { CustomDropdown } from "../../../editorial-source-components/CustomDropdown";
-import { InputGroup } from "../../../editorial-source-components/InputGroup";
 import type { ValidationError } from "../../../plugin/elementSpec";
 import type { Options } from "../../../plugin/fieldViews/DropdownFieldView";
 import type { CustomField } from "../../../plugin/types/Element";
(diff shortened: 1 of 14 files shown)


src/renderers/react/customFieldViewComponents/CustomDropdownView.tsx
```diff
@@ -20,18 +20,16 @@ export const CustomDropdownView = ({
 }: CustomDropdownViewProps) => {
   const [selectedElement, setSelectedElement] = useCustomFieldState(field);
   return (
-    <InputGroup>
```
SHession, line 23: I'm seeing issue with this locally as the import hasn't been removed.

SHession (commented):
Looks like CI hasn't run for this PR, locally I'm having compile issues.

SHession (commented):
I've add some further issues testing this due to a compile issue that has come from Main.
image


I have also noticed this odd tickbox display. I'm not sure if it's representative: 
image

jonathonherbert:
Thanks @SHession – hopefully just fixed that tick.

The compile issue on main is solved by #120, and shouldn't be a problem for testing – the app should still compile – but if there are problems of that sort, lmk!

SHession (approved):
Tickbox looks better! I still can't test the image element as it's broken due to the validation but everything else appears to be working correctly.

jonathonherbert:
@SHession oh, good point – guess we need to get #123 in to fix? I'll merge this now.

PR #111, opened 2021-09-09
refactor: Removes the hasErrors function and state as it's no longer needed

SHession:
## What does this change?

This PR removes the `hasErrors` state and function as it has now been usurped by the validator function.

## How to test

This should be a no-op but make the code cleaner by removing redundant code.

The code is easier to work with, in accordance with our vision document.

Is there any chance we may need this?

## Accessibility


jonathonherbert (approved):
Great work 🗜️

PR #113, opened 2021-09-10
fix: Fixes validation error border on fields

SHession:
## What does this change?

This PR fixes a regression introduced by #111, which removed the red border from field when an error is found.

## How to test

Add an element with validation, if the validation fails, does the field highlight red?

## Accessibility

- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)


jonathonherbert (approved):
Well spotted, sorry I missed!

PR #117, opened 2021-09-10
feat: Move validation into field descriptions

SHession:
## What does this change?

This PR changes the field description API to allow the possibility of colocating the validation with the field. This should make the API easier to work with in line with our vision document goals.

Before:

```javascript
export const pullquoteFields = {
  html: createTextField({ isMultiline: true, rows: 4 }),
};

export const pullquoteElement = createReactElementSpec(
  pullquoteFields,
  (_, errors, __, fields) => {
    return <PullquoteElementForm errors={errors} fields={fields} />;
  },
  createValidator({ pullquote: [htmlRequired()] })
);
```
After: 

```javascript 
export const pullquoteFields = {
  html: createTextField({
    multilineOptions: { isMultiline: true, rows: 4 },
    validators: [htmlRequired()],
  })
};

export const pullquoteElement = createReactElementSpec(
  pullquoteFields,
  (_, errors, __, fields) => {
    return <PullquoteElementForm errors={errors} fields={fields} />;
  }
);
```

## How to test

This should be a no-op from a functional perspective but make the API easier to work with. 

Creating new elements and updating existing ones is intuitive and easy.

## Accessibility

### DIFF
--- a/src/elements/code/CodeElementSpec.tsx
+++ b/src/elements/code/CodeElementSpec.tsx
@@ -1,12 +1,16 @@
 import React from "react";
 import { createCustomDropdownField } from "../../plugin/fieldViews/CustomFieldView";
 import { createTextField } from "../../plugin/fieldViews/TextFieldView";
-import { createValidator, required } from "../../plugin/helpers/validation";
+import { required } from "../../plugin/helpers/validation";
 import { createGuElementSpec } from "../createGuElementSpec";
 import { CodeElementForm } from "./CodeElementForm";
 
 export const codeFields = {
-  html: createTextField({ isMultiline: true, rows: 4 }, true),
+  html: createTextField({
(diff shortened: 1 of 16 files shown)


jonathonherbert (approved):
Great work. Tested locally. Two thoughts:

- by [overloading our function](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) to handle both signatures `(Validator[] | Options)` and `(Options, Validator[])`, we can remove the necessity for inconsistent method signatures (some field creators have options objects that include a `validators` prop, some just accept validators as an argument
- it'd be lovely to improve on the 'unknown' type for validators, but I appreciate our hands are tied WRT our consumer code at present.

Both are probably better solved in additional PRs.

SHession:
@jonathonherbert, I agree with your points mentioned above and that they would be better resolved in separate PRs. In the meantime I can create cards to address these.

PR #118, opened 2021-09-13
fix: Use createGuElementSpec for flexible-model elements

jonathonherbert:
## What does this change?

Use `createGuElementSpec` for all flexible-model elements. Previously, this was limited to just code – now we use that method for the lot, ensuring that our data is in the correct shape for flexible-conte.

## How to test

Use this library in flexible-content. Does it work as expected?


rhystmills (approved):
Looking good 👍

SHession (approved):

PR #120, opened 2021-09-15
feat!: Remove transformer

jonathonherbert:
## What does this change?

Removes the transformer function inherent to prosemirror-elements, in favour of a solution that exists outside of the core library functionality.

This is because, in the view of the team (@SHession @rhystmills @dskamiotis), the amount of complexity this transform introduced to the codebase exceeded its utility. 

We intend to write a decision doc for this shortly, but because this is blocking critical path work, we'd like to ship this as a priority, alongside its replacement – accomplishing this work outside of the library in a different way.

## How to test

- The library should work as before in the UI, and the automated tests should pass.
- Any transformer-related code should no longer appear in the codebase.


SHession (approved):
This changes looks good and behaves well locally.

Do you have a sense that this will have the desired improvement of simplifying the codebase?

jonathonherbert:
@SHession good question. Yes, I think so, quite substantially in some places. We needed to carry `ExternalData` through a whole bunch of creation functions, and we lose a few conditional types.

PR #121, opened 2021-09-15
feat: Update copy for empty pullquote element error

rhystmills:
## What does this change?
Updates the copy for the empty pullquote element error, from: "Html is required" to "Pullquote cannot be empty".


dskamiotis (approved):
Looks good, passing the argument seems most sensible and efficient way of affecting the error output

PR #123, opened 2021-09-15
feat: Changes validators to accept a partial representation of the element

SHession:
## What does this change?

Building on the work from #120, this PR changes the validator type to accept a partial representation of the element data. This should allow us validate elements that are missing data. This will be helpfully when validating elements in Composer which currently model empty fields as undefined.

## How to test

This should be a no-op for the demo and I have added an additional unit test to check a partial element.

We are able to determine if elements are valid or not with a partial set of data.

## Accessibility

### DIFF
--- a/src/plugin/elementSpec.ts
+++ b/src/plugin/elementSpec.ts
@@ -35,7 +35,7 @@ export type ValidationError = {
 export type FieldValidationErrors = Record<string, ValidationError[]>;
 
 export type Validator<FDesc extends FieldDescriptions<string>> = (
-  fields: FieldNameToValueMap<FDesc>
+  fields: Partial<FieldNameToValueMap<FDesc>>
 ) => undefined | FieldValidationErrors;
 
 export type FieldValidator = (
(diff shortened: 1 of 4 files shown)


src/plugin/elementSpec.ts
```diff
@@ -35,7 +35,7 @@ export type ValidationError = {
 export type FieldValidationErrors = Record<string, ValidationError[]>;
 
 export type Validator<FDesc extends FieldDescriptions<string>> = (
-  fields: FieldNameToValueMap<FDesc>
+  fields: Partial<FieldNameToValueMap<FDesc>>
```
jonathonherbert, line 38: Torn – would it be worth adding a comment to indicate why this is a partial?

jonathonherbert (approved):
LGTM! Trying to think of wrinkles but I think the test covers it. One question re: documenting the rationale for partial. Arguably this PR is documentation enough.

src/plugin/elementSpec.ts
SHession: I think that's fair, it's not clear at a glance why that might be.

PR #124, opened 2021-09-20
feat: add `absentOnEmpty` option for text fields

jonathonherbert:
_co-authored-by: @rhystmills_ 

## What does this change?

Adds an `absentOnEmpty` option for text fields. Setting this option to `true` will not set a key-value pair on data returned from `getElementDataFromNode` if the field is empty.

Empty here means no text content, so a rich text field that serialises to e.g. `<p></p>` will be considered empty.

This change is required to better match the way absent data is modelled in Composer, the Guardian's content management system.

## How to test

The automated tests should pass, and test for the presence and absence of fields that use this flag, as well as the typesafety of the resulting data.

## Dev notes

The surface area of this work is slightly larger than might be expected for two reasons:

- it was better to refactor the options object of the text fields to incorporate `absentOnEmpty`, which touched every element that used them
- the type work to ensure the output of `getElementDataFromNode` was correct necessitated a few helper types.

### DIFF
--- a/src/elements/code/CodeElementSpec.tsx
+++ b/src/elements/code/CodeElementSpec.tsx
@@ -10,6 +10,7 @@ export const codeFields = {
     rows: 4,
     isCode: true,
     validators: [required("empty code field")],
+    absentOnEmpty: true,
   }),
   language: createCustomDropdownField("text", [
     { text: "Plain text", value: "text" },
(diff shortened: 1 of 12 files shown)


rhystmills (approved):
All makes sense to me. Running as expected locally and tests are passing 👍

PR #126, opened 2021-09-21
feat: Add external transformers

SHession:
## What does this change?

This PR adds the transformers back in to the project but outside of the ProseMirror Elements plugin and into a place we anticipate being moved out of this repo at some point. This keeps the transforms independent of the plugin and should reduce complexity whilst ensuring we still have the functionality needed to transform elements.

## How to test

I have added unit tests for the functions, are these comprehensive enough? We could also link the package to Composer for additional testing.

We can transform elements without adding complexity to ProseMirrorElements.

This repository follows the Editorial Tools accessibility guidelines: https://github.com/guardian/accessibility/blob/main/editorial-tools-guidelines.md
-->

### DIFF
--- a/src/elements/__tests__/transform.spec.ts
+++ b/src/elements/__tests__/transform.spec.ts
@@ -1,29 +1,29 @@
 import {
-  transfromElementIn,
-  transfromElementOut,
+  transformElementIn,
+  transformElementOut,
 } from "../transformer/Transform";
 
 describe("transform", () => {
-  describe("transfromIn", () => {
+  describe("transformIn", () => {
     it("should not allow elements which are the wrong type", () => {
-      // @ts-expect-error -- we should not be able to transfrom a malformed element
-      transfromElementIn("code", {});
+      // @ts-expect-error -- we should not be able to transform a malformed element
+      transformElementIn("code", {});
 
-      transfromElementIn("code", {
+      transformElementIn("code", {
         assets: [],
-        // @ts-expect-error -- we should not be able to transfrom a malformed element
+        // @ts-expect-error -- we should not be able to transform a malformed element
         fields: { nonExistantField: "123" },
       });
     });
     it("should partially transform elements with no fields", () => {
       const codeElement = { assets: [], fields: {} };
-      const result = transfromElementIn("code", codeElement);
+      const result = transformElementIn("code", codeElement);
 
       expect(result).toEqual({ assets: [] });
     });
     it("should partially transform elements with some fields", () => {
       const codeElement = { assets: [], fields: { html: "123" } };
-      const result = transfromElementIn("code", codeElement);
+      const result = transformElementIn("code", codeElement);
 
       expect(result).toEqual({ assets: [], html: "123" });
     });
@@ -32,22 +32,22 @@ describe("transform", () => {
         assets: [],
         fields: { html: "123", langaue: "HTML" },
       };
-      const result = transfromElementIn("code", codeElement);
+      const result = transformElementIn("code", codeElement);
 
       expect(result).toEqual({ assets: [], html: "123", langaue: "HTML" });
     });
   });
   describe("transformOut", () => {
     it("should not allow elements which are the wrong type", () => {
-      // @ts-expect-error -- we should not be able to transfrom a malformed element
-      transfromElementOut("code", {});
+      // @ts-expect-error -- we should not be able to transform a malformed element
+      transformElementOut("code", {});
 
-      // @ts-expect-error -- we should not be able to transfrom a malformed element
-      transfromElementOut("code", { html: "123" });
+      // @ts-expect-error -- we should not be able to transform a malformed element
+      transformElementOut("code", { html: "123" });
     });
     it("should completely transform elements with all fields", () => {
       const codeElement = { html: "123", language: "HTML" };
-      const result = transfromElementOut("code", codeElement);
+      const result = transformElementOut("code", codeElement);
 
       expect(result).toEqual({
         assets: [],
--- a/src/elements/__tests__/transform.spec.ts
+++ b/src/elements/__tests__/transform.spec.ts
@@ -1,0 +1,58 @@
+import {
+  transfromElementIn,
+  transfromElementOut,
+} from "../transformer/Transform";
+
+describe("transform", () => {
+  describe("transfromIn", () => {
+    it("should not allow elements which are the wrong type", () => {
+      // @ts-expect-error -- we should not be able to transfrom a malformed element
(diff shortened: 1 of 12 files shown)


src/elements/__tests__/transform.spec.ts
```diff
@@ -0,0 +1,58 @@
+import {
+  transfromElementIn,
```
jonathonherbert, line 2:
```suggestion
  transformElementIn,
```

...etc 😁

src/elements/__tests__/transform.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  transfromElementOut,
+} from "../transformer/Transform";
+
+describe("transform", () => {
+  describe("transfromIn", () => {
```
jonathonherbert, line 7: And again – just in case an F2 doesn't catch this in text!

src/elements/image/ImageElement.tsx
```diff
@@ -77,7 +77,7 @@ export const createImageFields = (
     source: createTextField({
       validators: [htmlMaxLength(250), htmlRequired()],
     }),
-    weighting: createCustomField("none-selected", [
+    role: createCustomField("none-selected", [
```
jonathonherbert, line 80: To align w/ flex image element?

src/elements/transformer/Transform.ts
```diff
@@ -0,0 +1,36 @@
...
+
+type TransformMap = typeof transformMap;
+type TransformMapIn<Name extends keyof TransformMap> = TransformMap[Name][0];
+type TransformMapOut<Name extends keyof TransformMap> = TransformMap[Name][1];
+
+export const transfromElementIn = <Name extends keyof TransformMap>(
+  elementName: Name,
+  values: Parameters<TransformMapIn<Name>>[0]
+): ReturnType<TransformMapIn<Name>> => {
+  const transformer = transformMap[elementName][0];
```
jonathonherbert, line 22: Can I recommend an object w/ `transformIn`, `transformOut` keys or similar rather than a tuple – just to make it clearer? Then we'd see e.g. `transformMap[elementName]["transformOut"]`, which I think is a little clearer.

src/elements/transformer/Transformer.ts
```diff
@@ -0,0 +1,20 @@
+import type { FieldNameToValueMap } from "../../plugin/fieldViews/helpers";
```
jonathonherbert, line 1: We've two files here, `transform` and `transformer` – this one seems to hold helper types, perhaps this could be called `types` or similar?

src/elements/transformer/DefaultTransform.ts
```diff
@@ -0,0 +1,47 @@
+import type { FieldNameToValueMap } from "../../plugin/fieldViews/helpers";
```
jonathonherbert, line 1: Files that aren't JSX components shouldn't need capital letters ... although we might be horribly inconsistent elsewhere 😬

src/elements/__tests__/transform.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+        assets: [],
+        // @ts-expect-error -- we should not be able to transfrom a malformed element
+        fields: { nonExistantField: "123" },
+      });
+    });
+    it("should partially transform elements with no fields", () => {
+      const codeElement = { assets: [], fields: {} };
+      const result = transfromElementIn("code", codeElement);
+
+      expect(result).toEqual({ assets: [] });
```
jonathonherbert, line 22: Should we strip `assets` on the way in here by default, to ensure pme doesn't see them? E.g. it's fair to say the pme code element doesn't care about `assets`.

jonathonherbert (commented):
A few minor things. This looks great.

As discussed, a useful next step will be sorting a broader definition of a `GuElement` so we can declare everything in a way that's colocated as close to the element as possible – that'll belong in another PR.

src/elements/image/ImageElement.tsx
SHession: Exactly!

src/elements/__tests__/transform.spec.ts
SHession: Took me quite a while to spot the difference there!

src/elements/transformer/Transform.ts
SHession: Good suggestion, I'll do that.

src/elements/transformer/Transformer.ts
SHession: Yeah I thought about that, I also considered adding a type folder. Happy with either really.

src/elements/transformer/DefaultTransform.ts
SHession: Yep good spot, I'll do make that change

src/elements/transformer/Transformer.ts
jonathonherbert: As am I 👍

src/elements/__tests__/transform.spec.ts
SHession: Is the suggestion here that if we do care about them, then the element will has a custom transform?

jonathonherbert: Yeah, I think so – difficult to imagine a `default` that knows where to put them? I may be wrong here!

SHession: Seems fair, I'll do that.

jonathonherbert (commented):
Changes look great. One more thing to address – if we link this in Composer, we get `Cannot read property 'in' of undefined` – because at the moment, we expect to find every element we pass into our transformers.

Solution is probably to should add a check to both transformers in `transform` that returns `undefined` if the elementName isn't present in our map.

src/elements/transformer/transform.ts
```diff
@@ -0,0 +1,40 @@
+import type { codeFields } from "../code/CodeElementSpec";
+import type { embedFields } from "../embed/EmbedSpec";
+import { transformElement as imageElementTransfrom } from "../image/imageElementDataTransformer";
```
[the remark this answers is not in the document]
SHession: Thanks, got it!

SHession:
> Solution is probably to should add a check to both transformers in `transform` that returns `undefined` if the elementName isn't present in our map.

I was thinking about that, we could alternatively filter the elements to ones PME knows about on the Composer die. Support ing the "undefined" transforms was part of what caused the complexity with the types.

jonathonherbert (approved):
Tested via yarn link, this works well! Nice 👍

PR #129, opened 2021-09-22
fix: Address problem w/ html tags in plain text field

jonathonherbert:
## What does this change?

#127 did not address a problem with HTML chars on ingress via `getNodeFromElementData` – we were using `innerHTML` to parse text.

This PR makes text nodes a special case on ingress (as #127 does on egress), manually creating text nodes rather than parsing as HTML.

As well as being more correct, this should also be quicker!

## How to test

I don't _think_ there's a way to access this behaviour via the demo UI.  But the automated tests should pass, and you should be convinced they test for the problem. I wrote the unit test in a failing state, and then fixed. The amended integration test is also good confirmation.

### DIFF
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -171,13 +171,13 @@ describe("ImageElement", () => {
         );
       });
 
-      it(`should remove marks when content is created with them`, () => {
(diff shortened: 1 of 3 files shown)


SHession (approved):
Tests look thorough and I've tried various scenarios locally. LGTM!

PR #131, opened 2021-09-23
feat: Changes ImageElement to accept a plugin creator prop

SHession:
## What does this change?

This PR allows a consumer of the image element to provide a custom set of plugins via a plugin creator function. This should mean we can use Composer's custom menu bar within the caption field.

## How to test

This should be a no-op as the demo instance will use the same plugins.

We are able to use the Composer menu bar with the image element.

## Accessibility

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -45,7 +45,10 @@ type Name =
 
 const { plugin: elementPlugin, insertElement, nodeSpec } = buildElementPlugin({
   demoImageElement: createDemoImageElement(onSelectImage, onDemoCropImage),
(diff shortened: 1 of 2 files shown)


jonathonherbert (approved):
LGTM. Worth naming this `createCaptionPlugins`, as IIUC that's the only field these plugins are applied to?

SHession:
> LGTM. Worth naming this `createCaptionPlugins`, as IIUC that's the only field these plugins are applied to?

Fair point, I did consider doing that but decided against it just in case we wanted to use it elsewhere. Your rationale makes sense and I'll change it.

PR #132, opened 2021-09-23
docs: Add ADR for element transformers

jonathonherbert:
## What does this change?

Adds an ADR for the element transformers work. After chatting with the team, this is arguably a bit extra – but it's done now!

## How to review and test

Have a read of the doc. Is it useful? Anything to change?


SHession (approved):
LGTM

PR #133, opened 2021-09-24
feat: Image element tweaks

SHession:
## What does this change?

This PR makes various tweaks to the image element, in order to more closely align it to the existing element.

These changes are:

1. Handling the undefined value of the "weighting" dropdown
2. Displaying a character count for the "caption" field
3. Handling assets with a width and heigh of string 
4. Updates button to use "Guardian Agate Sans"

## How to test

The tests should pass and the behaviour described above should work correctly in the demo and composer.

Our PME image element more closely matches the existing composer version.

## Images

image

## Accessibility

### DIFF
--- a/src/editorial-source-components/InputHeading.tsx
+++ b/src/editorial-source-components/InputHeading.tsx
@@ -1,5 +1,6 @@
 import styled from "@emotion/styled";
 import { space } from "@guardian/src-foundations";
+import { Description } from "./Description";
 import { Error } from "./Error";
 import { Label } from "./Label";
 
@@ -20,12 +21,17 @@ const Errors = ({ errors }: { errors: string[] }) =>
 
 type Props = {
   label: React.ReactNode;
+  description?: React.ReactNode;
   errors: string[];
 };
 
-export const InputHeading = ({ label, errors }: Props) => (
+export const InputHeading = ({ label, description, errors }: Props) => (
   <InputHeadingContainer>
     <Label>{label}</Label>
-    <Errors errors={errors} />
+    {errors.length > 0 ? (
+      <Errors errors={errors} />
+    ) : description ? (
+      <Description>{description}</Description>
+    ) : null}
   </InputHeadingContainer>
 );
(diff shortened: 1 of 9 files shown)


src/editorial-source-components/InputHeading.tsx
```diff
@@ -20,12 +21,17 @@ const Errors = ({ errors }: { errors: string[] }) =>
...
-export const InputHeading = ({ label, errors }: Props) => (
+export const InputHeading = ({ label, description, errors }: Props) => (
   <InputHeadingContainer>
     <Label>{label}</Label>
-    <Errors errors={errors} />
+    {errors.length > 0 ? (
+      <Errors errors={errors} />
+    ) : description ? (
+      <Description>{description}</Description>
+    ) : null}
```
jonathonherbert, line 35: Ah, I like this approach 👍 not quite convinced these will always be mutually exclusive, but let's find out 😁

src/elements/image/ImageElementForm.tsx
```diff
@@ -145,26 +152,32 @@ const ImageView = ({ field, onChange, errors }: ImageViewProps) => {
     onChange(mediaPayload);
   };
 
-  const getImageSrc = () => {
+  const imageSrc = useMemo(() => {
```
jonathonherbert, line 155: 👍 nice optimisation

jonathonherbert (approved):
LGTM. One question – where do we get the inconsistent types (`string`|`number`) from?

One thought – there's a fair amount of weird, domain specific logic building in our transformer for the image element. It might be nice to have tests, if only to document why we need the logic.

SHession:
> LGTM. One question – where do we get the inconsistent types (`string`|`number`) from?
> 
> One thought – there's a fair amount of weird, domain specific logic building in our transformer for the image element. It might be nice to have tests, if only to document why we need the logic.

It's a fair point and something I was a bit uncomfortable with. The fact is that setting it to a string only doesn't do much to reduce the complexity. That being said I could give it another pass to see if I can improve it.

jonathonherbert (approved):
Tests look good – I think that was the only blocker! Lmk if there's anything else in this PR that needs 👀

PR #134, opened 2021-09-27
feat: minor refactors re: feedback

jonathonherbert:
## What does this change?

- Reorders our consumer arguments in order of most common usage.
- Removes the requirement for an error object on FieldWrapper, avoiding boilerplate for fields that do not have validation.

## How to test

The refactored code should look cleaner, with fewer missed arguments. (There's also an argument for an options object for our consumer arguments, but we can tackle that another time!)

I haven't removed unused error objects in existing fields, because it's a bit risky – I think there's a strong case to colocate the errors on the existing `Field` objects, preventing these sorts of problems in the first place.

### DIFF
--- a/src/renderers/react/ElementProvider.tsx
+++ b/src/renderers/react/ElementProvider.tsx
@@ -101,12 +101,12 @@ export class ElementProvider<
     );
     return (
       <ElementWrapper {...this.state.commands}>
-        {this.props.consumer(
-          this.props.fields,
+        {this.props.consumer({
+          fields: this.props.fields,
           errors,
-          this.state.fieldValues,
-          this.updateFields
-        )}
+          fieldValues: this.state.fieldValues,
+          updateFields: this.updateFields,
+        })}
       </ElementWrapper>
     );
   }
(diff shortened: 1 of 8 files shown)


src/renderers/react/ElementProvider.tsx
```diff
@@ -102,10 +102,10 @@ export class ElementProvider<
     return (
       <ElementWrapper {...this.state.commands}>
         {this.props.consumer(
-          this.state.fieldValues,
```
SHession, line 105: Maybe we should pass this as an object instead of arguments?

jonathonherbert: As mentioned in PR description I was a bit torn ... yeah, let's do it.

SHession (approved):
I think this looks much better and sensible changes from the demo rehearsal. 

The only issues I could anticipate is it can be difficult to figure out what arguments a Consumer type will provide. Maybe we could provide a ConsumerOptions type to make it clearer? (This is a minor concern and non-blocking)

jonathonherbert:
Strangely, giving the type an alias doesn't seem to change the behaviour – possibly b/c at the point it's resolved into a React component the alias has disappeared? In both cases I get:

Screenshot 2021-09-28 at 09 47 04

(Hardly easy to read 😅 fair to say we can improve here!)

I think it's easier to read the code w/ an alias anyhow, so I've added that, thanks for the suggestion.

PR #135, opened 2021-09-28
fix: Fix validation bug

SHession:
## What does this change?

This PR aims to fix a small bug which was causing validation errors to be lost. This would happen if a field had multiple validators. These validators could then overwrite previous errors. 

This PR fixes the bug and adds additional unit tests to ensure there are no further issues.

NB: I think the API for the validator needs to be reconsidered. The format of errors we output is slightly confusing, occasionally outputting undefined or fields with an empty error array. In addition, the error and field errors don't currently play nicely. I think these are changes which can be handled under a separate PR.

## How to test

Are the unit tests thorough enough to give us confidence the change is working? Does the image element source character limit display as expected? 

Field validation errors are surfaced correctly.

## Accessibility

### DIFF
--- a/src/plugin/elementSpec.ts
+++ b/src/plugin/elementSpec.ts
@@ -1,4 +1,5 @@
 import type { FieldNameToValueMap } from "./fieldViews/helpers";
+import { validateWithFieldAndElementValidators } from "./helpers/validation";
 import type { CommandCreator, Commands } from "./types/Commands";
 import type {
   ElementSpec,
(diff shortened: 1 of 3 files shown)


rhystmills (approved):
Good work - all makes sense and the tests are a good addition.

All the tests passed and it's running as expected locally 👍

jonathonherbert:
Incidentally, agreed re: validation output – worth addressing down the line.

PR #137, opened 2021-09-29
feat: Add select all shortcut to TextField and RichTextField

rhystmills:
## What does this change?

This adds a 'select all' shortcut to the TextField and RichTextField.

We filter the `baseKeymap` provided by `prosemirror-commands` here (based on key name) - removing the enter-based shortcuts which may interfere with intended behaviour in Composer.

| Before | After |
| ------  | ----- |
|  |  |

## How to test

Run `yarn start` from the root directory locally (with nginx running) to run the application.

The new cypress tests should also pass - run them with `yarn test:integration` while the application is running in another terminal.

## Accessibility

- [X] [Navigable with keyboard](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#keyboard-navigation)
- [X] [The change doesn't use only colour to convey meaning](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#use-of-colour)


jonathonherbert (commented):
Ace! Could we add an integration test to avoid regressions? E.g. select all in a text field, check that `win.navigator.clipboard.readText()` contains the relevant text?

(We can timebox this, so if this is a pain to test, we can just 🚢 !)

rhystmills:
> Ace! Could we add an integration test to avoid regressions? E.g. select all in a text field, check that `win.navigator.clipboard.readText()` contains the relevant text?
> 
> (We can timebox this, so if this is a pain to test, we can just 🚢 !)

Good idea, I'll give it a go today.

cypress/tests/ImageElement.spec.ts
```diff
@@ -224,13 +238,27 @@ describe("ImageElement", () => {
         setDocFromHtml(values);
         assertDocHtml(getSerialisedHtml(values));
       });
+
+      it.only("should select all of its text when the select all shortcut is used", () => {
```
jonathonherbert, line 242: Don't forget to strip the `.only`s! Could have a CI linting rule, I do this all the time 😅

cypress/tests/ImageElement.spec.ts
```diff
@@ -224,13 +238,27 @@ describe("ImageElement", () => {
...
     });
 
     describe("Checkbox field", () => {
       it(`should be clickable`, () => {
-        addImageElement();
-        getElementField("useSrc").find("input").click();
-        getElementField("useSrc").find("input").should("be.checked");
+        addImageElement({});
+        getElementRichTextField("useSrc").find("input").click();
+        getElementRichTextField("useSrc").find("input").should("be.checked");
```
jonathonherbert, line 261: Is this change related?

cypress/tests/ImageElement.spec.ts
rhystmills: Whoops

cypress/tests/ImageElement.spec.ts
rhystmills: No - not sure how I managed that. Will fix.

jonathonherbert (approved):

PR #139, opened 2021-09-29
feat: Image element tweaks II

SHession:
## What does this change?

This PR builds on the work of #133 and implements a few missing bits of functionality for the image element. These features currently exist in the Composer version.

These are: 

1. Preserving field values if a new crop is for the same image
2. Restricting image role values to "thumbnail" if the crop is too small
3. Adding an error level to validation errors to be used in pre-publish checks

## How to test

For the image element:
1. Changing an image crop should preserve the field values if the image is the same, and replace them if it's different
2. Choosing a small crop should restrict the role value to "thumbnail"
3. Validation errors should appear as normal but have an associated level

Our PME image element closely matches the Composer version.

## Images

image

## Accessibility

### DIFF
--- a/src/elements/image/ImageElement.tsx
+++ b/src/elements/image/ImageElement.tsx
@@ -1,7 +1,10 @@
 import type { Schema } from "prosemirror-model";
 import type { Plugin } from "prosemirror-state";
 import React from "react";
-import { createCustomField } from "../../plugin/fieldViews/CustomFieldView";
+import {
+  createCustomDropdownField,
+  createCustomField,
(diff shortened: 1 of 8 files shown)


jonathonherbert (approved):
Code looks great. Tested in sandbox:

[x] Changing an image crop should preserve the field values if the image is the same, and replace them if it's different
[x] Choosing a small crop should restrict the role value to "thumbnail"
[x] Validation errors should appear as normal but have an associated level (verified via breakpoint)

PR #138, opened 2021-09-29
feat: Add placeholder option for text and rich text fields

jonathonherbert:
## What does this change?

Adds a placeholder option for text and rich text fields.

The placeholder option can be a string, which PME will use to populate a span element of its own styling – or a function that returns a span, which the user can style and add content to themselves.

Behold!

Screenshot 2021-09-29 at 12 27 31

NB: This does not address placeholders for accessibility purposes (e.g. `aria-placeholder`). We can address this is a separate PR.

## How to test

- Both unit and integration tests should pass, and convince you that the placeholders are working as expected.
- Have a play locally – does everything look OK?

## Accessibility

- [ ] [Tested with screen reader](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#screen-readers)
- [ ] [Navigable with keyboard](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#keyboard-navigation)
- [x] [Colour contrast passed](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#colour-contrast)
- [x] [The change doesn't use only colour to convey meaning](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#use-of-colour)
- [ ] [Interactive elements show a focus ring when focused by keyboard](https://github.com/guardian/accessibility/blob/main/editorial-tools-guidelines.md#appendix-the-focus-ring)
- [ ] [Semantic elements have been used where possible](https://github.com/guardian/accessibility/blob/main/editorial-tools-guidelines.md#appendix-semantic-html)

PR #140, opened 2021-09-30
fix: Adds placeholder text to existing elements

SHession:
## What does this change?

This PR uses the work from #138 to add placeholder text to the fields in the existing elements.

## How to test

Create an image and pullquote element in the demo environment. Does placeholder text appear when there is no content?
Does the placeholder look as it should?

## Images

Image:
image

Pullquote:
image

## Accessibility

There are placeholder features not yet implemented as mentioned in #138

### DIFF
--- a/src/elements/image/ImageElement.tsx
+++ b/src/elements/image/ImageElement.tsx
@@ -63,6 +63,7 @@ export const createImageFields = ({
     alt: createTextField({
       rows: 2,
       validators: [htmlMaxLength(1000), htmlRequired()],
+      placeholder: "Enter some alt text…",
     }),
     caption: createFlatRichTextField({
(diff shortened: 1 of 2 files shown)


jonathonherbert (approved):
Looks good!

paperboyo:
Just looking at the pic… Isn’t it a tad too dark (fields look like they are fille-in a bit)? My Firefox says #747373, maybe it should be #cdcdcd like the inactive buttons in the toolbar? (current PROD is #8d8d8c, I think, so in between)?

Also, maybe they should be Agate (like in PROD)? We could go philosophical here: should placeholders be in the font of the content (Egyptian) instead of the UI (Agate)? Or are they part of the UI indicating the lack of content? I just think Agate is less in yer face.

PR #141, opened 2021-09-30
fix: Fixes the placeholder text overflow UI bug

SHession:
## What does this change?

This PR fixes a placeholder text overflow issue, causing placeholder text to stretch beyond its bounds. This is achieved through some small styling tweaks.

## How to test

Load the demo page, does the placeholder test reasonably stay within its bounds?

## Images
Before:
image

After (is it me or is this an optical illusion?):

PME_OVerflow

## Accessibility

### DIFF
--- a/src/plugin/helpers/placeholder.ts
+++ b/src/plugin/helpers/placeholder.ts
@@ -7,9 +7,12 @@ export const placeholderTestAttribute = "placeholder";
 const getDefaultPlaceholder = (text: string) => {
   const span = document.createElement("span");
   span.style.display = "inline-block";
-  span.style.height = "0px";
-  span.style.width = "0px";
+  span.style.width = "90%";
   span.style.whiteSpace = "nowrap";
+  span.style.overflowX = "hidden";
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
Looks great – this is a better fix than I had in mind, I had forgotten about `textOverflow`!

Edit: could also sneak in an Agate font whilst we're here?

jonathonherbert (changes requested):
Haha, the integration tests actually caught a bug – our placeholder is now preventing clicks from focusing our input.

I would have thought `pointer-events: none` would solve, but ... it doesn't. Hm.

jonathonherbert (approved):
Looks good! I can't see what's changed – what did the trick?

SHession:
> Looks good! I can't see what's changed – what did the trick?

I was using position absolute to get around issues with the height of the input, this seemed to be the cause of the issue with being able to focus the input (still not sure why). Instead of using position absolute, I used vertical align to prevent descenders increasing the text height, which is a nicer solution.

PR #142, opened 2021-09-30
feat: Add typerighter support

jonathonherbert:
## What does this change?

Facilitate Typerighter support in consumers, by allowing us to
1. Identify nodes that represent fields, generally
2. Identify nodes that we'd like to check with Typerighter.

I've added this to our pullquote element for now – there's a minor decorations bug when suggestions are accepted that would be good to solve before we add this to images that I'll document elsewhere for a fix.

## Dev notes

We do 1.) with a group name, and 2.) with an attribute we pass via the nodeSpec attribute. I've added an object helper to help make this easy, but ultimately this API is a little obscure – however, it gets us off the ground without coupling the core prosemirror-elements library to typerighter.

## How to test

This is quite difficult to test in action, because `link`ing this library with `npm` or `yarn` in a consumer has a few side-effects:
- it means we must modify the nodeSpec in the consumer (see the bottom of the readme in this file for the fix)
- decoration display does not work when libraries are linked (again, for library reasons).

Typerighter CODE is also not working at present 🤦‍♂️

As a result, I've used our beta channel to test this in Composer, our CMS, and it's working nicely – note, just pullquote fields accept Typerighter checks at present:

Screenshot 2021-09-30 at 15 31 51

So to test, bump prosemirror-elements to the latest `beta` release, run Composer with the branch `jsh/add-typerighter-to-elements `, run a local version of Typerighter, and try this out with some elements – appreciate there's a fair amount of setup!


SHession (approved):
I think the API change look really sensible and requires a little less knowledge of ProseMirror than previously. I noticed you wrote "but ultimately this API is a little obscure", do you still feel this is the case?

I've tested this on the branch mentioned in the description and it works well except for one console error I noticed (which isn't strictly related to this PR):
image

jonathonherbert:
Thanks for the review, @SHession – I think this is a little less obscure now we're flattening these properties into the object – at least it's easy to pass around!

PR #143, opened 2021-10-05
fix: Fix decoration jitter

jonathonherbert:
## What does this change?

Fixes an odd issue where decorations passed to child editors might appear in the wrong place, or not appear at all, in certain situations.

This was caused by the code responsible for applying the decorations reading from class properties that had yet to be updated. A cautionary tale about mutating and reading from local state!

## How to test

The added integration tests should pass. Take a look – do they clearly describe what they're testing?

To test manually, see the gifs below – adding and removing text adjacent to decorations, either in the same field or in a subsequent field, had odd results, as they were working with a stale offset and node.

| Before | After |
| --- | --- |
|  |  |

## Accessibility

N/A

### DIFF
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -68,10 +68,60 @@ describe("ImageElement", () => {
           .should("have.text", "deco");
       });
 
-      it(`caption – should map decorations passed from the parent editor correctly when they move`, () => {
+      it(`caption – should map decorations passed from the parent editor correctly when they move – add text before`, () => {
         addImageElement();
-        const text = `caption deco{leftarrow}{leftarrow}{leftarrow}{leftarrow}{leftarrow} more text`;
-        typeIntoElementField("caption", text);
+        typeIntoElementField(
+          "caption",
+          "start deco{leftarrow}{leftarrow}{leftarrow}{leftarrow}{leftarrow} more"
+        );
+        getElementRichTextField("caption")
+          .find(".TestDecoration")
+          .should("have.text", "deco");
+      });
+
+      it(`caption – should map decorations passed from the parent editor correctly when they move – add text after`, () => {
+        addImageElement();
+        typeIntoElementField("caption", "caption deco more text");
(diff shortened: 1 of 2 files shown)


src/plugin/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -108,14 +108,14 @@ export abstract class ProseMirrorFieldView implements FieldView<string> {
     const node = this.getNodeFromValue(value);
     const tr = this.innerEditorView.state.tr;
     tr.replaceWith(0, this.innerEditorView.state.doc.content.size, node);
-    this.onInnerStateChange(tr);
+    this.dispatchTransaction(tr);
```
jonathonherbert, line 111: Renamed to better reflect its role.

cypress/tests/ImageElement.spec.ts
```diff
@@ -68,10 +68,60 @@ describe("ImageElement", () => {
...
+      it(`caption – should map decorations passed from the parent editor correctly when they move – add text in following field`, () => {
+        addImageElement();
+        typeIntoElementField("caption", "caption deco");
+        typeIntoElementField("altText", "text");
+        getElementRichTextField("caption")
+          .find(".TestDecoration")
+          .should("have.text", "deco");
+      });
+
+      it(`caption – should map decorations passed from the parent editor correctly when they move – remove text in following field`, () => {
```
SHession, line 121: Is it worth adding test cases where we might have a decoration with a range which might change? We could then add characters within that range and outside of it, to ensure it handles those cases correctly.

SHession (approved):
Changes look sensible and the tests look like a good addition. I wasn't able to test the decorations as I wasn't sure how to add them in the demo environment.
Just to confirm, this issue was just a result of the function being called out of order with the class property assignments?

cypress/tests/ImageElement.spec.ts
jonathonherbert: Can't hurt! Followed up in #145.

PR #144, opened 2021-10-05
feat: Add typerighter attrs to image caption and alt text

jonathonherbert:
## What does this change?

Adds Typerighter attrs to the image caption.

## How to test

Run locally. Is the `useTyperighter` attribute set for the image caption node?

Screenshot 2021-10-05 at 09 54 02

## Accessibility

N/A

### DIFF
--- a/src/plugin/fieldViews/RichTextFieldView.ts
+++ b/src/plugin/fieldViews/RichTextFieldView.ts
@@ -64,9 +64,7 @@ type FlatRichTextOptions = RichTextOptions & {
  */
 export const createFlatRichTextField = ({
   createPlugins,
-  validators,
-  placeholder,
-  marks,
+  ...rest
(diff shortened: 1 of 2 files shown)


src/plugin/fieldViews/RichTextFieldView.ts
```diff
@@ -91,6 +93,8 @@ export const createFlatRichTextField = ({
     validators,
     placeholder,
     marks,
+    attrs,
+    absentOnEmpty,
```
jonathonherbert, line 97: Added as this wasn't being passed through, and will be inevitably needed at some point!

SHession (approved):
Tested this locally in Composer using the associated [PR](https://github.com/guardian/flexible-content/pull/3907). It worked as expected with element matches appearing in the TR panel and navigating correctly to the element.

PR #155, opened 2021-10-19
fix: Fix insert element bug

SHession:
## What does this change?

This PR attempts to fix an issue whereby inserting new elements with a selection inside an existing element, would cause it to be split in two.  This PR changes how elements are inserted, attempting to determine the correct position for an element to be added and then inserting it.

NB: In implementing this PR, an issue was discovered where the selection state is not being properly updated between PME and its parent instance. A new card has been created to address this. (Now, a PR: #158) If issues are detected with this change, please check the selection state is being updated properly by using the PM debug tools.

## How to test

See the GIF for previous replication steps. Is it possible to split an element by inserting a new one? Does inserting a new one always behave as expected.

Inserting elements works as intended.

## Images

PME_bug

## Accessibility

N/A


jonathonherbert:
I've added some unit tests to prove this behaviour, and I reckon this is now ready for review. We could potentially refactor so the helper function handles both cases (position w/in element, and position w/out – the latter is currently handled within `insertElement`) but suspect this will be fine for now.

rhystmills (approved):
I've tested locally and all seems to be working as expected.

Tests are also passing, and the code all looks good 👍

PR #156, opened 2021-11-29
feat: Handle hyphens in element names

jonathonherbert:
_co-authored-by: @dskamiotis_

## What does this change?

Introduces a transformation between an element name and its name as a node. This allows us to use characters that are reserved in [content expressions](https://prosemirror.net/docs/guide/#schema.content_expressions), like hyphens.

## How to test

- The automated tests should pass, and convince you they test an appropriate surface area.
- Take a look at the demo. The demo element has been renamed to `demo-image-element`. It should work as before.

We can use reserved chars in node names.

This is a bit of a tricky change. I can think of three ways of doing this:

1. Don't. Ask consumers to do this transformation. Warn explicitly when these chars are used when constructing elements, rather than letting Prosemirror blow up.
2. (This PR): Transform every time we cross the node-elementName boundary. This means a few little interventions every time this happens (there are three at present: one when we create the nodespec, one when we create the  nodeview, and one for the boundary between element data and node data). There's no typesafety to ensure this is respected, so this relies on good tests against this edge-casey behaviour.
3. Transform things once, when we first receive the element map. This is the safest change from a code point of view, but is pretty misleading re: internal types – we must figure out a template literal type that transforms our element name strings at the type level, and use them for our internal spec map, adding further complexity to our type system, which is already pretty ... busy.

I've chosen 2, because I think it represents a good balance between hiding complexity from our consumers, and managing complexity internally, but I'd be interested to hear other thoughts.


dskamiotis (approved):
Great solution @jonathonherbert 
Tested successfully for presence of demo-image-element
Additionally via `yalc` with the proposed `rich-link` element!
👏  🙌

PR #157, opened 2021-12-01
feat: Add alt text tooltip

rhystmills:
## What does this change?
This PR adds a tooltip to the image element which explains the purpose of alt text. It also links to an external document that goes into more detail about writing good alt text, with examples.

### Why?
A user noticed an article (of mostly images) had a '.' for all its alt text fields - a user circumventing the minimum character rule for alt text in Composer.

Any minimum character rule can be worked around, so it was decided that adding a tooltip explaining that alt text is valuable for users of assistive technology would be a good behavioural nudge, and useful to users of Composer who might want a refresher on its purpose.

I ran a query on our database and found that, of 1,499,519 images used in articles, 16,050 had alt text of less than 5 characters (which are unlikely to be good image descriptions) - just over 1%. I felt that this was a significant enough proportion to justify trying the tooltip.

The Production team also thought it was a sensible addition.

### Where did the guidance come from?
The guidance can be found [here](https://docs.google.com/document/d/1oW542iCRyKfI4DS22QU7AH0TQRWLYMm7bTlhJlX5_Ng/edit). The guidance is a collaboration between Kirsten Broomhall and I, based on existing guidance given to journalists, and my input on accessibility. We ran it past Jonathan Casson (Head of production) who thought it was sensible guidance.

### Why a google doc?
It turns out that most guidance shared with journalists is done via Google docs, so I stuck with that for the sake of using a familiar medium. Subeditors have edit rights.

(N.B. I wonder if an internal microsite with information for journalists would be a good idea - but there doesn't seem to be anything like this at the moment).

## How to test

Run `yarn start` locally in the repository root, and add an image element. Try hovering over the 'i' icon and using the link.

### DIFF
--- a/src/editorial-source-components/Tooltip.tsx
+++ b/src/editorial-source-components/Tooltip.tsx
@@ -1,4 +1,5 @@
 import { css } from "@emotion/react";
+import styled from "@emotion/styled";
 import { brand, neutral } from "@guardian/src-foundations";
 import { SvgInfo } from "@guardian/src-icons";
 import React, { useEffect, useState } from "react";
@@ -21,7 +22,34 @@ const fadeDuration = 300; //Milliseconds
 
 const timeouts: number[] = [];
 
-const tooltip = css`
+const Arrow = styled.div`
+  position: absolute;
+  width: 8px;
+  height: 8px;
+  background: inherit;
+  visibility: hidden;
+  &[data-show] {
+    ::before {
+      visibility: visible;
+    }
+  }
+  &[data-popper-reference-hidden] {
+    ::before {
+      visibility: hidden;
+      pointer-events: none;
+    }
+  }
+  ::before {
+    position: absolute;
+    width: 8px;
+    height: 8px;
+    background: inherit;
+    content: "";
+    transform: rotate(45deg);
+  }
+`;
+
+const TooltipBox = styled.div`
   font-family: "Guardian Agate Sans";
   background-color: ${neutral[10]};
   color: ${neutral[97]};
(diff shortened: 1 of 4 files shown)


src/elements/image/ImageElementForm.tsx
```diff
@@ -56,93 +57,112 @@ export const ImageElementForm: React.FunctionComponent<Props> = ({
...
+            />
+            <FieldWrapper
+              field={fields.alt}
+              errors={errors.alt}
+              headingLabel={<AltText>Alt text</AltText>}
+              headingContent={
+                <>
+                  <Tooltip>
+                    <p>
+                      'Alt text' describes what's in an image. It helps users of
```
paperboyo, line 109:
```suggestion
                      ‘Alt text’ describes what’s in an image. It helps users of
```

paperboyo:
👏 👏 👏 
cc @SiAdcock

SiAdcock:
Amazing work, thanks so much @rhystmills and the team!

src/editorial-source-components/Tooltip.tsx
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }, [opaque]);
+
+  return (
+    <>
+      <div
+        css={infoIcon}
+        ref={setReferenceElement}
+        onMouseEnter={() => {
+          makeOpaque();
```
jonathonherbert, line 133: Might be worth tucking these into functions that can be re-used, as AFAICS we're repeating ourselves below.

jonathonherbert (commented):
Outstanding, this is lovely to have. A few minor comments, and one thing I've spotted – Popper will move tooltips when there's not enough space for the usual placement, but our triangle doesn't currently account for that (there's not enough space to render the tooltip above):

Screenshot 2021-12-02 at 13 39 48

src/editorial-source-components/Tooltip.tsx
rhystmills: Good idea, have done so now.

rhystmills:
> Outstanding, this is lovely to have. A few minor comments, and one thing I've spotted – Popper will move tooltips when there's not enough space for the usual placement, but our triangle doesn't currently account for that (there's not enough space to render the tooltip above):
> 
> Screenshot 2021-12-02 at 13 39 48

Good spot, thanks for being diligent - not sure how I missed that one! I've managed to sort it now.

src/editorial-source-components/Tooltip.tsx
```diff
@@ -145,24 +158,23 @@ export const Tooltip = ({ children }: { children: React.ReactNode }) => {
...
         onMouseEnter={handleMouseEnter}
         onMouseLeave={handleMouseLeave}
       >
         {children}
-        <div
+        <Arrow
           ref={setArrowElement}
           style={styles.arrow}
-          css={arrow}
+          className={"arrow"}
```
jonathonherbert, line 174: Unbearably minor nit: as a string, this can be a literal in the template, we shouldn't need curlies

jonathonherbert (approved):
Perfect. Great work.

PR #158, opened 2021-12-01
fix: Update selection manually when nested text fields are focused

jonathonherbert:
## What does this change?

Updates the editor selection when the user focuses empty Prosemirror-based fields.

We must do this manually, because Prosemirror does not dispatch a transaction when this occurs in an empty editor. This makes sense for a root editor, because it's effectively a no-op from a state point of view – but for nested editors, this does result in a change of state!

## How to test

- The new integration test should pass, and you should be convinced it reflects the correct behaviour.
- Open the demo page, create an element with a text field, open the dev tools, and click into empty text fields. You should see the selection update.

### DIFF
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -52,10 +52,15 @@ describe("ImageElement", () => {
         focusElementField("caption");
         focusElementField("src");
 
+        getDocSelection().then(([from]) => {
+          // The selection should have moved from the field
+          expect(from).to.not.equal(12);
+        });
+
         // We click to ensure we select the end of the text, where the cursor would have previously been
         getElementRichTextField("caption").click();
         getDocSelection().then(([from, to]) => {
-          // The selection should be at the first field of the element
+          // The selection should be at the end of the field
           expect(from).to.equal(12);
           expect(to).to.equal(12);
         });
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -49,7 +49,7 @@ describe("ImageElement", () => {
         getElementRichTextField("caption").should("have.text", text);
       });
 
-      it.only(`caption – should allow mark shortcuts in an element`, () => {
+      it(`caption – should allow mark shortcuts in an element`, () => {
         addImageElement();
         const text = `text with ${boldShortcut()}bold and ${boldShortcut()}${italicShortcut()}italic`;
         typeIntoElementField("caption", text);
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -31,7 +31,7 @@ describe("ImageElement", () => {
     describe("Rich text field", () => {
       it("should update the document selection when the user focuses on the field", () => {
         addImageElement();
-        assertDocSelection(19, 19); // The selection immediately after the inserted element
+        assertDocSelection(21, 21); // The selection immediately after the inserted element
         focusElementField("caption");
         assertDocSelection(5, 5);
       });
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -2,10 +2,10 @@ import { UpdateAltTextButtonId } from "../../src/elements/demo-image/DemoImageEl
 import {
   addImageElement,
   assertDocHtml,
-  assertDocSelection,
   boldShortcut,
   changeTestDecoString,
   focusElementField,
+  getDocSelection,
   getElementField,
   getElementMenuButton,
   getElementRichTextField,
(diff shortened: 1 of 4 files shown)


rhystmills:
I'm getting a strange issue on this branch - there seem to be fewer integration tests than usual, and the 'Run 3 integration specs' button in Cypress only seems to be running a single test.

cypress/tests/ImageElement.spec.ts
```diff
@@ -41,13 +49,13 @@ describe("ImageElement", () => {
         getElementRichTextField("caption").should("have.text", text);
       });
 
-      it(`caption – should allow mark shortcuts in an element`, () => {
+      it.only(`caption – should allow mark shortcuts in an element`, () => {
```
rhystmills, line 52:
```suggestion
      it(`caption – should allow mark shortcuts in an element`, () => {
```

There's the culprit.

jonathonherbert: Thanks for spotting 😅

jonathonherbert:
That should now be sorted – but I've realised that assert on a precise position in this spec is likeely to be fragile. I'll have a try at make this is a bit more robust tomorrow.

SHession:
This possibly beyond the intended scope of this PR, but I have found other situations where the selection state doesn't appear to update as we'd expect. In particular, selecting the start or end of some content in a text field.

https://user-images.githubusercontent.com/4633246/145393917-4432a107-fa57-41af-b5c6-f228dc3c4d32.mov

src/plugin/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -312,6 +313,29 @@ export abstract class ProseMirrorFieldView implements FieldView<string> {
...
+        return;
+      }
+      if (this.innerEditorView.state.doc.textContent.length !== 0) {
+        return;
+      }
+      const { tr, doc } = this.innerEditorView.state;
+      tr.setSelection(TextSelection.create(doc, 0));
+      // Setting a text selection seems to clear out our stored marks,
+      // so we must add them to the transaction explicitly.
+      tr.setStoredMarks(this.innerEditorView.state.storedMarks ?? []);
```
SHession, line 333: Unfamiliar with `storedMarks`, what would be the consequence of them being removed?

cypress/tests/ImageElement.spec.ts
```diff
@@ -27,6 +29,19 @@ describe("ImageElement", () => {
 
   describe("Fields", () => {
     describe("Rich text field", () => {
+      it("should update the document selection when the user focuses on the field", () => {
+        addImageElement();
+        getDocSelection().then(([from]) => {
+          // The selection should be just after the inserted element
+          expect(from).to.be.greaterThan(5);
+        });
+        focusElementField("caption");
```
SHession, line 38: Could we make this test a little more thorough by also selecting another empty text field?

SHession (commented):
Changes make sense and look good. A couple of comments to think about. I think this change works as intended but there may be other issues with selections. Waiting for thoughts before +1ing.

src/plugin/fieldViews/ProseMirrorFieldView.ts
jonathonherbert: Stored marks [store styling information to be applied on the next input](https://prosemirror.net/docs/ref/#state.EditorState.storedMarks). It's the state that stores e.g. 'impending boldness' – for example, when you hit a keyboard shortcut or click a menu item to apply bold styling, the editor stores that state so bold styling is applied on the next keypress.

SHession: Thanks @jonathonherbert, that makes sense. Is there any particular reason to keep them, would it not be sensible to have them removed when clicking into a blank editor?

jonathonherbert: Ah, I didn't address your second point. Clicking into a blank field with 'bold' styling queued up (in storedMarks) and losing that styling as a result of the click would be incorrect behaviour. I only caught it 'cos an integration test failed 😅

SHession: I would have thought it would desirable to remove the stored mark if I focus on an empty editor. I would think it's a new selection, separate to where you have queued up the styling?

jonathonherbert: Discussed out of band – turns out stored marks were altered unnecessarily here, a good spot 🙏

cypress/tests/ImageElement.spec.ts
```diff
@@ -42,6 +42,24 @@ describe("ImageElement", () => {
...
       });
+      it("should update the document selection when the user focuses on the field and it contains content, even if the inner field's selection has not changed", () => {
+        addImageElement();
+        getDocSelection().then(([from]) => {
+          // The selection should be just after the inserted element
+          expect(from).to.be.greaterThan(13);
+        });
+        typeIntoElementField("caption", "Example");
+        focusElementField("caption");
+        focusElementField("src");
```
SHession, line 53: `nit` we could make this test more robust by asserting that clicking the field, updates the selection. Otherwise, we can't be sure the selection has changed between at all been editing text and focusing fields.

SHession (approved):
Tested locally and the selection state updates as intended (video issue fixed). It would be nice to document the concerns around a duplicate state select transaction. Other than that LGTM.

jonathonherbert:
Good point, added a comment.

PR #161, opened 2021-12-02
feat: Add resizeable property to text and richText fields, and apply to image altText

jonathonherbert:
## What does this change?

Adds an `isResizeable` property to text and richtext fields, to allow both sorts to be resizeable by users in the usual way. This PR also applies this to our image alt text, to reflect the current behaviour of the old image element:

resize

There's one decision I've made here that's worth mentioning – when an input is resizeable, we default its initial size to the number of rows specified in the `rows` property. I thought this made a sort of heuristic sense, as whenever we make something resizeable, we're likely expecting lots of content in a limited space – but others may disagree! We can use `maxRows` to ensure that users can't expand inputs indefinitely if required.

Incidentally, I was dusting off the HTML5 drag and drop API docs before I found out this was possible with CSS – phew! 😅 

## How to test

- Drag is difficult to test in integration directly, so I've added integration tests to check the appropriate css property is set, and to check that the height behaviour above is working correctly.
- Try the image element shown in the GIF. Does it work as expected?

### DIFF
--- a/src/elements/image/ImageElement.tsx
+++ b/src/elements/image/ImageElement.tsx
@@ -61,7 +61,7 @@ export const createImageFields = ({
 }: MainImageProps) => {
   return {
     alt: createTextField({
-      rows: 4,
+      rows: 2,
       validators: [htmlMaxLength(1000), htmlRequired()],
       placeholder: "Enter some alt text…",
       isResizeable: true,
--- a/src/elements/image/ImageElement.tsx
+++ b/src/elements/image/ImageElement.tsx
@@ -61,9 +61,10 @@ export const createImageFields = ({
 }: MainImageProps) => {
   return {
     alt: createTextField({
-      rows: 2,
+      rows: 4,
       validators: [htmlMaxLength(1000), htmlRequired()],
       placeholder: "Enter some alt text…",
+      isResizeable: true,
     }),
     caption: createFlatRichTextField({
       createPlugins: createCaptionPlugins,
--- a/src/plugin/fieldViews/TextFieldView.ts
+++ b/src/plugin/fieldViews/TextFieldView.ts
@@ -37,6 +37,7 @@ type TextFieldOptions = {
   validators?: FieldValidator[];
   placeholder?: PlaceholderOption;
   attrs?: Record<string, AttributeSpec>;
+  isResizeable?: boolean;
 };
 
 export const createTextField = (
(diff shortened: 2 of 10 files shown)


src/plugin/fieldViews/TextFieldView.ts
```diff
@@ -137,18 +149,26 @@ export class TextFieldView extends ProseMirrorFieldView {
...
 
         if (maxRows) {
-          (this.innerEditorView.dom as HTMLDivElement).style.maxHeight = `${
+          const maxHeightPx = `${
             parseInt(lineHeight, 10) * maxRows + parseInt(paddingTop)
           }px`;
-          (this.innerEditorView.dom as HTMLDivElement).style.overflowY =
-            "scroll";
+          domElement.style.maxHeight = maxHeightPx;
+          domElement.style.overflowY = "scroll";
```
dskamiotis, line 171: Great simple implementation with HTML 5, as not aware of this and expected some fancy js!

src/elements/image/ImageElement.tsx
```diff
@@ -61,9 +61,10 @@ export const createImageFields = ({
 }: MainImageProps) => {
   return {
     alt: createTextField({
-      rows: 2,
+      rows: 4,
```
dskamiotis, line 64: Increasing the default a result of 
`something resizeable, we're likely expecting lots of content in a limited space`?

dskamiotis (approved):
Really useful addition @jonathonherbert, and elegant simple solution with the HTML 5.

Tested successfully and working as expected locally in Composer, linking your branch via yalc.
One possible side effect, it also makes the width expand the bounds of the element body area.

rhystmills:
Amazing, I never knew that property existed - this will save me a lot of pain in the future, when I inevitably would have sunk many hours trying to build a custom React component.

I suspect a few of our other elements would benefit from something like this - off the top of my head, the code field in the code element and the preview in the embed element,.

dskamiotis:
@rhystmills good thinking, definitely will be applicable and easily implemented in some current/future elements 👍

src/elements/image/ImageElement.tsx
jonathonherbert: Yeah, precisely – but maybe it's now a little big?

dskamiotis: I think we could go for 3 as default middle ground :)
Alternatively I don't think it makes a huge difference IMO

jonathonherbert: Tested inn Composer, 2 is better – thanks for mentioning 👍

jonathonherbert:
@rhystmills good shout, I've added that and tested in Composer 👍

PR #165, opened 2021-12-13
feat: Resize embed preview on content height change, allow manual resize over 100px

rhystmills:
## What does this change?
This PR changes the height of the preview in the embed element.

Instead of having a fixed size, the preview dynamically resizes based on its content, up to a size of 100px. After reaching that cap, the preview can be manually resized further. (It will remain in resizeable mode until the page is refreshed)

## How to test
Run `yarn start` from this repository locally. Add some html to the `embed code` field. Does the preview appropriately resize?

### DIFF
--- a/src/elements/embed/embedComponents/Preview.tsx
+++ b/src/elements/embed/embedComponents/Preview.tsx
@@ -1,9 +1,9 @@
 import { css } from "@emotion/react";
-import { useRef, useState } from "react";
+import React, { useRef, useState } from "react";
 import { Label } from "../../../editorial-source-components/Label";
 import { unescapeHtml } from "./embedUtils";
 
-export const Preview = ({ html }: { html: string }) => {
+export const Preview = React.memo(({ html }: { html: string }) => {
   const [height, setHeight] = useState("0px");
   const [resize, setResize] = useState(false);
   const ref = useRef<HTMLIFrameElement>(null);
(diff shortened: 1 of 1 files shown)


src/elements/embed/embedComponents/Preview.tsx
```diff
@@ -1,17 +1,43 @@
...
+import { useRef, useState } from "react";
 import { Label } from "../../../editorial-source-components/Label";
 import { unescapeHtml } from "./embedUtils";
 
-const iframe = css`
-  background-color: white;
-  width: 100%;
-`;
-
 export const Preview = ({ html }: { html: string }) => {
```
jonathonherbert, line 6: Can we use React.memo to prevent this component rerendering if `html` doesn't change? Aware that unescapeHtml might be doing some 💪 work 💪 here!

jonathonherbert (approved):
Working as expected! Whilst we're taking a look at preview, one minor styling comment: do we need the white background (and incredible 3d border 😁 )? Composer does it thus:

Screenshot 2021-12-13 at 16 28 45

(Oops, that was meant to be a comment, not an approve.)

rhystmills:
I did choose to make the Preview more distinct as I didn't feel it was clearly demarcated before - whether the incredible 3d border is necessary: probably not - should we get some UX perspective?

jonathonherbert (approved):

PR #166, opened 2022-01-05
perf: Avoid updating consumers and fieldViews if nothing has changed

jonathonherbert:
## What does this change?

At the moment, we do a lot of unnecessary work updating fieldViews and consumers (e.g. React components) when that's not necessary – updating both consumer and fieldView whenever a position, value, or decoration has changed. There are three things that these entities care about:

1. field values (both)
2. decorations (fieldViews)
3. commands (consumer)

This PR adds some logic to optimise our calls to these two entities based on this logic. We can take advantage of Prosemirror's immutable `Nodes` and `Decorations` to detect 1. and 2., and check the output of our commands to detect 3.

I've added comments to explain this in `plugin.ts`, which will hopefully explain the logic in a way which won't go stale – please consider that as part of your review!

## Why does this matter? Isn't premature optimization [the root of all evil?](https://www.explainxkcd.com/wiki/images/d/d5/optimization.png)

A fair question! In this case, I think this is warranted, urgent even. Consider these values, captured with Chrome dev tools. They measure the amount of JS time spent typing ten characters into a document, in ms, averaged across three attempts:

|  | Before | After |
|---|---|---|
| Empty doc | 101 | 110 |
| 10 image elements | 543 | 289 |
| 100 image elements | 12439 | 1423 |

(These are pessimistic values – in production, neither the Prosemirror devTools nor unoptimised collaborative editing plugins will be running.)

## How to test

- Have a play – do you see the same performance gains?

Better performance when using prosemirror-elements, especially with large numbers of elements in documents.


src/renderers/react/ElementProvider.tsx
```diff
@@ -57,63 +56,63 @@ export class ElementProvider<
...
 
-  onStateChange(): void {
+  public onStateChange(): void {
     this.props.onStateChange(this.state.fieldValues);
   }
 
-  updateState(state: Partial<IState<FDesc>>, notifyListeners: boolean): void {
-    if (
-      !isEqual(state.fieldValues, this.state.fieldValues) ||
-      (state.commands && state.commands.pos != this.state.commands.pos)
```
SHession, line 82: It's great to be able to remove this!

src/plugin/__tests__/plugin.spec.tsx
```diff
@@ -0,0 +1,234 @@
...
+        },
+      });
+      const { view } = createEditorWithSingleElementPresent([decoPlugin]);
+
+      const initialConsumerUpdateCount = consumerRenderSpy.mock.calls.length;
+      const initialFieldViewUpdateCount = fieldViewRenderSpy.mock.calls.length;
+
+      const tr = view.state.tr.setMeta(
+        "NO_OP",
+        "A change that does not affect content"
```
SHession, line 160: Should this read "A change that does ~not~ affect content"

SHession (commented):
I encountered an issue when testing this around elements becoming split and not serialising properly. This should be fixed by #155 and should be resolved by a rebase.

src/plugin/plugin.ts
```diff
@@ -116,26 +119,26 @@ const createNodeView = <
...
-    // node, to update the renderer. It's difficult to be typesafe here,
-    // as the Node's name value is loosely typed as `string`, and so we
-    // cannot index into the element `fieldDescription` to discover the appropriate type.
-    const fieldValues: Record<string, unknown> = {};
-    node.forEach((node, offset) => {
-      const fieldName = getFieldNameFromNode(
-        node
-      ) as keyof FieldNameToField<FDesc>;
-      const field = fields[fieldName];
-      field.view.onUpdate(node, offset, decos);
```
SHession, line 133: This `getValuesFromNode` also used to update the field views?

SHession (commented):
After the rebase, which has fixed the issue I was experiencing, I have found an issue whereby adding a decoration causes `updateFieldViewsFromNode` to be continuously called. I have discussed this with @jonathonherbert and this is an issue with the `testDecorationPlugin`, which is beyond the control of PME. We should seek to ensure this optimisation isn't undermined elsewhere in ProseMirror.

SHession (approved):

src/plugin/plugin.ts
jonathonherbert: IIUC, it's only used for the consumer (which doesn't understand the relationship between node and value) – the fieldView understands this relationship and so can derive the value itself if needed.

SHession: Sorry, bit unclear on your response there. Just to confirm my understanding, it looks like the function to update field views was previously executed as a side effect of the `getValuesFromNode` function. We have now split it out into a separate function. This a) makes it clearer that it's happening and b) allows us to conditionally call it.

In regards to your reply, I'm slightly confused by the term consumer (it's been a while since I've worked with PME).

jonathonherbert: Ah, gotcha – yes, it was a side effect (albeit in an enclosed function), so nice to eliminate!

Re: naming, agreed – I've created https://github.com/guardian/prosemirror-elements/issues/168 to track.

PR #167, opened 2022-01-10
feat: Type callback urls for automatic embed creation

jonathonherbert:
## What does this change?

Types the embed callback conversion urls with template string literals, per @davidfurey's suggestion.

## How to test

Typechecking with these types in the consuming context (see `demo/index.ts` below) should not cause a compile error.

## NB

I can't see a way of refining on the string such that I can achieve this without coercing with `as` – is this possible?

### DIFF
--- a/src/elements/embed/embedComponents/EmbedRecommendations.tsx
+++ b/src/elements/embed/embedComponents/EmbedRecommendations.tsx
@@ -3,9 +3,14 @@ import { buttonBrandAlt } from "@guardian/src-button";
 import { debounce } from "lodash";
 import { useCallback, useEffect, useState } from "react";
 import { Button } from "../../../editorial-source-components/Button";
-import type { TwitterUrl, YoutubeUrl } from "../EmbedSpec";
 import { conversionMessage, message } from "./embedStyles";
-import { getEmbedSource, parseHtml } from "./embedUtils";
+import type { TwitterUrl, YoutubeUrl } from "./embedUtils";
+import {
+  getEmbedSource,
+  isTwitterUrl,
+  isYoutubeUrl,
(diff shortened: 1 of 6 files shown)


src/elements/embed/embedComponents/EmbedRecommendations.tsx
```diff
@@ -35,7 +36,10 @@ export const EmbedRecommendation = ({
             embed code.{" "}
           </span>
           <ThemeProvider theme={buttonBrandAlt}>
-            <Button priority="secondary" onClick={() => convertTwitter(source)}>
+            <Button
+              priority="secondary"
+              onClick={() => convertTwitter(source as TwitterUrl)}
```
davidfurey, line 41: Maybe I'm being a bit OTT, but if you write this with `source.startsWith("https://twitter.com/")` as a user-defined type guard you'd avoid the need for the cast.

It would be really nice if Typescript could infer this.

jonathonherbert: As you say, a type guard probably the most elegant thing to do in the meantime – just done that.

davidfurey: I worked out how to change the definition of startsWith to be a type guard, only to find that someone has already done it and it has already been rejected

https://github.com/microsoft/TypeScript/pull/46959/files
https://github.com/microsoft/TypeScript/issues/46958

jonathonherbert: Well spotted – I'd never have thought to go back to the library typedef, good thinking. Can see the rationale.

davidfurey (approved):

PR #169, opened 2022-01-14
fix: Element validation delay bug by reversing ternary

SHession:
## What does this change?
This PR fixes a small bug which was causing sub-optimal performance and also a validation bug. The validation bug would manifest as a delay between changes being made to content and the validation updating. I.e. in a required field adding a single character wouldn't set the state to valid, you would then have to make an additional edit to the element.

## How to test

Make changes to elements which would cause it to be come valid or invalid, ensure that these validation warnings displayed when intended.

## Images

image


jonathonherbert (approved):
Great spot. I've added a unit test for this behaviour as well, a useful regression to defend against.

PR #170, opened 2022-01-14
fix: Reintroduce deserialisation to HTML for required or length checks

SHession:
## What does this change?

This PR Reverts guardian/prosemirror-elements#148. This is because we now perform our validation more [efficiently within Composer](https://github.com/guardian/flexible-content/pull/3929), this should allow us to include the original less performant deserialising HTML checks without compromising our application.

## How to test

Does the validation still work as expected? Does linking the package to Composer result in performance issues?

### DIFF
--- a/src/plugin/helpers/validation.ts
+++ b/src/plugin/helpers/validation.ts
@@ -66,8 +66,10 @@ export const htmlMaxLength = (
       },
     ];
   }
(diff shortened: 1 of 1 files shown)


SHession:
I've tested this locally using `yalc` to link the package. I experienced no issues working in a large Composer document with many images.

jonathonherbert (approved):
🎬

PR #174, opened 2022-01-17
feat: add target _blank behaviour to open in new tab

dskamiotis:
## What does this change?

At the moment the rich-link element does not open the depressed hyperlink in a new tab/window.
This simple PR adds the behaviour to open the selected hyperlink to replicate the current rich-link element in Composer
## How to test

Run locally or in code and add a rich-link - trying to click on the link to see the behaviour


jonathonherbert (approved):
Paired on testing 👍

PR #176, opened 2022-01-19
fix: Make photographer field absent on empty

jonathonherbert:
## What does this change?

Makes the photographer field absent when empty.

## How to test

TBC.

### DIFF
--- a/src/elements/image/imageElementDataTransformer.ts
+++ b/src/elements/image/imageElementDataTransformer.ts
@@ -1,3 +1,4 @@
+import pickBy from "lodash/pickBy";
 import type { FieldNameToValueMap } from "../../plugin/helpers/fieldView";
 import type { TransformIn, TransformOut } from "../helpers/types/Transform";
 import type { Asset, createImageFields, MainImageData } from "./ImageElement";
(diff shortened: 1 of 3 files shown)


src/elements/image/imageElementDataTransformer.ts
```diff
@@ -4,14 +4,14 @@ import type { Asset, createImageFields, MainImageData } from "./ImageElement";
...
+  caption?: string;
   displayCredit: string;
   imageType: string;
   isMandatory: string;
   mediaApiUri: string;
   mediaId: string;
-  photographer: string;
+  photographer?: string;
   role: string | undefined;
   source: string;
```
SHession, line 16: Source can also be optional.

SHession (approved):
Tested it locally and I can see that photographer is no longer included in the output if empty. However it might be nice to be consistent with the previous implementation of the element with other fields too (see my comment).

Element with empty text fields:

image

SHession (approved):
Re-tested with the latest changes by adding and removing content to the various fields. Also, paired on the solution. All seems to be working as expected! 
image

PR #177, opened 2022-01-20
fix: Fixes small crop, role setting bug

SHession:
## What does this change?
This PR fixes an issue with the "role" filed not being correctly set for images with small assets. Prior to this PR, the role setting would remain as the default until the element crop was changed. This meant if you added an image with small assets, the role would be incorrectly set but still display the weighting as "thumbnail". 

This PR fixes the issues by checking asset changes (including when a new image element is added) and sets the role to "thumbnail" if they are too small. This is done by using a `useEffect` in the `ImageView` component. 

## How to test

Add an image with dimensions less than 460 pixel to the sandbox environment. Use the ProseMirror dev tools to confirm the role is correct set to "thumbnail".


jonathonherbert (approved):

jonathonherbert:
Tested w/ @SHession, looks good to me!

PR #179, opened 2022-01-21
fix: Fix the thumbnail update issue

jonathonherbert:
_co-authored-by: @SHession_ 

## What does this change?

Our current implementation of the thumbnail constraint logic violates a lifecycle constraint under certain conditions in our rich text editor. This switches to another implementation that shouldn't be able to produce this bug.

## How to test

- Does the image addition/recrop behaviour behave as expected?
- Does the bug still exist in our CMS?

Tested w/ @SHession in Composer:

- initial image is too small
- recrop is too small
- regular image


SHession (approved):
Co-authored and tested locally using yalc, we will also test again in the equivalent Composer PR.

PR #180, opened 2022-01-24
docs: Update README to remove redundant guidance on yarn link

jonathonherbert:
## What does this change?

Update README to remove redundant guidance on yarn link – it's superceded by the docs on `yalc`.

## How to test

Are these docs redundant?


SHession (commented):
I'm not sure if we should be too prescriptive about developers workflows. Presumably the advice on `yarn link` could be useful for someone who would prefer to use it over `yalc`?

jonathonherbert:
Yeah, could be. Just feels odd to include this massive caveat!

We could instead retain it, and move underneath our recommended workflow?

SHession:
> We could instead retain it, and move underneath our recommended workflow?

I think this makes sense. We can recommend `yalc` first but it might helpful to mention why i.e. the issues with `yarn link`.

SHession (approved):

PR #183, opened 2022-01-24
feat: Add state management and UI for selected Elements

jonathonherbert:
## What does this change?

Adds a selected state to the element node attributes, and surfaces this state in our renderer to allow the UI to respond.

Before:

https://user-images.githubusercontent.com/7767575/150793398-c17e6c17-ec19-4366-a7ab-7cb22d9f6b07.mov

After (updated to reflect `user-select: none`):

https://user-images.githubusercontent.com/7767575/151000747-2ec0b7d8-b4e4-4e46-8b1c-29e54b0daa44.mov

### Why is that hardcoded highlight colour not in Source?

I've based it on the Chrome and Firefox selection colour for now 😅 we could pass it in on element creation if we really wanted to provide a way to customise, or perhaps there's another way to do it.

Incidentally, Chrome will automagically change the highlight colour when there's no contrast with the background colour, hence selection outlines showing in the labels and validation. We _could_ remove those with `user-select: none` for aesthetic reasons, but that means people can't copy/paste these labels/validation warnings, and I'm reluctant to restrict that unless there's a good reason – maybe there is!

## Developer notes.

There's a decision to make as to where to store the state that marks a node as selected:

 1. **Storing state in the nodes directly** gives us a single source of truth, but means we must check the document each time a selection changes and update nodes accordingly.
 2. **Storing state in the PME plugin** makes it easier to keep track of which nodes are selected in a way which doesn't require us to recurse through the document to check, but also gives us an opportunity to introduce bugs if that state became our of sync with our document.

I tried 1. first, and it turned out to be pretty quick, so I don't think there's a concern. Because we can short-circuit our evaluation of the document tree every time we encounter an element node, even very large documents don't present a problem. I profiled the `appendTransaction` function with a document containing 150 elements (far more than we'd be likely to experience in an actual document, there are almost a thousand fields being managed here). The iteration count is the number of times we're calling into callbacks passed into `descendants` or similar:

Screenshot 2022-01-24 at 08 33 12

## How to test

- The automated tests should pass, and they should be reasonably comprehensive. I've tried to cover a few edge cases WRT rerenders.
- Try this locally. Does it behave itself with content and elements of different kinds?

A clearer indication that an element is part of a user selection.

### DIFF
--- a/src/plugin/plugin.ts
+++ b/src/plugin/plugin.ts
@@ -71,24 +71,18 @@ export const createPlugin = <
         }
       );
 
+      // Update every relevant node with the new selection state.
       newState.doc.descendants((node, pos) => {
-        // Update nodes representing any elements that are no longer selected.
-        if (isProseMirrorElementSelected(node) && !elementNodeToPos.get(node)) {
+        const isCurrentlySelected = elementNodeToPos.get(node) !== undefined;
+        const shouldUpdateNode =
+          (isProseMirrorElementSelected(node) && !isCurrentlySelected) ||
+          (!isProseMirrorElementSelected(node) && isCurrentlySelected);
+
+        if (shouldUpdateNode) {
           tr.setNodeMarkup(pos, undefined, {
             ...node.attrs,
-            [elementSelectedNodeAttr]: false,
+            [elementSelectedNodeAttr]: isCurrentlySelected,
           });
-
-          return false;
-        }
-
-        // Update nodes representing any elements that are now selected.
-        if (elementNodeToPos.get(node) !== undefined) {
-          const newAttrs = {
-            ...node.attrs,
-            [elementSelectedNodeAttr]: true,
-          };
-          tr.setNodeMarkup(pos, undefined, newAttrs);
 
           return false;
         }
--- a/src/plugin/plugin.ts
+++ b/src/plugin/plugin.ts
@@ -63,7 +63,7 @@ export const createPlugin = <
           if (
             isProseMirrorElement(node) &&
             newState.selection.from <= pos &&
-            newState.selection.to > pos + node.nodeSize
+            newState.selection.to >= pos + node.nodeSize
           ) {
             elementNodeToPos.set(node, pos);
             return false;
@@ -211,7 +211,6 @@ const createNodeView = <
     dom,
     fields,
     (fields) => {
-      console.log("update", fields);
       view.dispatch(
         view.state.tr.setNodeMarkup(getPos(), undefined, {
           ...initNode.attrs,
@@ -225,18 +224,18 @@ const createNodeView = <
 
   return {
     dom,
-    update: (node, _, innerDecos) => {
+    update: (newNode, _, innerDecos) => {
       if (
-        node.type.name === nodeName &&
-        node.attrs.type === initNode.attrs.type
+        newNode.type.name === nodeName &&
+        newNode.attrs.type === initNode.attrs.type
       ) {
-        const newIsSelected = isProseMirrorElementSelected(node);
+        const newIsSelected = isProseMirrorElementSelected(newNode);
         const newCommands = commands(getPos, view);
         const newCommandValues = getCommandValues(newCommands);
 
         const isSelectedChanged = currentIsSelected !== newIsSelected;
-        const fieldValuesChanged = node !== currentNode;
         const innerDecosChanged = currentDecos !== innerDecos;
+        const fieldValuesChanged = fieldValuesHaveChanged(currentNode, newNode);
         const commandsChanged = commandsHaveChanged(
           currentCommandValues,
           newCommandValues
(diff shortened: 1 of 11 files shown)


src/plugin/plugin.ts
```diff
@@ -32,6 +42,62 @@ export const createPlugin = <
...
+            elementNodeToPos.set(node, pos);
+            return false;
+          }
+        }
+      );
+
+      // Update nodes representing any elements that are no longer selected.
+      newState.doc.descendants((node, pos) => {
+        if (isProseMirrorElementSelected(node) && !elementNodeToPos.get(node)) {
+          if (node.attrs[elementSelectedNodeAttr] === true) {
```
SHession, line 77: There is some redundancy here as `isProseMirrorElementSelected(node)` and `node.attrs[elementSelectedNodeAttr] === true` are equivalent.

src/plugin/plugin.ts
```diff
@@ -32,6 +42,62 @@ export const createPlugin = <
...
+            });
+          }
+          return false;
+        }
+      });
+
+      // Update nodes representing any elements that are now selected.
+      elementNodeToPos.forEach((pos, node) => {
+        if (node.attrs[elementSelectedNodeAttr] === true) {
+          return;
```
SHession, line 90: We could avoid this iteration through `elementNodeToPos` by updating the nodes selected state, true or false, during our `newState.doc.descendants` traversal.

SHession (commented):
Changes look good and tests are very thorough. I have left a couple of comments about the function updating the attribute state of nodes. 

I also noticed an issues with the down arrow command being set to disabled if there is white space returns underneath the element. This may be unconnected, but was noticeable when copy and pasting elements.

image

src/plugin/plugin.ts
jonathonherbert: Yes! That's really neat.

jonathonherbert:
Great feedback – just had a go at resolving both suggestions.

_> This may be unconnected, but was noticeable when copy and pasting elements._

Just tried it on `main` and can reproduce – may be to do with our predicate, which by default is ... `defaultPredicate`. There's a card here to take a look at positioning in general, I think, as IIRC we've also got some strange artefacts in Composer?

jonathonherbert:
(Hm, couple of failing tests, taking a look.)

src/plugin/plugin.ts
```diff
@@ -72,21 +72,25 @@ export const createPlugin = <
       );
 
       newState.doc.descendants((node, pos) => {
+        // Update nodes representing any elements that are no longer selected.
         if (isProseMirrorElementSelected(node) && !elementNodeToPos.get(node)) {
-          // Update nodes representing any elements that are no longer selected.
           tr.setNodeMarkup(pos, undefined, {
```
SHession, line 77: One alternative might be to do:
```Typescript
if(isProseMirrorElementSelected(node)){
  const newAttrs = {
      ...node.attrs,
      [elementSelectedNodeAttr]: elementNodeToPos.get(node) === undefined,
  };
  
  tr.setNodeMarkup(pos, undefined, newAttrs);
  
  return false;
}
```

You could then remove the second if block.

SHession, line 77: One alternative might be to do:
```Typescript
if(isProseMirrorElementSelected(node)){
  const newAttrs = {
      ...node.attrs,
      [elementSelectedNodeAttr]: elementNodeToPos.get(node) === undefined,
  };
  
  tr.setNodeMarkup(pos, undefined, newAttrs);
  
  return false;
}
```

You could then remove the second if block.

jonathonherbert:
Just given this a try – in the original code the second case depends on the early return of the first, so the tests fail w/ this suggestion. Agree that it'd be nice to remove the repetition, though – how about:

```ts 
      // Update every relevant node with the new selection state.
      newState.doc.descendants((node, pos) => {
        const isCurrentlySelected = elementNodeToPos.get(node) !== undefined;
        const shouldUpdateNode =
          isProseMirrorElementSelected(node) || isCurrentlySelected;

        if (shouldUpdateNode) {
          tr.setNodeMarkup(pos, undefined, {
            ...node.attrs,
            [elementSelectedNodeAttr]: isCurrentlySelected,
          });

          return false;
        }
      });
```

jonathonherbert:
Better, to avoid unnecessary updates:

```ts
      // Update every relevant node with the new selection state.
      newState.doc.descendants((node, pos) => {
        const isCurrentlySelected = elementNodeToPos.get(node) !== undefined;
        const shouldUpdateNode =
          (isProseMirrorElementSelected(node) && !isCurrentlySelected) ||
          (!isProseMirrorElementSelected(node) && isCurrentlySelected);

        if (shouldUpdateNode) {
          tr.setNodeMarkup(pos, undefined, {
            ...node.attrs,
            [elementSelectedNodeAttr]: isCurrentlySelected,
          });

          return false;
        }
      });
```

src/plugin/plugin.ts
SHession: Discussed with @jonathonherbert, this approach causes all PME elements to re-render. There are more efficient ways to only update what's necessary. See: https://github.com/guardian/prosemirror-elements/pull/183#issuecomment-1025905153

src/plugin/plugin.ts
```diff
@@ -32,6 +42,54 @@ export const createPlugin = <
...
+          (isProseMirrorElementSelected(node) && !isCurrentlySelected) ||
+          (!isProseMirrorElementSelected(node) && isCurrentlySelected);
+
+        if (shouldUpdateNode) {
+          tr.setNodeMarkup(pos, undefined, {
+            ...node.attrs,
+            [elementSelectedNodeAttr]: isCurrentlySelected,
+          });
+
+          return false;
```
SHession, line 87: A small possible hiccup, we are no longer descend nodes. This could potentially trip us up. We could wrap this descendants function in an `if(isProseMirrorElement)`, to resolve this.

SHession (approved):
Comments addressed. LGTM!

PR #184, opened 2022-01-25
feat: Add helper to programmatically update React components outside of a React context

jonathonherbert:
## What does this change?

Adds a helper function that allows us to programmatically update React components outside of their React context. This is useful when our CMS state informs our elements about what they should be renderering.

See the component itself and its tests for details on usage – hopefully, it's self-documenting, and any feedback to improve either API or docs would be great.

This is one approach to solving this problem, but others exist – definitely approach with an open mind! There are a few advantages to doing things in this way: 
- it doesn't extend the library API in solving what is essentially a UI problem
- it provides an imperative way of updating state (useful for consumers) that only rerenders the parts of the application that depend on that state, so it should be reasonably efficient

## How to test

- Does the new component make sense, conceptually and in use? Is there anything we could do to improve usability?
- Do the tests pass, and cover the functionality of the component?
- Does the new button, `Randomise image role options`, affect the image role options in the Image element as expected?


SHession:
It doesn't look like the checks have run on this PR, I'm not sure why.

src/elements/image/ImageElement.tsx
```diff
@@ -99,26 +103,28 @@ export const createImageFields = ({
...
+    role: createCustomDropdownField(undefinedDropdownValue, roleOptions),
   };
 };
 
-export const createImageElement = (props: MainImageProps) =>
-  createReactElementSpec(
-    createImageFields(props),
+export const createImageElement = (options: ImageElementOptions) => {
+  const { update: updateRoleOptions, Store: RoleStore } = createStore(
+    options.additionalRoleOptions
```
SHession, line 112: We might want to be consistent with out external facing API by calling `updateRoleOptions` -> `updateAdditionalRoleOptions`.

src/elements/image/ImageElementForm.tsx
```diff
@@ -53,34 +55,64 @@ const htmlLength = (text: string) => {
...
-            <CustomDropdownView
-              field={fields.role}
-              label="Weighting"
-              errors={errors.role}
-              options={
-                minAssetValidation(fieldValues.mainImage, "").length
+            <RoleOptionsStore>
+              {(additionalRoleOptions) => {
+                const roleOptions = minAssetValidation(
+                  fieldValues.mainImage,
```
SHession, line 77: Still on the fence about this being a separate component. It might look a bit tidier if it was split out, but I acknowledge the number of props it would require.

src/elements/image/ImageElementForm.tsx
```diff
@@ -53,34 +55,64 @@ const htmlLength = (text: string) => {
...
+                 */
+                useEffect(() => {
+                  const roleHasBeenRemoved = !roleOptions.some(
+                    ({ value }) => value === fieldValues.role
+                  );
+
+                  if (roleHasBeenRemoved) {
+                    fields.role.update(roleOptions[0].value);
+                  }
+                }, [roleOptions]);
```
SHession, line 97: Not an ideal solution, but should do the job!

src/plugin/types/Element.ts
```diff
@@ -32,7 +32,7 @@ export type FieldDescription =
   | TextFieldDescription
   | RichTextFieldDescription
   | CheckboxFieldDescription
-  | CustomFieldDescription
+  | CustomFieldDescription<unknown, unknown>
```
SHession, line 35: This line is producing a linter error, but the checks haven't run for this PR.

src/elements/image/imageElementDataTransformer.ts
```diff
@@ -52,15 +48,9 @@ export const transformElementIn: TransformIn<
...
 
-  let newRole = role;
-
-  if (minAssetValidation({ assets }, "").length) {
-    newRole = thumbnailOption.value;
-  }
-
   return {
     displayCredit: displayCredit === "true",
-    role: newRole ?? undefinedDropdownValue,
```
SHession, line 63: It's pleasing to be able to remove this, hopefully it still works as expected!

src/renderers/react/externalStore.ts
```diff
@@ -0,0 +1,70 @@
+import type { ReactElement } from "react";
+import { useEffect, useState } from "react";
+
+/**
+ * Creates a store to allow state within React to be programmatically updated
+ * outside of a React context.
+ */
+export const createStore = <Value>(getValue: Value) => {
```
SHession, line 8: I think this is a really eloquent solution. As discussed when pairing, there may be a more fitting name.

src/renderers/react/externalStore.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const store = {
+    /**
+     * Update the value in the store, informing subscribers.
+     *
+     * @return {boolean} true if there are any subscribers, false if not.
+     */
+    update: (val: Value): boolean => {
+      subscribers.forEach((sub) => sub(val));
+      return subscribers.length ? true : false;
```
SHession, line 21: Out of interest, what's the rationale / value behind this returning a boolean? Would a number of subscribers be as useful?

SHession (commented):
Changes look sensible and I have added some comments for your consideration. 
I have had some issues running this locally, with what looks like an infinite loop when adding a new image in the sandbox.

src/renderers/react/externalStore.ts
jonathonherbert: We could also do that! It's to ensure that we're able to test unsubscription without having to dig in to implementation detail of the store.

src/plugin/types/Element.ts
jonathonherbert: This is not needed any more – we can remove.

src/elements/image/ImageElementForm.tsx
jonathonherbert: Haha, yeah 😅

jonathonherbert:
@SHession yes, there's a pathological case as a result of `none-selected` never appearing in our options list! Easy to resolve.

jonathonherbert:
@SHession should now be resolved – I've tested with normal and thumbnail images on creation and crop. Interesting that we don't have tests for this (alongside our other elements, really!) – a blind spot we might want to remove.

demo/index.ts
```diff
@@ -66,12 +66,25 @@ const mockThirdPartyTracking = (html: string) =>
 
 const createCaptionPlugins = (schema: Schema) => exampleSetup({ schema });
 
+const additionalRoleOptions = [
+  { text: "inline (default)", value: "inline" },
```
SHession, line 70: Should the value of this item be `undefinedDropdownValue` to more closely reflect Composer?

src/elements/image/ImageElementForm.tsx
```diff
@@ -162,19 +170,61 @@ export const ImageElementForm: React.FunctionComponent<Props> = ({
...
+   * of options. If it doesn't, we update our field value to the
+   * first available option.
+   */
+  useEffect(() => {
+    if (!value || value === undefinedDropdownValue) {
+      return;
+    }
+
+    const roleHasBeenRemoved = !roleOptions.some(
+      ({ value }) => value === value
```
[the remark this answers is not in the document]
jonathonherbert: Whups! Good spot. Surprised this isn't a linting error. I'll raise on https://github.com/guardian/configs/tree/main/packages/eslint-config-typescript.

jonathonherbert:
@SHession that should now be fixed – worth noting that the first attempt at fixing the loop problem wasn't quite right.

Thinking that it was the logical result of a reference change in the `useEffect` dependencies, I memoised the options being passed to useEffect, but the problem persisted – creating a small image sent us into a loop where the image kept being destroyed and recreated.

It turns out that if we call `update` when the field is first created, it can be interfere with the Prosemirror lifecycle. Or at lest  that's my theory, which looks like:

 1. dispatch(1) starts: Image added
 2. Node created
 3. nodeView called
 4. React renders, and `useEffect` is triggered.
 5. Within `useEffect`, `field.update` is called, calling `updateOuterEditor` to update parent editor
 6. dispatch(2) starts in parent editor: but dispatch(1) not yet complete!

It's worth noting that I always thought `useEffect` was applied asynchronously, but that may not be the – React maintainers suggest that code run in effects [should be resilient to being run sync or async.](https://github.com/facebook/react/issues/22506#issuecomment-934542050).

So the solution is to defer the state updates for `CustomFieldView` by a tick or so to ensure they're run after Prosemirror is done with its state work. `CustomFieldView` is the only FieldView that exposes the opportunity to update state in this way, so it should be the only place we need to make `update()` async.

I've now tested by:

- creating a new image, with and without thumbnail
- updating image, with and without thumbnail
- randomising options and ensuring updates are correct

and all seems well.

src/plugin/fieldViews/CustomFieldView.ts
```diff
@@ -103,7 +103,14 @@ export class CustomFieldView<Value = unknown> implements FieldView<Value> {
...
   public update(value: Value) {
-    this.updateOuterEditor(value);
+    // We must wrap our update call in a setTimeout to avoid a bug. If update
+    // may be called synchronously as the field is initially rendered by a
+    // consumer, we will be calling `updateOuterEditor` in the middle of the
+    // ProseMirror dispatch lifecycle, which can produce odd results. Running
+    // our update asynchronously avoids this issue.
+    setTimeout(() => {
+      this.updateOuterEditor(value);
+    });
```
SHession, line 113: This makes sense as a solution given your thorough explanation. I have also tested this fix against a bug I found in Composer and found it to be fixed.

SHession (approved):
@jonathonherbert, thanks for the thorough explanation. It explains the issue we saw with the Thumbnail option, and the solution should protect us from other potential problems in the future. 

I have tested this locally in the usual ways, I have also tested it against my ongoing work in Composer. It all appears to be working as expected and responding to changes nicely. 

Approving this now, but we should hold off merging until we are happy with the Composer changes.

SHession (approved):
Paired on these changes with @jonathonherbert after testing in the related Composer PR. I've Issues with adding a new element to a document displaying all dropdown options due to the Store not maintaining any state have now been resolved. The exports of certain values have also tidied up sone gaps in Composer. LGTM!

PR #185, opened 2022-01-25
feat: Add video element

Fweddi:
Paired on with @jonathonherbert.

## What does this change?

Adds a video element:

Screenshot 2022-01-25 at 10 03 47

Handles empty caption fields and defaults role / weighting dropdown to null. This null dropdown logic is also used by the image and embed elements, so we have abstracted this out to a helper file. A similar abstraction is made for html length limits.

The video is rendered within an iframe that uses the classic CSS aspect ratio trick.

## How to test
Add this element directly in the demo page. It should behave as expected.

Successful replication of old video element.

We'll need to test thoroughly in our consuming CMS before shipping.

## Accessibility


SHession (approved):
I tested the element locally and compared it to one in Composer. Changes look good. One comment regarding the number of fields not rendered within the element.

PR #188, opened 2022-01-31
fix: Make alt, caption and url optional for embed, add isSafe field

rhystmills:
## What does this change?
This PR alters the embed's output, to more closely match the pre-existing setup. The intention is to reduce the chance of downstream bugs caused by a change in behaviour.

The following elements are now optional:
- `alt`
- `caption`
- `url`

There is now also an `isSafe` field. This checks if the html is for a single iframe, replicating a check that came from our [fence](https://github.com/guardian/fence/blob/9448f9a059d3264dfe0b3f3465cbacb498b3ffe9/fence.js#L139) package.

## How to test

Run the unit tests locally with `yarn:test-unit`

### DIFF
--- a/src/elements/helpers/html.ts
+++ b/src/elements/helpers/html.ts
@@ -14,7 +14,7 @@ export const parseHtml = (html: string) => {
   return parsedHtml.body.firstElementChild;
 };
 
-export const isHtmlSafe = (html: string) => {
+export const htmlContainsSingleIframe = (html: string) => {
   const holder = document.createElement("div");
   holder.innerHTML = (html || "").trim();
   const element = holder.firstElementChild;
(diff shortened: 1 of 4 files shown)


src/elements/helpers/html.ts
```diff
@@ -13,3 +13,12 @@ export const parseHtml = (html: string) => {
   );
   return parsedHtml.body.firstElementChild;
 };
+
+export const isHtmlSafe = (html: string) => {
```
jonathonherbert, line 17: One thing – might be nice to add a test for this, if only to clarify what it's for!

jonathonherbert: (And possibly rename to make it clearer what this does – `htmlContainsSingleIframe` ? 😅 )

rhystmills: Good suggestion, will do so!

PR #189, opened 2022-02-01
feat: Rename all tracking-related variables to use consistent naming

rhystmills:
## What does this change?
There was a mixture of variable names related to third party tracking in the `EmbedElement` and `InteractiveElement`. This PR updates all the relevant variables to refer to `ThirdPartyTracking` rather than `EmbedTracking`.

This will change the interface of the object expected by the `createEmbedElement` and `createVideoElement` functions to require a `thirdPartyTracking` property rather than an `embedTracking` property. As this is the main interface for that element with outside consumers (e.g. flexible-content) this could be considered a breaking change.

However, the Embed and Video elements are still in testing, so perhaps we can avoid the major version bump on that basis. Interested in the thought of reviewers on this.

## How to test
Run the application locally with `yarn start`. Run the tests with `yarn test:unit` and `yarn test:integration`.

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -103,7 +103,7 @@ const { plugin: elementPlugin, insertElement, nodeSpec } = buildElementPlugin({
   richlinkElement,
   videoElement: createVideoElement({
     createCaptionPlugins,
-    checkEmbedTracking: mockThirdPartyTracking,
+    checkThirdPartyTracking: mockThirdPartyTracking,
   }),
 });
 
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -89,7 +89,7 @@ const { plugin: elementPlugin, insertElement, nodeSpec } = buildElementPlugin({
   "demo-image-element": createDemoImageElement(onSelectImage, onDemoCropImage),
   imageElement,
   embedElement: createEmbedElement({
-    checkEmbedTracking: mockThirdPartyTracking,
+    checkThirdPartyTracking: mockThirdPartyTracking,
     convertTwitter: (src) => console.log(`Add Twitter embed with src: ${src}`),
     convertYouTube: (src) => console.log(`Add youtube embed with src: ${src}`),
     createCaptionPlugins,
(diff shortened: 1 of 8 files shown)


SHession (commented):
Changes make sense, I had an issue running it locally due to a small error in [Index.ts](https://github.com/guardian/prosemirror-elements/blob/b115e735065cce4eeb65511e17268afc54a0d7ba/demo/index.ts#L106).

In regards to bumping a major version, I understand your trepidation as it feels like a big deal to change it. The idea of semantic versioning is to apply a systematic process to version numbers, this reluctance is the intention of the process. Once a project reaches version 1.x, a change to public API requires a major release. There is a [useful FAQ on the subject](https://github.com/guardian/prosemirror-elements/blob/b115e735065cce4eeb65511e17268afc54a0d7ba/demo/index.ts#L106).

With this change specifically, we know that ProseMirror Elements and the GU Elements will eventually want to be split. I suspect this will make the versioning easier to reason with. I think in terms of semver we would still be treating GU Elements as 0.x.

To strictly adhere to the semver process, we should release a major version. However, I think if you want to leave it as a minor, that will be fine as the elements are still in testing.

jonathonherbert:
I'd say absolutely don't fear major releases once they're necessary – or we risk adhering to [another spec altogether 😁 ](http://sentimentalversioning.org/). This is a breaking change, so it's a major release. (I suspect there are other breaking changes we've introduced that probably warranted a major release between 1.0 and now. If that's true, it isn't by design.)

This software is used in production, so I think the usual versioning practices apply.

To force a major release, when squash-merging, I think we need to alter the body of the commit message to include `BREAKING CHANGE: `.  (This is the [Angular convention)[https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#footer] that's standard w/  sematic release. We should document this in our PR template.)

(Removing elements to gu-elements will be a breaking change when we get to it,  too, and that'll be a major release.)

jonathonherbert (approved):

PR #193, opened 2022-02-07
feat: Uses "crop" icon for image re-crop button instead of "camera" icon

SHession:
## What does this change?

This PR updates the image element re-crop icon from a camera to a conventional crop icon.

Before:

image

After:

image

## How to test

Checkout this branch, does the icon display correctly?

Users have a clear, more intuitive understanding of what the button does.


jonathonherbert (approved):
Tested locally, LGTM!

PR #194, opened 2022-02-07
feat: Include interactive in transform.ts

rhystmills:
## What does this change?
This PR includes the interactive element in `transform.ts`, which means it can properly interact with `flexible-content`.

It also:
- updates its fields to better reflect the data fields of the existing interactive element, which reduces the risk of problems downstream of `flexible-content` 
- updates the demo element to import the interactive more consistently with the other elements.
- Includes tests for the `interactiveDataTransformer`

## How to test

This should not affect the running of `prosemirror-elements`, only update its API - though you can run the application and test the demo locally following the instructions in the readme, or run the tests with `yarn test:unit` or `yarn test:integration`.

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -12,12 +12,12 @@ import {
   createDemoImageElement,
   createEmbedElement,
   createImageElement,
+  createInteractiveElement,
   pullquoteElement,
   richlinkElement,
 } from "../src";
 import { undefinedDropdownValue } from "../src/elements/helpers/transform";
 import type { MediaPayload } from "../src/elements/image/ImageElement";
-import { createInteractiveElement } from "../src/elements/interactive/InteractiveSpec";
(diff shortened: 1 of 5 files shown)


rhystmills:
@jonathonherbert has pointed out that the transforms don't remove optional fields, e.g. the alt text and caption, if they're empty strings. I will update and include those changes in this PR.

rhystmills:
> @jonathonherbert has pointed out that the transforms don't remove optional fields, e.g. the alt text and caption, if they're empty strings. I will update and include those changes in this PR.

Now included, as are tests for the `interactiveDataTransformer`.

jonathonherbert (approved):
Tests look good. Tested locally with the transforms (I'll make this a separate PR, as I think it makes it easier to observe changes), and actual behaviour matches expected for all three fields (alt, caption, role). 👍

PR #196, opened 2022-02-09
docs: Add data pane to show transformed element data

jonathonherbert:
## What does this change?

Adds a data pane, to make it easy to see what the data our elements  produce once they're run through our transformer.

data-pane

When elements aren't recognised, they produce `null`.

Along the way, updates our layout with a few ergonomic bits and pieces: the top buttons are now sticky, and each pane for content and data now scrolls independently.

## How to test

Have a play. Does the pane perform as expected? Is it useful?

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -21,7 +21,6 @@
     </div>
   </div>
 
-
   <div id="content-template" style="display: none;">
     <p>First paragraph</p>
     <p>Second paragraph</p>
--- a/demo/index.html
+++ b/demo/index.html
@@ -13,18 +13,20 @@
 </head>
 
 <body>
-  <div id="button-container"><button id="content-data-toggle">Toggle data view</button></div>
-  <div id="content">
-    <div id="editor-container"></div>
-    <div id="content-data"></div>
+  <div id="content-container">
+    <div id="button-container"><button id="content-data-toggle">Toggle data view</button></div>
+    <div id="content">
+      <div id="editor-container"></div>
+      <div id="content-data"></div>
+    </div>
   </div>
+
 
   <div id="content-template" style="display: none;">
     <p>First paragraph</p>
     <p>Second paragraph</p>
   </div>
 
-  <hr />
   <script src="bundle.js"></script>
 
   <div class="modal">
--- a/demo/index.html
+++ b/demo/index.html
@@ -13,12 +13,17 @@
 </head>
 
 <body>
-  <div id="button-container"></div>
-  <div id="editor-container"></div>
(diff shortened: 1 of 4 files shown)


Fweddi:
> When elements aren't recognised, they produce null.

Could you expand on this? I notice a few of the demoable elements are returning null at the moment. Does this mean we haven't added the necessary data transformers for them yet?

jonathonherbert:
@Fweddi the demo-image-element and the interactive element should be the only ones that don't have transformers – the latter should now be solved by a recent rebase, as the transformer was added in #194. (The demo element doesn't need one, as it's not used in Composer.)

jonathonherbert:
Test comment – please ignore.

demo/index.ts
```diff
@@ -85,23 +94,28 @@ const {
...
     checkThirdPartyTracking: mockThirdPartyTracking,
     createCaptionPlugins,
   }),
-  codeElement,
-  pullquoteElement,
-  richlinkElement,
-  videoElement: createVideoElement({
+  code: codeElement,
+  pullquote: pullquoteElement,
+  "rich-link": richlinkElement,
```
rhystmills, line 117: Interested in why you've gone for kebab case when it creates this necessity?

demo/index.html
```diff
@@ -13,13 +13,20 @@
...
-  <div id="editor-container"></div>
+  <div id="content-container">
+    <div id="button-container"><button id="content-data-toggle">Toggle data view</button></div>
+    <div id="content">
+      <div id="editor-container"></div>
+      <div id="content-data"></div>
+    </div>
+  </div>
+
+
```
rhystmills, line 24: Very minor nit but the double space is probably unnecessary

rhystmills (approved):
Really useful feature as far as I'm concerned, have tested locally and it's working well.

I've added a couple of minor comments but they're not blockers. 

I'd be tempted to have the feature on by default; it might also be nice to have a different button-style when it's toggled 'on' in that case.

demo/index.ts
jonathonherbert: Good question! It's because this matches our requirements in Composer. Before #156, this small change broke our plugin, so it's nice to have it in here for testing purposes. Could add a comment to that effect!

jonathonherbert: (There are also automated tests for this)

jonathonherbert:
@rhystmills agreed on toggling button, if not default (I often work in half windows) – I'm going to leave for now pending a React rewrite of our demo, which'll make the state less cumbersome, but we could persist the preference in localstorage if we fancied!

PR #199, opened 2022-02-15
feat: Add the Callout

rhystmills:
Co-authored with @SHession 

## What does this change?

This PR adds the Callout. This is rendered when the `html` field of an Embed element contains the string "data-callout-tagname".

I've taken the opportunity to make the Callout's UX more useful.

In Composer, the user doesn't get much feedback from the old Callout: 
image
This makes it hard for a user to understand what the element actually represents, or whether it's working as intended.

The new Callout fetches information from [Targeting](https://targeting.gutools.co.uk/) - the same information that is used by platforms to render the Callout in articles.
image

There are also two error states representing:
- a failure to find a callout matching the tag
- a failure to parse a tag

(Previously these would silently fail)
image

## Context: 

We recently turned on the Embed element's feature switch. Shortly after, @SHession noticed that Callouts were being displayed as Embed elements in [Composer](https://github.com/guardian/flexible-content).

'Under the hood' - Callouts **are** embed elements - specifically any Embed element that contains the string "data-callout-tagname". Instead of being rendered as a form (like other Embed elements) - the Callout is rendered as an uneditable DOM element. In Composer, they're normally added with the 'Create Callout' button, but an Embed will also be converted into a Callout if the `html` field contains the string "data-callout-tagname".

It's worth mentioning that the Callout element could previously add an optional 'title' field to its representation in flexible-api. This is not used downstream so we are not continuing to add it.

## How to test

Hit the 'add callout element' button.

Alternatively, paste callout html into the embed element's `Embed code` field.

Valid callout:
```
<div data-callout-tagname="callout-brexit"><h2>Callout<h2><p>callout-brexit</p></div>
```
Invalid callout:
```
<div data-callout-tagname="tag-name-here"><h2>Callout<h2><p>tag-name-here</p></div>
```
Tag read error:
```
data-callout-tagname
```

## Accessibility

- [X] [Navigable with keyboard](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#keyboard-navigation)
- [X] [Colour contrast passed](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#colour-contrast)
- [X] [The change doesn't use only colour to convey meaning](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#use-of-colour)
- [X] [Interactive elements show a focus ring when focused by keyboard](https://github.com/guardian/accessibility/blob/main/editorial-tools-guidelines.md#appendix-the-focus-ring)
- [X] [Semantic elements have been used where possible](https://github.com/guardian/accessibility/blob/main/editorial-tools-guidelines.md#appendix-semantic-html)

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -35,6 +35,7 @@ import type { WindowType } from "./types";
 // they click a text field.
 FocusStyleManager.onlyShowFocusOnTabs();
 const embedElementName = "embedElement";
+const calloutElementName = "embedElement";
 const imageElementName = "imageElement";
 const demoImageElementName = "demo-image-element";
 const codeElementName = "codeElement";
@@ -45,6 +46,7 @@ const interactiveElementName = "interactiveElement";
 
 type Name =
   | typeof embedElementName
+  | typeof calloutElementName
   | typeof imageElementName
   | typeof demoImageElementName
   | typeof codeElementName
@@ -89,6 +91,12 @@ const { plugin: elementPlugin, insertElement, nodeSpec } = buildElementPlugin({
   "demo-image-element": createDemoImageElement(onSelectImage, onDemoCropImage),
   imageElement,
   embedElement: createEmbedElement({
+    checkThirdPartyTracking: mockThirdPartyTracking,
+    convertTwitter: (src) => console.log(`Add Twitter embed with src: ${src}`),
+    convertYouTube: (src) => console.log(`Add youtube embed with src: ${src}`),
+    createCaptionPlugins,
+  }),
+  calloutElement: createEmbedElement({
     checkThirdPartyTracking: mockThirdPartyTracking,
     convertTwitter: (src) => console.log(`Add Twitter embed with src: ${src}`),
     convertYouTube: (src) => console.log(`Add youtube embed with src: ${src}`),
@@ -219,6 +227,17 @@ const createEditor = (server: CollabServer) => {
     caption: "",
     altText: "",
     required: false,
+  });
+
+  createElementButton("Add callout element", calloutElementName, {
+    weighting: "",
+    sourceUrl: "",
+    embedCode: "",
+    caption: "",
+    altText: "",
+    required: false,
+    html:
+      '<div data-callout-tagname="callout-brexit"><h2>Callout<h2><p>callout-brexit</p></div>',
   });
 
   createElementButton("Add demo image element", demoImageElementName, {
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -35,7 +35,6 @@ import type { WindowType } from "./types";
 // they click a text field.
 FocusStyleManager.onlyShowFocusOnTabs();
 const embedElementName = "embedElement";
-const calloutElementName = "embedElement";
 const imageElementName = "imageElement";
 const demoImageElementName = "demo-image-element";
 const codeElementName = "codeElement";
@@ -46,7 +45,6 @@ const interactiveElementName = "interactiveElement";
 
 type Name =
   | typeof embedElementName
-  | typeof calloutElementName
   | typeof imageElementName
   | typeof demoImageElementName
   | typeof codeElementName
@@ -91,12 +89,6 @@ const { plugin: elementPlugin, insertElement, nodeSpec } = buildElementPlugin({
   "demo-image-element": createDemoImageElement(onSelectImage, onDemoCropImage),
   imageElement,
   embedElement: createEmbedElement({
-    checkThirdPartyTracking: mockThirdPartyTracking,
-    convertTwitter: (src) => console.log(`Add Twitter embed with src: ${src}`),
-    convertYouTube: (src) => console.log(`Add youtube embed with src: ${src}`),
-    createCaptionPlugins,
-  }),
-  calloutElement: createEmbedElement({
     checkThirdPartyTracking: mockThirdPartyTracking,
     convertTwitter: (src) => console.log(`Add Twitter embed with src: ${src}`),
     convertYouTube: (src) => console.log(`Add youtube embed with src: ${src}`),
@@ -229,13 +221,9 @@ const createEditor = (server: CollabServer) => {
     required: false,
   });
 
-  createElementButton("Add callout element", calloutElementName, {
-    weighting: "",
-    sourceUrl: "",
-    embedCode: "",
-    caption: "",
+  createElementButton("Add callout element", embedElementName, {
     altText: "",
-    required: false,
+    caption: "",
     html:
       '<div data-callout-tagname="callout-brexit"><h2>Callout<h2><p>callout-brexit</p></div>',
   });
--- a/src/elements/embed/Callout.tsx
+++ b/src/elements/embed/Callout.tsx
@@ -54,7 +54,7 @@ const calloutStyles = css`
     vertical-align: top;
   }
   th {
-    text-align: right;
+    text-align: left;
   }
   p,
   p:first-child {
--- a/src/elements/embed/Callout.tsx
+++ b/src/elements/embed/Callout.tsx
@@ -184,7 +184,8 @@ export const Callout: React.FunctionComponent<Props> = ({ tag }) => {
         }
       })
       .catch((e) => console.log(e));
-  }, []);
+    return () => setCallout(undefined);
+  }, [tag]);
 
   return (
     <FieldLayoutVertical data-cy={EmbedTestId}>
--- a/src/elements/embed/Callout.tsx
+++ b/src/elements/embed/Callout.tsx
@@ -7,6 +7,7 @@ import { Label } from "../../editorial-source-components/Label";
 import { FieldLayoutVertical } from "../../editorial-source-components/VerticalFieldLayout";
 import type { FieldNameToValueMap } from "../../plugin/helpers/fieldView";
 import { unescapeHtml } from "../helpers/html";
+import { EmbedTestId } from "./EmbedForm";
 import type { createEmbedFields } from "./EmbedSpec";
 
 type Props = {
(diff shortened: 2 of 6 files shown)


src/elements/embed/EmbedSpec.tsx
```diff
@@ -62,15 +67,18 @@ export const createEmbedElement = (props: MainEmbedProps) =>
...
+        return (
+          <EmbedForm
+            fields={fields}
+            errors={errors}
+            fieldValues={fieldValues}
+            checkThirdPartyTracking={props.checkThirdPartyTracking}
+            convertYouTube={props.convertYouTube}
+            convertTwitter={props.convertTwitter}
+          />
+        );
```
SHession, line 82: I think this looks slightly nicer as ternary but makes no real difference:

```Typescript
({ fields, errors, fieldValues }) =>
      isCallout(fieldValues.html) ? (
        <Callout fieldValues={fieldValues} />
      ) : (
        <EmbedForm
          fields={fields}
          errors={errors}
          fieldValues={fieldValues}
          checkThirdPartyTracking={props.checkThirdPartyTracking}
          convertYouTube={props.convertYouTube}
          convertTwitter={props.convertTwitter}
        />
      )
```

demo/index.ts
```diff
@@ -221,6 +229,17 @@ const createEditor = (server: CollabServer) => {
...
 
+  createElementButton("Add callout element", calloutElementName, {
+    weighting: "",
+    sourceUrl: "",
+    embedCode: "",
+    caption: "",
+    altText: "",
+    required: false,
+    html:
+      '<div data-callout-tagname="callout-brexit"><h2>Callout<h2><p>callout-brexit</p></div>',
```
SHession, line 240: We should be able to only include the fields we know Composer will provide, giving us a more accurate reflection of production.

src/elements/embed/EmbedSpec.tsx
```diff
@@ -21,6 +22,10 @@ export type MainEmbedProps = {
   createCaptionPlugins?: (schema: Schema) => Plugin[];
 };
 
+const isCallout = (html: string) => {
+  return html.includes("data-callout-tagname");
```
SHession, line 26: It's a shame we don't have anything else on the element data we can use to determine this!

src/elements/embed/Callout.tsx
```diff
@@ -0,0 +1,217 @@
...
+import { Error } from "../../editorial-source-components/Error";
+import { Label } from "../../editorial-source-components/Label";
+import { FieldLayoutVertical } from "../../editorial-source-components/VerticalFieldLayout";
+import type { FieldNameToValueMap } from "../../plugin/helpers/fieldView";
+import { unescapeHtml } from "../helpers/html";
+import { EmbedTestId } from "./EmbedForm";
+import type { createEmbedFields } from "./EmbedSpec";
+
+type Props = {
+  fieldValues: FieldNameToValueMap<ReturnType<typeof createEmbedFields>>;
```
SHession, line 14: `nit`, we could simply pass it the HTML value as that's the only part of the `fieldValues` we need.

src/elements/embed/Callout.tsx
```diff
@@ -0,0 +1,217 @@
...
+    .then((data: Callout[]) => {
+      return data.find((callout) => callout.fields.tagName === tag);
+    });
+};
+
+export const extractTag = (html: string) => {
+  const pattern = 'data-callout-tagname="(.*?)"';
+  const tag = RegExp(pattern).exec(html);
+  return tag ? tag[1] : undefined;
+};
```
SHession, line 87: If we wanted to avoid RegEx, we could also extract the attribute value by parsing creating an HTML element and setting the `innerHtml`. You could then retrieve the value using the [dataset api](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset).

src/elements/embed/Callout.tsx
```diff
@@ -0,0 +1,217 @@
...
+    margin-bottom: 0px;
+  }
+`;
+
+const marginBottom = css`
+  margin-bottom: ${space[2]}px !important;
+`;
+
+const getCampaigns = (tag: string) => {
+  return fetch("https://targeting.gutools.co.uk/api/campaigns")
```
SHession, line 74: Is there a more specific endpoint we can use for this?

SHession (commented):
This looks great, I think the new way of displaying callouts is a bit improvement! We may just want to confirm with product that they're happy with the change, I understand we have already spoken to UX. 

I have left a few comments for you to consider. I was also wondering if the loss of the tracking text and tick was significant, was there any chance we would want to keep this?

demo/index.ts
```diff
@@ -94,6 +96,12 @@ const { plugin: elementPlugin, insertElement, nodeSpec } = buildElementPlugin({
     convertYouTube: (src) => console.log(`Add youtube embed with src: ${src}`),
     createCaptionPlugins,
   }),
+  calloutElement: createEmbedElement({
+    checkThirdPartyTracking: mockThirdPartyTracking,
+    convertTwitter: (src) => console.log(`Add Twitter embed with src: ${src}`),
+    convertYouTube: (src) => console.log(`Add youtube embed with src: ${src}`),
+    createCaptionPlugins,
+  }),
```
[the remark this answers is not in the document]
rhystmills: Good point - unnecessary in hindsight. I've removed it and some other additions to `demo/index.ts`

src/elements/embed/EmbedSpec.tsx
rhystmills: It is a bit rubbish, though consistent with how `flexible-content` determines callouts in the old element.

A minor improvement might be checking that the html is valid, and the parent node contains a property called "data-callout-name" which has an actual value. For now I think I'd go with consistency with the old element though.

src/elements/embed/EmbedSpec.tsx
rhystmills: I think so too - have done so now.

src/elements/embed/EmbedSpec.tsx
SHession: That might be a good way to do it, we could memoise the function and extract the `data-callout-name` value, it also relates to my other comment about the current tag extraction.

demo/index.ts
rhystmills: Good point, have done so now.

src/elements/embed/Callout.tsx
rhystmills: I've done so now!

src/elements/embed/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+      </Error>
+      <p>Try refreshing the page. If the problem persists, you may wish to:</p>
+      <ul>
+        <li>
+          Check that the callout exists in{" "}
+          <a href="https://targeting.gutools.co.uk/campaigns/">Targeting</a>
+        </li>
+        <li>
+          Contact Central Production (
+          <a href="mailto:central.production@guardian.co.uk">
```
paperboyo, line 146:
```suggestion
          <a href="mailto:central.production@theguardian.com">
```

src/elements/embed/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+            central.production@guardian.co.uk
```
paperboyo, line 147:
```suggestion
            central.production@theguardian.com
```

src/elements/embed/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+          </a>
+          ) and the Editorial Tools team (
+          <a href="mailto:editorial.tools.dev@guardian.co.uk">
+            editorial.tools.dev@guardian.co.uk
```
paperboyo, line 151:
```suggestion
          <a href="mailto:editorial.tools.dev@theguardian.com">
            editorial.tools.dev@theguardian.com
```

src/elements/embed/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+        Error: Composer was unable to extract a tag from this callout.
+      </Error>
+      <p css={marginBottom}>
+        It is likely that the callout element is malformed. Try deleting and
+        re-creating the the element.
+      </p>
+      <p>
+        If the problem persists, you may wish to contact Central Production (
+        <a href="mailto:central.production@guardian.co.uk">
+          central.production@guardian.co.uk
```
paperboyo, line 169:
```suggestion
        <a href="mailto:central.production@theguardian.com">
          central.production@theguardian.com
```

src/elements/embed/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+        </a>
+        ) and the Editorial Tools team (
+        <a href="mailto:editorial.tools.dev@guardian.co.uk">
+          editorial.tools.dev@guardian.co.uk
```
paperboyo, line 173:
```suggestion
        <a href="mailto:editorial.tools.dev@theguardian.com">
          editorial.tools.dev@theguardian.com
```

paperboyo:
Super-nice! The nittest nits of all from moi.

src/elements/embed/EmbedSpec.tsx
rhystmills: I'm now getting the tag up front in `EmbedSpec` and passing only the tag into the Callout, since that's all it really needs. `extractTag` is gone - the functionality is part of `getCalloutTag` (replacing `isCallout`)

src/elements/embed/Callout.tsx
rhystmills: (mentioned above that I'm doing this in `EmbedSpec` now - though I'm using `getAttribute` to avoid the hassle of demonstrating to Typescript that it's an HTMLElement with a `dataset` property, rather than an Element that doesn't have one.

src/elements/embed/Callout.tsx
rhystmills: All resolved now, cheers!

src/elements/embed/EmbedSpec.tsx
SHession: This looks great, I would recommend we [memoise it](https://reactjs.org/docs/hooks-reference.html#usememo) as a performance enhancement.

rhystmills: (Sam and I paired on this and decided only to memoise the network request-producing call to Targeting)

src/elements/embed/Callout.tsx
rhystmills: (We did memoise this one in the end while pairing)

SHession (approved):
Changes look good and comments have been addressed. The element now no longer re-fetches the campaign items. I have tested locally and the UI looks good and behaves well. LGTM!

PR #200, opened 2022-02-15
feat: Resize interactive iframe based on content

rhystmills:
## What does this change?
This PR causes the interactive element's preview to resize based on its content, provided that the interactive observes the `Window.postMessage` interface specified by [iframe-messenger](https://github.com/guardian/iframe-messenger).

Resizing the preview is more complicated for interactives than the existing embed, because interactives are on another domain, therefore CORS prevents us from directly querying the height of the iframe's window. Therefore, we can interact with the iframe's document only through the messaging API.

Along the way, we noticed that React elements representing our Prosemirror Elements were not being dismounted when we deleted or moved them. This meant that cleanup functions in the `useEffect` hook were never being called, and we were unable to cleanup, for example, event listeners that were registered in a Component. As a result, this PR also includes a fix to properly unmount elements.

## How to test

Run `prosemirror-elements` locally and add the interactive element. Does the height of the interactive element resize to fit the entire interactive? This may be easier to observe if you shrink the window, as the elements will grow to fullscreen width.

## Images

| Before | After |
| --- | --- |
|   |  |

## Accessibility


src/elements/helpers/Preview.tsx
```diff
@@ -24,6 +47,28 @@ const getDocHeight = (doc: Document | undefined) => {
   return undefined;
 };
 
+function isUntypedMessageData(data: unknown): data is UntypedMessageData {
```
SHession, line 50: We should aim to use arrow functions consistently within this project.

src/elements/helpers/Preview.tsx
```diff
@@ -24,6 +47,28 @@ const getDocHeight = (doc: Document | undefined) => {
...
+  if (
+    typeof data === "object" &&
+    data !== null &&
+    "value" in data &&
+    "type" in data
+  ) {
+    return true;
+  }
+  return false;
+}
```
SHession, line 60: This seems like a slightly unnecessary step. It might be better to incorporate this logic into `isMessageData` and remove the `UntypedMessageData` type.

src/elements/helpers/Preview.tsx
```diff
@@ -40,23 +85,45 @@ export const Preview = ({
     }
   };
 
+  const listenToMessages = () => {
+    window.addEventListener("message", (message) => {
```
SHession, line 89: We need to be a bit careful here, we are adding event listeners that never being cleaned up.

src/elements/helpers/Preview.tsx
```diff
@@ -40,23 +85,45 @@ export const Preview = ({
...
+      try {
+        const data = JSON.parse(message.data) as unknown;
+
+        if (isMessageData(data) && data.type === "set-height") {
+          const height = data.value;
+          setHeight(`${(height + 4).toString()}px`);
```
SHession, line 95: `nit` It might be nice to move this `4` into a const to be explicit about what it represents.

SHession (commented):
Changes make sense and work locally. I've added a couple of comments for your consideration.

src/elements/helpers/Preview.tsx
rhystmills: Strongly agree, not sure where that came from. Retro.

src/elements/helpers/Preview.tsx
rhystmills: This was the result of some frustrating type shenanigans with Typescript & Eslint TS was not happy that the properties existed unless I used a type predicate .I've had another go this morning, with a Typescript friendly generalised property-finding type predicate helper (`hasOwnProperty`) - and have been able to remove `isUntypedMessageData`.

src/elements/helpers/Preview.tsx
```diff
@@ -24,6 +43,19 @@ const getDocHeight = (doc: Document | undefined) => {
   return undefined;
 };
 
+const isMessageData = (data: unknown): data is MessageData => {
+  if (
```
jonathonherbert, line 47: IIUC could just return this set of comparisons directly w/o an if statement, but you might prefer this for readability – not a strongly held opinion.

src/elements/helpers/ThirdPartyStatusChecks.tsx
```diff
@@ -86,27 +86,33 @@ export const TrackingStatusChecks = ({
...
+        if (options.isSubscribed) {
+          updateTrackingStatus(response);
+        }
       })
       .catch((e) => console.log(e));
   };
 
   const checkTrackingAndUpdateDebounced = useCallback(
-    debounce(checkTrackingAndUpdate, 3000),
+    debounce(checkTrackingAndUpdate, 5000),
```
SHession, line 103: We should set this down to something quicker maybe 500 - 1000ms?

src/plugin/fieldViews/CustomFieldView.ts
```diff
@@ -114,7 +114,7 @@ export class CustomFieldView<Value = unknown> implements FieldView<Value> {
   }
 
   public destroy() {
-    this.subscribers = [];
+    // Do nothing
```
SHession, line 117: It would be nice to avoid this, but it's a blocker.

SHession (commented):
Comments addressed and changes look great. Unfortunately I found a bug testing this with two different interactive elements. The event listener doesn't discriminate whether the message is for its own component or not, meaning the height is set the same for every element.

src/elements/helpers/Preview.tsx
rhystmills: That would be neater - have implemented now.

SHession (approved):
Solution makes sense, changes LGTM!

PR #203, opened 2022-02-21
feat: ds/rich-link warning unpublished article

dskamiotis:
## What does this change?

Currently the behaviour for the rich link identifies whether the article which is being linked has been published or is in draft/preview mode.

However the new rich link element does not, this PR aims to amend this by adding:
- `draftReference` as in the rich link spec to communicate the status to Composer
- Styling to the element which renders a warning message to the user
- New style for warning designed by Ana Pradas - https://www.figma.com/file/Ji1XENAJJjzcxwwwfWZnJc/Embeds?node-id=606%3A5771

## How to test

1 - Create an article in Composer up to a publishable state - but do not publish
2 - Create/open another Composer article
3 - Add the (non published) URL as a rich link in the new article and at this point the warning message will render
4 - Add a URL of an existing published article to see that no warning message shows

### Before
image

### Original element
image

### New Element
image

## Accessibility

### DIFF
--- a/src/elements/rich-link/RichlinkForm.tsx
+++ b/src/elements/rich-link/RichlinkForm.tsx
@@ -63,7 +63,7 @@ export const RichlinkElementForm: React.FunctionComponent<Props> = ({
         <span css={warningStyle}>
           <SvgAlertTriangle />
           This rich link references unpublished content. It will not appear
-          until the tarket has been published.
+          until the target has been published.
         </span>
       ) : null}
     </div>
--- a/src/elements/rich-link/RichlinkForm.tsx
+++ b/src/elements/rich-link/RichlinkForm.tsx
@@ -1,3 +1,4 @@
+import { css } from "@emotion/react";
 import React from "react";
 import { FieldLayoutVertical } from "../../editorial-source-components/VerticalFieldLayout";
 import type { FieldValidationErrors } from "../../plugin/elementSpec";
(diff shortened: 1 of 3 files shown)


src/elements/rich-link/RichlinkForm.tsx
```diff
@@ -30,5 +58,14 @@ export const RichlinkElementForm: React.FunctionComponent<Props> = ({
       errors={errors.weighting}
       display="inline"
     />
+    <div css={warningContainer}>
```
jonathonherbert, line 61: Could we put the conditional one level up, so there's no warning markup at all unless we have an alert?

jonathonherbert (commented):
Tested locally by adding content to the `draftReference` field manually – can't see an easy way to test in Composer, but the logic looks fine 👍 One blocking comment, once addressed, let's 🚢 !

dskamiotis:
Tested in Composer LOCAL successfully, not sure about a robust way in CODE also.

src/elements/rich-link/RichlinkForm.tsx
dskamiotis: > put the conditional one level up, so there's no warning markup

Really good spot @jonathonherbert the mark up was definitely there even when no alert needed. I have moved the conditional to contain the parent <div>

jonathonherbert (approved):

PR #204, opened 2022-02-23
feat: Minor amends to rich text element

jonathonherbert:
## What does this change?

 - Removes the warning container to ensure warning text wraps
 - Ensures that draftReference is absent in output data when it is an empty string, for consistency with PROD

## How to test

- Add a draftReference to the rich link creation function in `demo/index.ts`, and check that the text wraps
- It's actually quite difficult to test the draftReference w/o using it in consuming code! You'll need to use `yalc` to confirm. Alternatively, merge with #196 locally and observe the output data.

Consistency w/ our current setup.

### DIFF
--- a/src/elements/rich-link/RichlinkForm.tsx
+++ b/src/elements/rich-link/RichlinkForm.tsx
@@ -32,14 +32,6 @@ const warningStyle = css`
   }
 `;
 
-const warningContainer = css`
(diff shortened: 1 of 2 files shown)


SHession (approved):
It was a bit tricky to test locally as noted but I managed it in the end. LGTM.

PR #207, opened 2022-02-25
feat: Improve interactive preview

jonathonherbert:
## What does this change?

Two minor improvements to interactive preview as a result of an investigation with @rhystmills:

- Restrict the iframe from filling more than 50% of the screen by height, to prevent long interactives from making documents difficult to scroll through. (Number arbitrary and up for discussion – I did what felt about right.)
- Consistently escape HTML in the iframe URL to prevent query parameters from breaking preview.

Before, the interactive below would not display a preview. Now, it displays, and its height is limited:

https://user-images.githubusercontent.com/7767575/155760524-120f58c2-cd56-4eb3-9f61-feabec5546c3.mov

## How to test

Fire up prosemirror-elements locally, and add an interactive element to the document – this PR changes it to an element that demonstrates both the improvements above. Does it behave as expected?

Previews work more consistently, and do not occupy too much document real estate.

There's a possibility that users will think that some of the interactive is missing when the interactive height is restricted. In mitigation: users can expand the element containing the preview to show more of it; they can also preview the document.


rhystmills (approved):
Running as expected locally and looks to have solved the issue with an unescaped iframe URL 👍

PR #209, opened 2022-02-28
chore: Introduce esbuild for project build

jonathonherbert:
## What does this change?

Adds esbuild to the repo, speeding up build startup and watch times considerably.

To ensure typechecking continues, we use `tork-ts-checker-webpack-plugin`, which runs in parallel with the main build and does not block module output in watch mode.

Unfortunately, esbuild doesn't support the [new jsx runtime yet](https://github.com/evanw/esbuild/issues/334), which means we need to explicitly import `React` when using jsx. (We were inconsistent before.) I've added a linting rule via the `react/recommended` plugin to ensure this is the case, and fixed a couple of other things that this new linting config spotted, too.

## How to test

Run `yarn watch` and `yarn build`. (Don't forget to run `yarn` in the project root beforehand, to install the new dependencies!) The build should work as before, and the tests should pass.

Both operations should be noticeably faster.

Faster watch and build operations.


.eslintrc.cjs
```diff
@@ -2,14 +2,21 @@ module.exports = {
...
   overrides: [
     {
       files: ["*.ts", "*.tsx"],
       extends: ["@guardian/eslint-config-typescript"],
       rules: {
         "@typescript-eslint/unbound-method": 0,
         "@typescript-eslint/explicit-module-boundary-types": 0,
-        "@typescript-eslint/no-unused-vars": 2
+        "@typescript-eslint/no-unused-vars": 2,
+        "react/display-name": 0
```
jonathonherbert, line 19: This is nice to have, but makes it difficult to define private class methods that represent elements, so I've turned it off for now.

twrichards:
in watch mode, does this run _tsc watch_ in parallel with the rebuilds, if not you might want to do that part manually?

jonathonherbert:
@twrichards IIUC that's the job of `fork-ts-checker-webpack-plugin`:

_> runs in parallel with the main build and does not block module output in watch mode._

jonathonherbert:
(Incidentally AFAICS pinboard is not using esbuild for the client, but if you were and found a neat solution to `jsxImportSource`, I'd be interested to hear – at the moment, emotion components do not render correctly, I presume because they are not receiving their jsx function from `emotion`)

jonathonherbert:
Closing in favour of an eventual PR to use `vite`, which has worked well [elsewhere.](https://github.com/guardian/prosemirror-typerighter/pull/208)

PR #211, opened 2022-03-01
chore: Put integration tests in separate workflow

jonathonherbert:
## What does this change?

Puts integration tests in a separate workflow, to speed up build time.

## How to test

Do the actions complete successfully? Does the overall time to build decrease?

|Before|After (with warm `yarn` cache, running in parallel)|
|--|--|
|Screenshot 2022-03-01 at 18 01 50|Screenshot 2022-03-01 at 18 05 21Screenshot 2022-03-01 at 18 05 50|

Ooh, that's twice as fast!

Quicker feedback in CI/CD 😁 

## Any risks?

At the moment, the release has no dependency on integration tests – so if an integration test failed as result of a merge to main, we would still issue a release.

This has never happened before! But it's something that could happen. What do we think?

We could have `release` depend upon `build`, but unless we persisted an artefact between these two workflows, we'd need to run install and build in the release job, which would inflate our build time accordingly.


davidfurey:
Does this change really achieve the goal of quicker feedback in CI/CD, or does it just make it more likely that integration test failures will be ignored?

jonathonherbert:
@davidfurey unsure if this addresses your concern – it won't be possible to merge PRs without both checks passing.

In this way, the behaviour of CI remains as before.

It's only in CD, when the PR is merged to main, that these tests would not block a release.

Unsure if that alters your understanding. This still isn't ideal!

I wonder if it's possible to have some logic to run these tests on CD only – that is to say, when a release is detected, to ensure that we've the same level of protection in both scenarios.

davidfurey:
It is reassuring to know that a PR cannot be merged until both checks pass.

But it is still a little concerning that the full test suite isn't required on `main` before a release.  You could add a conditional step to the workflow that runs these tests on main and beta branches only.

jonathonherbert:
@davidfurey good suggestion, I've just done that.

The solution I've found is a little repetitious, but allows us to share an environment in which dependencies are installed and a build has already been created.

We could introduce another job, `release`, that `needs: build`, IIUC this would mean either repeating the same installation and build process, or sharing [artifacts between jobs](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts), which seems a little excessive for this pattern but may be worth considering.

I'll follow up in DevX to see if there's anything else that we might reasonably do to DRY this out.

jonathonherbert:
TIL – jobs are by default run in parallel, so we achieve the same affect with multiple jobs, and can have our release correctly depend on both jobs with `needs: [build, integration-tests]`.

davidfurey (approved):

PR #213, opened 2022-03-02
chore: Add fixture based tests, and fixtures for rich-link element

jonathonherbert:
## What does this change?

Adds fixture-based tests for element data, with an example dataset for the rich-link element.

These tests allow us to pull data from Composer and roundtrip it through prosemirror-elements. Here's a sample SQL query:

```sql
SELECT element
FROM livecontent, jsonb_array_elements(content -> 'blocks') as blocks,
     jsonb_array_elements(blocks -> 'elements') as element
WHERE elements->> 'elementType' = 'embed'
LIMIT 100
```

We've been caught by small differences in our data before, and these tests should help make these differences visible before we ship new elements.

We have to compromise in a few places because it's sometimes reasonable for our data to be modified by this library:

- some rich links are missing a `role` field, which we supply by default
- some rich links are missing an `isMandatory` field, which we supply by default

As a result, in our test, where these properties don't exist on incoming elements, we ignore them.

## How to test

- Do the automated tests pass?
- Is it roughly clear what they're doing, and why the compromises mentioned about have been made?

### DIFF
--- a/src/elements/helpers/__tests__/fixtures/index.ts
+++ b/src/elements/helpers/__tests__/fixtures/index.ts
@@ -1,5 +1,9 @@
 import richLink from "./rich-link.json";
 
+/**
+ * A map of our element fixtures, where the key represents the element name that
+ * corresponds to the fixture, and the value is the fixture object.
+ */
 export const allElementFixtures = {
   "rich-link": richLink,
 };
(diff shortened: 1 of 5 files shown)


rhystmills (approved):
A really useful change - having tried it on the membership element, this helped:
- Figure out two optional fields that were missing from our definitions
- Feel comfortable that no other unintended changes had been introduced by our work

As discussed in person earlier - it would be good to add instructions to the readme as to how fixtures can be added.

(Also all tests passing locally)

PR #214, opened 2022-03-02
feat: Add table element

jonathonherbert:
## What does this change?

This PR adds the membership element, which is an element representing some HTML that contains tabular data. It's chiefly, and regularly, used by Sport to embed premier league tables, player profiles etc.

|Before|After|
|--|--|
|Screenshot 2022-03-08 at 21 40 33|Screenshot 2022-03-08 at 21 40 18|

## How to test

- We've added fixtures containing a hundred tables randomly sampled from the production database. They should pass, indicating that this change should be non-destructive for existing data.
- Add a table element. It should correctly display a table (I've included CSS in this PR to style it that's been copied from our CMS).

## Accessibility


src/renderers/react/createReactElementSpec.tsx
```diff
@@ -9,7 +9,7 @@ import { ElementProvider } from "./ElementProvider";
 
 export const createReactElementSpec = <FDesc extends FieldDescriptions<string>>(
   fieldDescriptions: FDesc,
-  consumer: Consumer<ReactElement, FDesc>,
+  consumer: Consumer<ReactElement | null, FDesc>,
```
jonathonherbert, line 12: This matches the signature of the spec with React's definition of a React component, allowing us pass the TableForm component directly to 
 `createReactElementSpec`:

```ts
createReactElementSpec(tableFields, TableForm);

// rather than

createReactElementSpec(tableFields, props => <TableForm {...props} />);
```

src/elements/helpers/defaultTransform.ts
```diff
@@ -20,16 +21,27 @@ type FlexibleModelElement<FDesc extends FieldDescriptions<string>> = {
   assets?: string[];
 };
 
-export const transformElementDataIn = <
-  FDesc extends FieldDescriptions<string>
->(): TransformIn<FlexibleModelElement<FDesc>, FDesc> => ({ fields }) => {
-  return ({ ...fields } as unknown) as FieldNameToValueMap<FDesc>;
+export const transformElementDataIn = <FDesc extends FieldDescriptions<string>>(
+  transformRole: boolean
```
jonathonherbert, line 25: Rather than create a new transformer, I modified the existing one to conditionally handle the role transform. This can probably be backported to a few other elements if we're happy with this approach.

Ideally, it'd be nice to write a series of transformers that we can composer, e.g.

```ts
createTransformIn(
  transformRole,
  transformIsMandatory(true),
  // ... additional transforms
)
```

to get code that's more reusable, but it's quite difficult to do typesafely, and is beyond the scope of this PR.

Fweddi: I like the direction of travel here. Transformer code is fairly repetitive at the moment.

Fweddi (approved):
LGTM.

PR #215, opened 2022-03-03
chore: Pin 3rd party actions to specific commit hash

jonathonherbert:
## What does this change?

Pins our semantic-release related GHAs, which are third party actions, to a particular commit hash, in line with the security advice for [third party actions.](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)

## How to test

- Does the release portion of the github action still run as expected?
- Does the PR linting still run as expected?

### DIFF
--- a/.github/workflows/node-js.yml
+++ b/.github/workflows/node-js.yml
@@ -23,7 +23,7 @@ jobs:
       - run: yarn ci
       - run: yarn build
       - name: Semantic Release
-        uses: cycjimmy/semantic-release-action@v2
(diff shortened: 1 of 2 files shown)


rhystmills (approved):
Looks sensible - have checked the commits and they match [2.7.0](https://github.com/cycjimmy/semantic-release-action/commit/5982a02995853159735cb838992248c4f0f16166) and [3.4.0](https://github.com/amannn/action-semantic-pull-request/commit/db6e259b93f286e3416eef27aaae88935d16cf2e) respectively

Issue #181, opened 2022-01-24
Naming: let's make sure Props is only used to describe React components


jonathonherbert:
We've got a few places in the codebase where we use the word 'props' to describe options passed in a non-React context. This has confused some developers arriving at the project, who wonder what these types have to do with React.

Let's find another word to describe these things.

PR #216, opened 2022-03-04
chore: Rename CustomField.Props to Config

jonathonherbert:
## What does this change?

Rename CustomField.Props to Config, addressing #181. 

## How to test

Is `Config` a good name, rather than `Props`?

### DIFF
--- a/src/elements/demo-image/DemoImageElementForm.tsx
+++ b/src/elements/demo-image/DemoImageElementForm.tsx
@@ -122,19 +122,21 @@ const ImageView = ({ field, onChange }: ImageViewProps) => {
         <button
           onClick={() => {
             if (imageFields.mediaId) {
-              field.description.props.onCropImage(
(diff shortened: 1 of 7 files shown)


Fweddi (commented):
This is a good direction of travel as the term `props` is overloaded, but I think I would prefer something a bit more descriptive than `config` here.

1. I usually assume config to have a scope that is more global than it is local.
2. I note that you invoke methods on the config - usually I would assume config methods to be optional (based on the assumption that the methods could be absent if the config is empty).

Perhaps options, or customMethods, would be more appropriate.

jonathonherbert:
@Fweddi agreed. Closing as I think this will be easier to address once the repeater PRs are merged – it's likely to be out of date by then.

I'm also tempted to just remove this property after #248, as I'm not convinced the API is easy to use, and colocating our element rendering code with the `createReactElementSpec` function makes it largely redundant.

PR #217, opened 2022-03-07
fix: Consolidate two methods of moving from node -> field data

jonathonherbert:
## What does this change?

At the moment, we have two ways of serializing nodes representing fields into the values they represent:

- A method, `getElementDataFromNode`, which does what it says on the tin
- Instance methods on each `FieldView` instance, `getNodeValue`, which given a field node pass back data.

Although they're called in different contexts and with different sorts of data, they both at root do the same thing – translate node representing field data into the data itself.

We should only have one way of doing this. I've consolidated both ways into a single method, `getFieldValueFromNode`, and removed the relevant method(s) from our `FieldView` interface.

This fixes a problem where we treat text nodes as rich text on serialisation within the renderer, which means we're incorrectly escaping HTML chars in our serialised output. For an example, insert the interactive element in the demo, and click on 'original URL'. You should find that there's an `&amp;` string in the URL that breaks the resulting link.

## How to test

- Insert an interactive element in the demo, and click 'original URL'. You should find that the URL now works.

### DIFF
--- a/src/plugin/__tests__/plugin.spec.tsx
+++ b/src/plugin/__tests__/plugin.spec.tsx
@@ -4,28 +4,19 @@ import type { EditorView } from "prosemirror-view";
 import { Decoration, DecorationSet } from "prosemirror-view";
 import { createElementSpec } from "../elementSpec";
 import { ProseMirrorFieldView } from "../fieldViews/ProseMirrorFieldView";
+import {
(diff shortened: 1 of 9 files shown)


SHession (approved):
Reviewed the code and tested locally. The changes are sensible, LGTM.

PR #219, opened 2022-03-07
chore: Move element samples into separate folder, and create element buttons from data

jonathonherbert:
## What does this change?

A few changes to make our demo more concise, as working on the table element pushed me over the edge 😅 

- Moves our element samples into a separate folder
- Creates our element buttons from data, rather than calling `createElementButton` for each element.

## How to test

In the demo, the element creation buttons should continue to work as expected. 

in `demo/index.ts`, things should look a bit neater.


Fweddi (approved):
This looks good to me. The only change I am unsure on is 661b34da - I'm not sure what exactly 'Make every field in the standard element absent on empty' means.

demo/index.ts
```diff
@@ -251,91 +279,46 @@ const createEditor = (server: CollabServer) => {
...
+    { label: "Code", name: codeElementName, values: sampleCode },
+  ] as const;
+
+  buttonData.map(({ label, name, values }) =>
+    createElementButton(label, name, values)
+  );
 
   const imageElementButton = document.createElement("button");
-  imageElementButton.innerHTML = "Add image element";
+  imageElementButton.innerHTML = "Image element";
```
Fweddi, line 321: A bold change!

image

jonathonherbert:
@Fweddi discussed out of band but to confirm – this means that when these fields contain empty strings, they're not added as properties to the object created by prosemirror-elements when we translate a prosemirror-node to element data.

PR #220, opened 2022-03-07
feat!: Adds basic telemetry sending functionality

SHession:
## What does this change?

The PR aims to add some basic telemetry in order to improve the observability of how our users interact with elements.
It makes use of the new [Ed Tools telemetry package](https://github.com/guardian/editorial-tools-user-telemetry-service/pull/10) to achieve this, and adds telemetry events to the element commands only. 

We will likely want to expand this API so users can add their own telemetry events to elements.

## How to test

Add an element in the sandbox environment, interact with the element using the wrapper buttons (move arrows, select and delete). Observe outgoing network requests to the telemetry endpoint (i.e. "example.com").

We have visibility of user's interacting with element commands.

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -135,13 +135,14 @@ const {
     membership: membershipElement,
   },
   {
-    sendTelemetryEvent: (type: string) =>
+    sendTelemetryEvent: (type: string, tags) =>
       telemetryEventService.addEvent({
         app: "ProseMirrorElements",
         stage: "TEST",
         eventTime: new Date().toISOString(),
         type,
         value: true,
+        tags,
       }),
   }
 );
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -1,4 +1,5 @@
 import { FocusStyleManager } from "@guardian/src-foundations/utils";
+import { UserTelemetryEventSender } from "@guardian/user-telemetry-client";
 import omit from "lodash/omit";
 import type OrderedMap from "orderedmap";
 import { collab } from "prosemirror-collab";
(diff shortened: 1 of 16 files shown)


src/plugin/helpers/prosemirror.ts
```diff
@@ -126,47 +133,92 @@ const moveNode = (consumerPredicate: Predicate) => (
...
-const moveNodeUp = (predicate: Predicate) => (
-  getPos: () => number | undefined
-) => (view: EditorView, run = true) =>
-  moveNode(predicate)(getPos, view.state, run && view.dispatch, "up");
+const moveNodeUp = (
+  predicate: Predicate,
+  sendTelemetryEvent: SendTelemetryEvent
+) => (getPos: () => number | undefined) => (view: EditorView, run = true) => {
+  const sendEvent = () =>
+    sendTelemetryEvent?.(CommandTelemetryType.PMEUpButtonPressed);
```
jonathonherbert, line 141: Reckon it's worth having a separate event for 'top' and 'bottom' presses – be really interested to know if they're used!

jonathonherbert (commented):
Tested locally, and works as expected.

Discussed this approach out of band, and this works well, but adds knowledge of the telemetry service into commands within our plugin, where side effects are usually reserved for dispatching transactions.

We mentioned two other approaches – 
- making  this a concern of the UI, and adding the service calls directly to the UI. This would likely require a refactor to ensure that a Guardian-specific `createReactElementSpec` was passed the relevant configuration from the element plugin on instantiation.
- keeping this a plugin concern, but making the callback generic: for example, providing an `onCommand` callback that provided details of the command being run, which could be wired into telemetry by the consuming code.

Both approaches would be extra work that's probably not warranted in this case, but worth mentioning.

One blocking comment re: 'top' and 'bottom' commands, but otherwise this looks good and will be very useful!

demo/index.ts
```diff
@@ -97,34 +98,53 @@ const {
...
+    membership: membershipElement,
+  },
+  {
+    sendTelemetryEvent: (type: string) =>
+      telemetryEventService.addEvent({
+        app: "ProseMirrorElements",
+        stage: "TEST",
+        eventTime: new Date().toISOString(),
+        type,
+        value: true,
```
jonathonherbert, line 144: Interesting – wonder if this is a hint that we should relax our interface, as sometimes value is not required?

SHession: I had the same thought!

jonathonherbert (approved):
Nice – the UI-based approach definitely feels better – and opens the door to other UI-based events, nice to see recropping in there! 👍 

Tested locally, events look good.

There's one piece of work we could follow up with that could make this generic, and I think there may be case for it in the future – there may be other sorts of configuration that is received by elements but can only be passed in at the point of plugin creation.

We could make the object we're passing (with the `sendTelemetryEvent`) a type parameter, so that elements could specify their configuration requirements from the plugin, and the plugin would have to satisfy them.

Happy to discuss further, but I think for now this does a good job 👌

PR #221, opened 2022-03-10
feat: Add element for deprecated elements (witness, vine, instagram)

jonathonherbert:
## What does this change?

Adds an element to handle the three element types we have that we know are deprecated – the witness, vine, and instagram element types.

Screenshot 2022-03-10 at 14 56 33

Because we don't know what might be in these elements, I've added a way of checking out the raw data, hidden in a details tag.

## How to test

- The automated tests should pass, which include fixtures that represent 1000 elements of each type. This should give us confidence that roundtripping data in these elements will not be destructive.
- Add the deprecated elements in the UI. Does it make sense? Does a generic 'show data' details tag do the job, or do we want to be more specific about what we show here on a per-element basis?
- Does the approach I've taken – stringifying our data and adding it all into a  'data' field – make sense?


Fweddi:
Looks great, I really like the approach. The placeholder data for the instagram element is actually a vine, but that's a very small nitpick.

jonathonherbert:
@Fweddi hah, nice spot. Just updated.

Fweddi (approved):

PR #223, opened 2022-03-14
chore: Move semantic-release out of github action into package.json script

jonathonherbert:
_co-authored-by: @SHession_ 

## What does this change?

Moves our semantic-release setup out of github action into package.json script. There are two reasons:
- We must configure our release to match our PR linter, and this was not straightforward with the GHA
- It's better not to have 3rd-party actions at all, for security reasons.

## How to test

Take a look at the [recent runs](https://github.com/guardian/prosemirror-elements/actions/runs/1980314132) on the `beta` branch. They should:
- result in a release
- respect the `!` suffix to the commit prefix, which indicates that they're using the `conventionalcommits` preset, proving our configuration works.

Less confusion when issuing breaking change releases, and one fewer third party dependency in our actions. (Still a dependency in our package.json, mind!)

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -69,16 +69,4 @@ jobs:
         run: yarn build
 
       - name: Release
-        uses: cycjimmy/semantic-release-action@5982a02995853159735cb838992248c4f0f16166
-        with:
-          branches: |
-            [
-              'main',
-              {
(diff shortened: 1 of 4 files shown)


SHession (approved):
This looks great! It's difficult to confirm it works exactly as intended but we can hopefully refine it if we need to.

PR #224, opened 2022-03-15
chore: Add additional plugins to release to npm and github

jonathonherbert:
## What does this change?

#223 is missing a few plugins in our semantic release configuration that we need to publish to NPM, issue a release in Github, and fancy up our release notes.

This PR adds those changes.

## How to test

Take a look at the `beta` branch., which contains these changes. Is there a new release in `npm` as a result of the push?

Releases work as before.

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -69,4 +69,7 @@ jobs:
         run: yarn build
 
       - name: Release
+        env:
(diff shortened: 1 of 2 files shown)


SHession (approved):
Looks great, I can see the release in [npm](https://www.npmjs.com/package/@guardian/prosemirror-elements) and the tags / release in the repo look correct!

PR #225, opened 2022-03-16
chore: Add a script to sanitise element fixtures

jonathonherbert:
## What does this change?

Adds a script to sanitise element fixtures as we generate them, parsing the element data out of string format and redacting certain fields that we'd prefer not to commit to a public repository.

Also includes data for the `content-atom` element, currently WIP.

## How to test

- Does the script describe itself adequately?
- Stretch goal – can you use it on a given dataset? (Happy to supply example data! It's been used to create all of the fixtures we currently have)

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,3 +78,9 @@ const mySchema = new Schema({
   marks
 });
 ```
+
+## Element-specific docs – for Guardian users only
+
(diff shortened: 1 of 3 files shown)


SHession (approved):
Script looks useful, and the code makes sense. 
It would be difficult to discover the file if you didn't already know about it. It might be worth mentioning it in a README or perhaps moving it into a more noticeable directory.

PR #210, opened 2022-03-01
feat: Convert video element to 'standard' element

jonathonherbert:
## What does this change?

Renames the `video` element to a `standard` element, and adds the supported types to the relevant data transformer.

A standard element represents every element covered by https://github.com/guardian/flexible-model/blob/f9d30ad2bb19446a9226ab1bd8418b4aaa03d762/src/main/thrift/content.thrift#L696.
 
In practice, it also covers the legacy (non-content-atom) audio and video elements, which only use a subset of their flexible-model fields.

This PR makes one modification to the element fields – every text-based element field is removed if empty, to match the current element behaviour.

Per the description in flexible-model, the data requirements for this element also match the `table` element, but that has a diferent presentation, and so we'll add it in a separate PR.

## How to test

It's worth testing with a variety of elements of different types, from different periods, in our consuming CMS' CODE environment. I've included a sample of pieces from that env, ordered by creation date, to check.

This PR contains fixture tests for each element type per #213, which offer some assurance that our new element type will not affect existing elements.

<details>
<summary>Pieces with video elements (these should be unchanged)</summary>
617805ee8f08f18a166abde6,2021-10-26
61d4740d8f08f06c8872433b,2022-01-04
6013f63a8f08be3f4348fd83,2021-01-29
5b8959e3e4b05cd0bedf4a7b,2018-08-31
610969698f088b0b4e9fdcd8,2021-08-03
619682ab8f08664e0af85c89,2021-11-18
5fc617948f086ce9248f3761,2020-12-01
5ff6e97f8f08c7a368e6e4ae,2021-01-07
600800b28f0850df55867fa4,2021-01-20
60ee4bb28f08331a6b2864e4,2021-07-14
60ee4bb28f08331a6b2864e4,2021-07-14
601a859b8f0837302b86d97c,2021-02-03
60cb61488f081fcfa2d35b01,2021-06-17
60cb61488f081fcfa2d35b01,2021-06-17
611bc8008f085022754cbe31,2021-08-17
60420bfc8f08e95b70aead0c,2021-03-05
6059c7cf8f083158d979fecf,2021-03-23
606dade58f081a7273c4dea2,2021-04-07
607d57568f08b5689bedf1e1,2021-04-19
5ad6f5eae4b04e8efbbb7457,2018-04-18
</details>

Screenshot 2022-03-01 at 09 04 04

<details>
<summary>Pieces with audio elements</summary>
60cb61488f081fcfa2d35b01,2021-06-17
[...]

PR #226, opened 2022-03-16
fix: Reset white-space for iFrame layout component

jonathonherbert:
## What does this change?

... because our consuming CMS has CSS opinions about white-space that break the layout.

|Before|After|
|--|--|
|Screenshot 2022-03-16 at 15 46 16|Screenshot 2022-03-16 at 15 44 08|

## How to test

Via `yarn yalc`, use this package in Composer. Does the layout bug disappear?


SHession (approved):
Spotted this locally when testing #210 but forgot to flag. This fixes the issue locally in PME!

PR #227, opened 2022-03-17
feat: Rearrange standard element

jonathonherbert:
## What does this change?

Makes the 'standard' element a little nicer across the board:

- Moving the fields about to be more consistent with other elements
- Providing a 'preview' heading in line with the interactive and embed elements, which can contain the original content link
- Providing a large preview for documents, which were previously unreadable

|Before|After|
|--|--|
|Screenshot 2022-03-15 at 15 50 15|Screenshot 2022-03-17 at 13 04 24|
|Screenshot 2022-03-15 at 15 50 24|Screenshot 2022-03-17 at 13 04 46|

I would say that the result is that everything is a bit neater, and document element previews are approx ~50% less tragic than before.

Do note that the aspect ratio and size of a document iframe will make it appear very large in wide viewports. In Composer, this should not be a problem because of the fixed width of our elements.

## How to test

Fire it up locally and have a test, or use in Composer.

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -125,6 +125,7 @@ const {
 const standardElement = createStandardElement({
   createCaptionPlugins,
   checkThirdPartyTracking: mockThirdPartyTracking,
+  useLargePreview: false,
 });
 
 const telemetryEventService = new UserTelemetryEventSender("example.com");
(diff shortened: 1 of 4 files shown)


demo/index.ts
```diff
@@ -320,9 +325,6 @@ const createEditor = (server: CollabServer) => {
     { label: "Map", name: mapElementName, values: sampleMap },
     { label: "Document", name: documentElementName, values: sampleDocument },
     { label: "Table", name: tableElementName, values: sampleTable },
-    { label: "Audio", name: audioElementName, values: sampleAudio },
```
jonathonherbert, line 323: These duplicate existing examples above.

Fweddi (approved):

PR #234, opened 2022-04-01
feat: [repeater PR 1] Add repeater field to NodeSpec

jonathonherbert:
## What does this change?

Adds a field of type `repeater` to our `NodeSpec` generation.

A repeater field is a field that contains 0-many children, each containing their own set of field descriptions. We can imagine it being used to implement any field that includes repeated content – for example, a photo gallery, or tabular data.

This PR is a step in this direction, providing the necessary userland field creator, and modelling that field within Prosemirror. Subsequent PRs will need to flesh out:

- how this new field type interacts with field values as they are passed in and out of the RTE
- how this new field type is represented as a field within the renderer
- adding the ability to add, remove, and possibly move repeater node children in relation to their parent.

## Implementation notes

There are two ways I could think of to model our repeater nodes:

1. As a series of nodes modelled without a parent, where repeater nodes exist alongside existing fields:
```mermaid
flowchart TB
  doc-->textFieldInstance
  doc-->repeaterInstance1
  doc-->repeaterInstance2
  repeaterInstance1-->childFieldInstance1
  repeaterInstance2-->childFieldInstance2
```

2. As a single parent that contains all the repeater nodes of its type:
```mermaid
flowchart TB
  doc-->textField
  doc-->repeaterParentField1
  repeaterParentField1-->repeaterInstance1
  repeaterParentField1-->repeaterInstance2
  repeaterInstance1-->childFieldInstance1
  repeaterInstance2-->childFieldInstance2
```

(Let me know if another occurs to you, dear reader!)

The second approach looks neater, in that it's very clear where all instances of a particular repeater field are located. However, it adds another layer of complexity in the schema generation, and there's also the performance cost (unmeasured, and it may be trivial for our use-cases) of building, maintaining and serializing a larger tree during runtime.

As a result, I've opted for 1. for now, but if there turns out to be a compelling use-case for 2., it's fairly straightforward to switch – feedback would be much appreciated.

## How to test

- The existing automated tests should pass – they're pretty comprehensive, so we shouldn't have regressions.
- The new tests should pass. They ensure that the right `NodeSpec`s are created, and that they permit the appropriate content, for both single repeaters and nested repeater fields. Hopefully they are comprehensive, but it might be good to add additional tests if there are edge cases that aren't covered here.

### DIFF
--- a/src/plugin/fieldViews/RepeaterFieldView.ts
+++ b/src/plugin/fieldViews/RepeaterFieldView.ts
@@ -22,8 +22,6 @@ export interface RepeaterFieldDescription<
 
 /**
  * A FieldView representing a node that contains user-defined child nodes.
- *
- * Offers methods to add, remove, and move nodes.
  */
 export class RepeaterFieldView {
   public static fieldName = repeaterFieldName;
@@ -49,14 +47,6 @@ export class RepeaterFieldView {
     this.decorations = decorations;
 
     return true;
-  }
-
-  public addNode() {
-    console.log("To be implemented: add a node");
-  }
-
-  public removeNode() {
-    console.log("To be implemented: remove a node");
   }
 
   public destroy() {
--- a/src/plugin/fieldViews/RepeaterFieldView.ts
+++ b/src/plugin/fieldViews/RepeaterFieldView.ts
@@ -28,6 +28,7 @@ export interface RepeaterFieldDescription<
 export class RepeaterFieldView {
   public static fieldName = repeaterFieldName;
   public static fieldType = FieldType.REPEATER;
+  public static defaultValue = [];
 
   public constructor(
     private node: Node,
@@ -56,5 +57,9 @@ export class RepeaterFieldView {
 
   public removeNode() {
     console.log("To be implemented: remove a node");
+  }
+
+  public destroy() {
+    console.log("To be implemented: destroy");
   }
 }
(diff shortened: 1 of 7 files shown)


src/plugin/fieldViews/RepeaterFieldView.ts
```diff
@@ -0,0 +1,65 @@
...
+  FDesc extends FieldDescriptions<string>
+> extends BaseFieldDescription<unknown> {
+  type: typeof repeaterFieldName;
+  fields: FDesc;
+}
+
+/**
+ * A FieldView representing a node that contains user-defined child nodes.
+ *
+ * Offers methods to add, remove, and move nodes.
```
rhystmills, line 26: Should this be 'add, remove, and destroy'? (Matching the method names). What is the distinction between remove and destroy here?

rhystmills (commented):
This looks like it will be a useful addition.

I've left a question around the methods. When do you think we should add implementations for those methods?

src/plugin/fieldViews/RepeaterFieldView.ts
jonathonherbert: Great question. These were added as I tried to imagine what this FieldView would be responsible for – but I've left them unimplemented as there will be no meaningful way to test them until they're wired up to Prosemirror's state.

I've removed them for now, as no interface requires them yet, and they're probably more confusing that enlightening!

rhystmills (approved):
This looks like it will be a useful addition. The code looks sensible to me and all tests are passing locally 👍

PR #235, opened 2022-04-01
chore: [repeater PR 2] Move nodeSpec tests to nodeSpec.spec.ts

jonathonherbert:
## What does this change?

Moves some nodeSpec tests to the proper file – previously we were testing via `buildElementPlugin`, but with additional tests in #234 it makes sense to have our own spec for this functionality.

Also shuffles around some tests within `element.spec.ts` as the hierarchy was odd (validation tests nested within the serialisation description block)

## How to test

Tests should pass. This is a no-op, nothing should have changed.

### DIFF
--- a/src/plugin/__tests__/element.spec.ts
+++ b/src/plugin/__tests__/element.spec.ts
@@ -190,7 +190,7 @@ describe("buildElementPlugin", () => {
     });
   });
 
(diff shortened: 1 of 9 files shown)


rhystmills (approved):
Looks good. Running as intended locally, and all tests passing.

PR #236, opened 2022-04-01
feat: [repeater PR 3] Handle ingress of values into repeater nodes

jonathonherbert:
## What does this change?

Implements the next phase of #234 – handling the way data moves in and out of repeater nodes. See that PR for a description of the new structure of these nodes.

To achieve this, we need to add code to iterate over repeater field **values** and create nodes from them recursively.

The work to iterate over repeater field **nodes** and create values from them recursively is incomplete, and should be added in a subsequent PR – we need to consolidate the code that produces these values before changing it, as this operation currently happens in two places.

## How to test

Over half of this PR's line count is used by tests – either expanding existing ones or writing new ones. They should pass, and we should be convinced that they cover ingress and egress cases in enough depth to give us confidence.

We can also inspect the result of this work at runtime. I've added a repeater field to the demo image element. If you run the library with `yarn start` and add an element of this sort ('Add demo image'), you'll be able to see that the repeater nodes have been added to the document content, which should mirror the diagram in #234: 

Screenshot 2022-04-13 at 10 14 41

One thing to note – the code that produces repeater field values is incomplete, and as a result values from repeater fields will be incorrect. This requires a bit of a rethink into how we structure repeater nodes, and is addressed in #265.

### DIFF
--- a/src/plugin/helpers/element.ts
+++ b/src/plugin/helpers/element.ts
@@ -94,6 +94,25 @@ export const createGetElementDataFromNode = <
     return undefined;
   }
 
+  const values = getFieldValuesFromNode(
+    node,
+    element.fieldDescriptions,
+    serializer
+  );
+
+  return {
+    elementName,
+    values,
+  } as ExtractDataTypeFromElementSpec<ESpecMap, ElementNames>;
+};
+
+export const getFieldValuesFromNode = <
+  FDesc extends FieldDescriptions<Extract<keyof FDesc, string>>
+>(
+  node: Node,
+  fieldDescriptions: FDesc,
+  serializer: DOMSerializer
+) => {
   // We gather the values from each child as we iterate over the
   // node, to update the renderer. It's difficult to be typesafe here,
   // as the Node's name value is loosely typed as `string`, and so we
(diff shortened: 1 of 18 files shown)


rhystmills (commented):
All looks good other than one issue; I'm seeing a discrepancy between what is reported by the ProseMirror dev tools and the element values for the demo image:

One repeater field:
Screenshot 2022-05-20 at 09 27 04

Two repeater fields:
Screenshot 2022-05-20 at 09 27 16

jonathonherbert:
@rhystmills good spot. This PR description is misleading – the scope of the work changed as I realised that #248 was needed!

I've amended the description – this PR now only covers ingress and serialisation of repeater data.

src/plugin/helpers/element.ts
```diff
@@ -115,13 +134,28 @@ export const createGetElementDataFromNode = <
       return;
     }
 
-    values[fieldName] = value;
+    if (fieldDescription.type === "repeater") {
+      values[fieldName] = values[fieldName]
+        ? (values[fieldName] as unknown[]).concat(value)
+        : [value];
+
+      console.log({ value, values: values[fieldName] });
```
rhystmills, line 142: Do we still want this `console.log` here?

jonathonherbert: Ta for spotting, removed.

rhystmills (approved):
Looks good to me, and running as expected locally with all tests passing (other than some integration tests that also seem to be failing on main) 👍

PR #237, opened 2022-04-06
feat: add comment element

dskamiotis:
## What does this change?
Currently we do not have a Comment Element rendered in prosemirror-elements.
We would like to replicate the existing Comment element in Composer flexible-content (rendered in Angular)

## How to test
This is tested within PME locally in our sandbox environment `https://prosemirror-elements.local.dev-gutools.co.uk/`
To run locally: `yarn start`
Click on `Add Comment` button
Also running `yarn test:unit` (all tests should pass)

The element replicates the previous element as closely as possible (with noted upgrades).

## Additional Notes:
1. We are dropping the tracking embed message as this was uniformly added to all elements. This is not needed as the embed is internally ours and therefore not required.
2. Testing added: whether the new element produces accurate and consistent JSON representation (`src/elements/helpers/__tests__/fixtures/comment.json`).  This is added to the transform testing: `src/elements/helpers/__tests__/transformFixtures.spec.ts`
3.  Sizing of the font in the Dropdown menu is consistent with all new elements created in `prosemirror-elements`

### Old Comment Element
image

### New Comment Element
image

### DIFF
--- a/src/elements/helpers/__tests__/transformFixtures.spec.ts
+++ b/src/elements/helpers/__tests__/transformFixtures.spec.ts
@@ -7,7 +7,6 @@ import {
   richlinkElement,
   tableElement,
 } from "../../..";
-import { fieldNameAttr } from "../../../plugin/nodeSpec";
 import { commentElement } from "../../comment/CommentSpec";
 import { deprecatedElement } from "../../deprecated/DeprecatedSpec";
 import { createTestSchema } from "../test";


jonathonherbert (commented):
Tested in sandbox, code looks great and it behaves as expected – with one exception! At the moment, the `role` field is not transformed as it should be.

Our fixture tests currently don't pick this up, but they should! I've written a change to surface this issue in our tests – I'd suggest us pairing on including that change in this PR, as it'll then surface the problem, and we can also pair on the fix.

dskamiotis:
yes `role` is one of those that transforms the `Weighting` label. Not sure if that causes the issue. But sure lets pair 👍

jonathonherbert (approved):
Paired on the change above, looking good! 🎉

PR #238, opened 2022-04-07
chore: Specify full node version

akash1810:
## What does this change?

Include the patch number in `.nvmrc` to increase specificity and determinism.

This also allows tools like [`asdf`](https://asdf-vm.com/) to work - the [nodejs plugin](https://github.com/asdf-vm/asdf-nodejs) wants the exact version. Installing w/out a patch number results in:

```bash
➜ asdf install nodejs 16.13
Trying to update node-build... ok
node-build: definition not found: 16.13
```

It's worth noting `16.13.2` is the latest patch version on the `16.13` minor. It is _not_ the latest in the 16 major, that is `16.14.2`.

## How to test

CI should continue to pass.

Increased determinism, and less variance across machines/environments.


jonathonherbert (approved):

jonathonherbert:
Thanks @akash1810!

PR #241, opened 2022-04-22
feat: Improves publication checks

Fweddi:
## What does this change?

We used to have a feature where we checked an atom's publication state, but this was dropped for unknown reasons. This PR revives that feature and expands on the state check.

We now check for two states: whether the atom is published at all, and if it is published, whether or not there are unpublished changes available in Preview.

image

image


jonathonherbert (approved):
Tested locally by modifying the object returned in `demo/index.ts` @ L205, works as expected. I think we missed changing those so I've pushed a commit.

PR #242, opened 2022-04-22
feat: Add form element

zekehuntergreen:
## What does this change?

- Adds the form element to prosemirror elements based on the flexible-content version
   - spec and form for form element
   - button to demo page which creates form element with sample data
   - Transformer functions transformElementIn and transformElementOut to manipulate data to and from the shape required by the form element spec
   - fixture tests for the form element

## How to test

- Run prosemirror-elements locally 
- Add a form element with "Add Form" button

## Images

Flexible content version of form element in Composer
Screenshot 2022-04-22 at 12 04 56

PME Form Element
Screenshot 2022-04-22 at 11 44 38

An example of a form element in an article
Screenshot 2022-04-22 at 13 00 14

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -23,6 +23,7 @@ import {
   tableElement,
 } from "../src";
 import { deprecatedElement } from "../src/elements/deprecated/DeprecatedSpec";
+import { createFormElement } from "../src/elements/form/FormElementSpec";
 import {
   transformElementOut,
   undefinedDropdownValue,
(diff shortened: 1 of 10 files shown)


jonathonherbert (approved):
This is looking great, @zekehuntergreen – thanks for pairing with us on this! I've made some very small changes after taking a look in Composer – because of the limited width of the editing area in Composer, I think we can safely stack our elements, like so:

Screenshot 2022-04-25 at 10 20 56

Excited to 🚢  to users, I'll copy you into the e-mails once it's out 🎉

zekehuntergreen:
Thanks again, @SHession and @jonathonherbert for pairing on this! 🙂

PR #244, opened 2022-04-26
feat: Remove form element in favour of deprecated element

jonathonherbert:
## What does this change?

Hot on the heels of #242, we've discovered that the element in question hasn't been used for a while. As a result, sadly, we won't need the UI or modelling work in that PR, although the fixture tests are still very useful. Instead, we can use the standard deprecated element type to represent our forms:

Screenshot 2022-04-26 at 15 42 09

## How to test

- Add a form element in the UI. Does it display with the deprecated type? Is the usual data available?
- Do the fixture tests still pass?


src/elements/deprecated/DeprecatedForm.tsx
```diff
@@ -14,6 +14,7 @@ const elementTypeToName = {
   instagram: "Instagram",
   vine: "Vine",
   witness: "Guardian Witness",
+  form: "FormStack",
```
Fweddi, line 17:
```suggestion
  form: "Formstack",
```

Fweddi (approved):
The most minor nitpick of all, but otherwise LGTM! Tested locally and data persists as expected. Fixture tests pass. 

I had some strange bugs due to an older version of the element appearing in the sandbox, but once this was deleted, everything worked as expected.

PR #245, opened 2022-05-04
feat: Mod-a within field selects only that field's text

rhystmills:
## What does this change?
<kbd>Mod-a</kbd> (i.e. <kbd>ctrl-a/cmd-a</kbd>) currently selects all Prosemirror elements when used within an element. 

This PR means that using <kbd>mod-a</kbd> within a text field selects only that field's text. 

## How to test
1. Run prosemirror-elements locally following the instructions in the readme.
2. Create an element, click on a text field, type some text in an element and use <kbd>Mod-a</kbd>.
3. Try this in both rich text field (e.g. the image caption) and non-rich text fields (e.g. the pullquote)\

Does this select all of the the text in the field?

## Images
| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/cypress/tests/ImageElement.spec.ts
+++ b/cypress/tests/ImageElement.spec.ts
@@ -438,7 +438,7 @@ describe("ImageElement", () => {
         );
       });
 
-      it.only("should select the correct range in ProseMirror when the select all shortcut is used within the field", () => {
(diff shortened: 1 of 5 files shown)


jonathonherbert (approved):
Tested locally, works as expected. Suspect the reason this surfaced was the additional UI provided by #183 – great to have a robust fix!

PR #246, opened 2022-05-09
fix: Minor tweaks to element fields in response to expanded fixture tests

jonathonherbert:
## What does this change?

Filling out our fixture tests, this PR makes a few minor tweaks to fields and data transformers to correspond with CMS behaviour:

- A few fields are missing! (although unlikely to be used downstream)
- A few fields are absent from the model when empty.
- A few elements are missing an `assets` property (which is always empty.)

## How to test

- The changes should look fairly sensible.
- The fixture tests should pass, and should cover every element in our repository (apart from the test element!)

Closer behaviour to PROD.


rhystmills (approved):
This all looks good to me!

I found that the first time I ran the unit tests, it took an unexpectedly long time(98 seconds) versus 4-5 seconds on main) - most of which was spent on `transformFixtures.spec.ts`. When running the tests subsequently , they completed in 4-5 seconds. I did not experience this on `main`.

Regardless, all seems to be working for me. Perhaps it was a one off - though it would be good to see if any other developers experience the same behaviour.

jonathonherbert:
@rhystmills thanks for this. I'm a bit suspicious of these runtimes as well – I think the type checker may be doing lots of work to infer types from these .json files. I'll do a bit of research and try to fix before I merge.

rhystmills:
> @rhystmills thanks for this. I'm a bit suspicious of these runtimes as well – I think the type checker may be doing lots of work to infer types from these .json files. I'll do a bit of research and try to fix before I merge.

Re-running after your changes, I found it took 14 seconds on first run and 7 seconds subsequently, which seems fine ti me.

PR #248, opened 2022-05-13
feat!: [repeater PR 4] Colocate `value` and `error` properties with `Field` type

jonathonherbert:
## What does this change?

At the moment, when we're representing an `Element` in the UI, we're passed three properties, `field`, `fieldValues`, and `errors`. Here's what they might look like at runtime for an element with a single field, `html`, with the value 'content':

```ts
// fields
{
  html: {
    description: FieldDescription,
    name: 'html',
    // ... a few other field-related things
  }
}

// fieldValues
{
  html: 'content'
}

// errors
{
   html: []
}
```

Lots of repetition of that field name in different objects, right?

This PR colocates our `error` and `value` properties within the initial `Field` type, giving

```ts
{
  html: {
    description: FieldDescription,
    name: 'html',
    errors: [],
    value: 'content'
  }
}
```

This is a big ergonomic improvement:

- We can pass a single `Field` into components that need to know everything about a field:
```ts
// Before – note the potential for mistakes!
<FieldWrapper headingLabel="Caption" field={fields.caption} errors={errors.caption} value={fieldValues.caption} />

// Now
<FieldWrapper headingLabel="Caption" field={fields.caption} />
```
- We simplify the types our consuming code (and our library) needs to understand.
- As a result, this PR removes hundreds of lines from our element-building code 🥳 

## Why didn't we do this before?

The duplicate of the field structure across difference sorts of values is a hangover from a previous API that didn't require the extra context we now need to expose as part of our changes to the [elements field model](https://github.com/guardian/prosemirror-elements/blob/main/docs/decision-records/000-element-fields-model.md).

It's been at the back of my mind for a while, but other feature work took priority! It makes sense to address it before the repeater feature work, as doing this work after that feature is complete is likely to be more effort.

## How to test

This is a big change – it alters the signature of our `Field` type, which is used from the point we instantiate fields in a Prosemirror context, to the point at which we use fields in our renderer context. In short, this type threads through the entire plugin lifecycle and into consumer code, so it touches a lot.

However, it's also _almost_ a no-op – everything should continue to work as expected. I've added tests to the two new functions that have been broken out for this change, `getFieldsFromNode` and `updateFieldsFromNode`, to test that they add and update values correctly.

There's one change that does affect behaviour – this PR fixes an issue @rhystmills noticed in a [previous PR](https://github.com/guardian/prosemirror-elements/pull/236#pullrequestreview-979724849). Now, we correctly serialise out repeater node values, and I've altered the DemoImage component to demonstrate that the initial repeater field values are correctly passed to the component: 

Screenshot 2022-07-18 at 08 45 55

The changed type at the heart of this PR is in `Element.ts`, where we've added two new properties to our `Field` type:

```ts
  value: F extends FieldView<infer Value> ? Value : never;
  errors: ValidationError[];
```

The rest of the logic to create and update this type is now in `plugin/field.ts`. This has new unit tests that should give us confidence that they work as expected.

All the other changes, from the plugin logic to the element updates, should be adjustments to make use of this new type signature.

Try adding and updating lots of elements in the UI. Creating elements, updating their values (either via the UI or programmatically, via e.g. 'copy to caption' in the Image element), and validation (e.g. required fields, etc.) should all work as expected.


rhystmills (approved):
Code looks good to me, tests passing and not noticing any issues running locally 👍 

Colocating `errors` and and `fieldValues` under `field` seems much neater.

PR #252, opened 2022-06-08
Handle absent suppliersReference

jonathonherbert:
## What does this change?

Sometimes, `suppliersReference` is absent in images. We don't account for this in some of our types.

By coincidence, when we instantiate an image element node directly, we provide this value via the ingress transformer – but we don't do this when we recrop. This means that if recropped images are missing a suppliersReference field, they'll blow up on egress.

This PR fixes this and adds a test.

## How to test

- The test should pass, and they should convince you we're testing the right thing.
- Within the editor, try adding an image without a `suppliersReference`, and recrop it. We shouldn't see any red ink, and the content should save as normal.


rhystmills (approved):
Looks good to me 👍

PR #259, opened 2022-07-13
docs: adding troubleshooting docs

rebecca-thompson:
## What does this change?
Adds some advice on a possible culprit if a release is not triggered after merging a PR.


dskamiotis (approved):
Could this also be done with `--amend` flag? Or would that mudd the instructions with too many permutations?

Otherwise really useful addition to docs :)

Thanks! I think there are also notes on that in the linked documentation if that approach is appropriate

PR #265, opened 2022-09-02
feat: [repeater PR 5] Restructure repeater fields

jonathonherbert:
## What does this change?

#234 contains a description of two ways we might model repeater nodes in our document: 

1. As a series of nodes modelled without a parent, where repeater nodes exist alongside existing fields:
```mermaid
flowchart TB
  doc-->textFieldInstance
  doc-->repeaterInstance1
  doc-->repeaterInstance2
  repeaterInstance1-->childFieldInstance1
  repeaterInstance2-->childFieldInstance2
```

2. As a single parent that contains all the repeater nodes of its type:
```mermaid
flowchart TB
  doc-->textField
  doc-->repeaterParentField1
  repeaterParentField1-->repeaterInstance1
  repeaterParentField1-->repeaterInstance2
  repeaterInstance1-->childFieldInstance1
  repeaterInstance2-->childFieldInstance2
```

Initially, we went with 1. Although this results in fewer nodes, managing the transition from nodes to values and vice-versa becomes complicated – there's no guarantee where repeater nodes will exist within an element, or whether there are any at all. As a result, you must accumulate repeater nodes at the root of the document.

Approach 2 makes writing a recursive approach to mapping between fields and values more practical – you can treat the parent repeater node as the root of a new element, and apply the same mapping operation that you'd do to a top-level element, adjusting for an offset.

This PR makes that switch, and completes the incomplete mappings between fields and values for repeater nodes. As a result, we have repeater nodes that we can (de)serialize, render, and edit.

## How to test

The automated tests should pass. There are unit tests for node generation, (de)serialisation, and updating field values and FieldViews in place.

Add a demo image element, and take a look at the 'repeater' fields at the end of the element. You should see the values have been carried through from the sample element indicated in `sampleElements.ts`. Editing these elements should work as expected, and when you refresh the page, the values should persist.


rhystmills (approved):
Code looks good to me, all tests passing locally and the repeater field in the demo is working as expected 👍

PR #266, opened 2022-09-02
chore: Add prosemirror-dev-tools as a dependency

jonathonherbert:
## What does this change?

At the moment, we use a CDN to source the prosemirror-dev-tools. Why do that, when you can include it as a npm package? 

This PR also tidies up the `index.html` a bit, and removes an `index.html` in the root of the project that as far as I can tell is redundant.

## How to test

Run `yarn` in the project root, and then run the project normally. You should still see the devtools icon in the right hand side of the page. Open it – the devtools should work as expected: 

tr-dev-tools


ParisaTork:
So glad the owl is back!

Owl

Tested and can confirm this has been successful:

Typing

demo/style.css
```diff
@@ -155,6 +155,341 @@ h6 {
...
+  position: relative;
+}
+
+.ProseMirror {
+  word-wrap: break-word;
+  white-space: pre-wrap;
+  white-space: break-spaces;
+  -webkit-font-variant-ligatures: none;
+  font-variant-ligatures: none;
+  font-feature-settings: "liga" 0; /* the above doesn't seem to work in Edge */
```
ParisaTork, line 168: You're right to highlight issues of cross-browser compatibility, as this only works in 2 specific versions of Edge: https://caniuse.com/?search=font-feature-settings

demo/style.css
```diff
@@ -155,6 +155,341 @@ h6 {
...
+  left: -32px;
+  right: -2px; top: -2px; bottom: -2px;
+  border: 2px solid #8cf;
+  pointer-events: none;
+}
+
+/* Protect against generic img rules */
+
+img.ProseMirror-separator {
+  display: inline !important;
```
ParisaTork, line 205: Is !important absolutely essential here?

demo/style.css
```diff
@@ -155,6 +155,341 @@ h6 {
...
+  font-size: 12px;
+}
+
+.ProseMirror-invalid {
+  background: #ffc;
+  border: 1px solid #cc7;
+  border-radius: 4px;
+  padding: 5px 10px;
+  position: absolute;
+  min-width: 10em;
```
ParisaTork, line 458: I've seen a few examples of px and em units mixed together in this file. Would it be worth having a discussion around whether we should stick to relative units throughout?

ParisaTork:
I've added a few tiny comments to open up some interesting conversations about some choices we currently make in CSS. This is purely driven by curiosity and congrats on this PR! 🎉

ParisaTork (approved):

demo/style.css
jonathonherbert: So sorry, Parisa, I took your comment as an approval! My bad.

This is all imported straight from Prosemirror (https://prosemirror.net/css/editor.css), and it's super low stakes.

Perhaps we could keep it in a separate file to make its provenance clearer? Happy to raise another PC.

ParisaTork: I've approved all the changes, so you were right to do so! Thanks for clarifying here, this all looks good

demo/style.css
ParisaTork: ^Resolved given explanation above in another style.css review

PR #267, opened 2022-09-02
feat: [repeater PR 6] Add and remove repeater nodes in place

jonathonherbert:
_NB: depends on #280 and #281 – please review these first!_

## What does this change?

Adds an API to add and remove repeater nodes in place, exposed on the `Repeater` field.

They're exposed on the repeater field passed to the UI, and the index array of `children` matches the index we should pass to `remove`. For example, in React, a repeater field called `images` might have the following markup to enable adding and removing repeater children:

```tsx
<div>
  {fields.images.children.map((repeater, index) => (
    <li key={repeater.__ID}>
      <button onClick={() => fields.repeater.view.remove(index)}>Remove</button>
      {/* content */}
    </li>
  ))}
  <button onClick={() => fields.repeater.view.add()}>Add</button>
</div>
```

See the `DemoImageElementForm` for an example of how this might work with nested repeaters.

For testing:

- The automated tests should pass. I've added unit and integration tests for add and remove operations.
- Add and removing repeater fields in the demo element should behave as expected, across multiple editors (the cursor logic is slightly buggy at the leading edge of content at present, but the document state should always be correct):

repeater


dskamiotis (approved):
Tested Locally with multiple editors, adding removing repeater nodes. These are persisted and consistent between editors.

the Nodes are also accurately represented in the Prosemirror dev tools, displaying the various nested and parent Nodes. Showing the text content in each. 
Great work - difficult to see anything with the code and ace unit tests to top it off :)

PR #270, opened 2022-09-27
feat: add campaign callout list element

marjisound:
Paired on this: @abeddow91 & @marjisound & @jonathonherbert 
## What does this change?

This PR creates a new prosemirror element `campaign-callout-list` through which would fetch the campaigns using the `fetchCampaignList` prop and render a dropdown showing the callouts. 

image

When the user chooses one of the callouts, the `callout` element shows up with the details about the callout. 
image

Whilst pairing we also noticed that there were improper imports of `undefinedDropdownValue` from `membershipSpec`. We have removed this export in favour of always importing from `helpers/transform`. 

## How to test

1- The component was added to the demo project, so was tested by running `yarn start` and playing around with it.

2- This component was also tested with composer locally

TODO:
Needs to be tested in CODE env

## Accessibility


demo/index.ts
```diff
@@ -131,6 +135,69 @@ const mockThirdPartyTracking = (html: string) =>
         reach: { unsupportedPlatforms: [] },
       });
 
+const mockCampaignList = [
```
jonathonherbert, line 138: Let's put this in `sampleElements`, as it's effectively fixture data.

jonathonherbert (approved):
Looks great. A few non-block comments –

- Does it make sense to rename `CampaignCalloutListSpec.tsx` -> `Callout.tsx`?
- It'd be nice to consolidate the types related to callouts (`Fields`, `Rules`, `Campaign`) with the existing callout components, either in this PR or a follow-up.

marjisound:
Thanks @jonathonherbert for looking at the PR. We just addressed 2 of your comments regarding the name changing. 
We also added the `html` to the fields so that can be used for composer preview. 
Could you please have another look at the changes we did. Thank you 

> It'd be nice to consolidate the types related to callouts (Fields, Rules, Campaign) with the existing callout components, either in this PR or a follow-up.
We'll do this in a separate PR if that's OK with you.

jonathonherbert:
Yes, that's fine. For other reviewers, it might be worth noting – am I right? – that this is effectively a draft version of this element – the fields especially may change!

marjisound:
> 



> Yes, that's fine. For other reviewers, it might be worth noting – am I right? – that this is effectively a draft version of this element – the fields especially may change!

Yes that's true. This version is not going to be used by editorials. We are going to have another PR that will finalise the new Callout component with updated fields 👍

jonathonherbert (commented):
Changes look good!

One thing to mention – we should not need to store any html to preview the element in Composer. We can write some code in Composer to come up with a suitable preview, based on the campaign id.

In the medium term, it'd be nice to be able to express both the editable and preview states from within `prosemirror-elements` – this is a good reminder that at the moment, for people creating new elements, we cannot define that behaviour in this repo.

Happy to talk through the above if anything's not clear.

jonathonherbert (approved):

PR #271, opened 2022-09-30
Add a way of programmatically updating a text field's placeholder value

jonathonherbert:
## What does this change

Adds a way of programmatically updating a text field's placeeholder value.

We've added this to the view API. Calls will look like:

```ts
fields.fieldName.view.updatePlaceholder("New placeholder value")
```

## How to test

- The automated tests should pass (I've refactored the tests for the placeholder plugin), and should cover the new update functionality.
- I've altered our demo element to give us an example. Run the sandbox locally, add a 'Demo Image Element' to the editor, and click the 'update placeholder' button underneath the caption. The placeholder should update as expected.

### DIFF
--- a/src/plugin/helpers/placeholder.ts
+++ b/src/plugin/helpers/placeholder.ts
@@ -1,5 +1,6 @@
 import type { Node } from "prosemirror-model";
-import { Plugin } from "prosemirror-state";
+import type { EditorState } from "prosemirror-state";
+import { Plugin, PluginKey } from "prosemirror-state";
 import { Decoration, DecorationSet } from "prosemirror-view";
 
 export const placeholderTestAttribute = "placeholder";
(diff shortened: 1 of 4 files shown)


src/plugin/helpers/placeholder.ts
```diff
@@ -38,26 +39,51 @@ const getFirstPlaceholderPosition = (node: Node, currentPos = 0): number =>
...
+        }
+
+        const getPlaceholder =
+          typeof placeholder === "string"
+            ? getDefaultPlaceholder(placeholder)
+            : placeholder;
+
+        if (state.doc.textContent) {
+          return DecorationSet.empty;
+        }
```
marjisound, line 78: I think this statement could also be done before line 71 (`const getPlaceholder`), right?

```
        if (!placeholder || state.doc.textContent) {
          return DecorationSet.empty;
        }
```

marjisound (approved):

src/plugin/helpers/placeholder.ts
jonathonherbert: Lovely!

PR #275, opened 2022-10-13
feat: Add an onRemove callback, which is called with the removed element's field data

abeddow91:
_NB: depends on #278 – please review that first!_

## What does this change?

This PR adds a `onRemove` prop for `createCalloutElement`. `onRemove` is a handler for when the element is deleted. The parameter passed to the handler is of type `FieldNameToValueMap` which includes all the element field values. 

To be able to pass field values to onRemove, the `elementSpec.ts` was updated in order to have access to element data at render stage. 

## How to test

This was tested both with prosemirror demo and also with composer locally.

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -27,6 +27,7 @@ import {
   richlinkElement,
   tableElement,
 } from "../src";
+import type { calloutFields } from "../src/elements/callout/Callout";
 import {
   transformElementOut,
   undefinedDropdownValue,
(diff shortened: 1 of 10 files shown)


jonathonherbert:
This is on hold until we're sure that updating tags on the client is necessary.

jonathonherbert:
I've added a unit test for the React component, so I think this is ready for review – anything else to add, @marjisound, @abeddow91?

marjisound:
> I've added a unit test for the React component, so I think this is ready for review – anything else to add, @marjisound, @abeddow91?

The test looks great. Thanks @jonathonherbert . I think the PR is in a good state now for review 👍

PR #278, opened 2022-11-01
feat: Improve label accessibility

jonathonherbert:
## What does this change?

#275 requires some React-based unit tests, but in order for `testing-library` to pass tests using our React wrappers, our accessibility story for labels needed to be improved, or we get the error:

```
Found a label with the text of: field1, however no form control was found associated to that label. Make sure you're using the "for" attribute or "aria-labelledby" attribute correctly.
```

(How lovely for a testing library to point that out!)

This PR adds a test that verifies the accessibility of our labels by proxy, and adds the functionality that satisfies the test. Because we cannot add a `for` attribute to our label (`contenteditable` elements cannot be addressed by this attribute [per the spec](https://stackoverflow.com/questions/54792126/html-label-with-for-div-id-to-focus-a-contenteditable-div-does-label-label)), we add an a tag with a link to simulate the usual label behaviour – clicking on the label should give a pointer on hover, and focus the field input on click.

## How to test

- The automated test should pass.
- Have a play with the labels that relate to text fields of any sort, on any element. Hovering should show a pointer cursor. Clicking should focus the field.

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -185,6 +185,7 @@ const {
       fetchCampaignList: () => Promise.resolve(sampleCampaignList),
       targetingUrl: "https://targeting.code.dev-gutools.co.uk/",
       applyTag: (tag: string) => console.log(`Apply ${tag} tag`),
+      onRemove: (fields) => console.log("Remove callout", fields),
     }),
     interactive: createInteractiveElement({
       checkThirdPartyTracking: mockThirdPartyTracking,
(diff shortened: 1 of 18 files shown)


jonathonherbert:
cc. @marjisound and @abeddow91 – I will rebase #275 on this to enable us to add the appropriate tests for `onRemove`, so we'll want to merge this first.

abeddow91 (approved):
Nice, this looks good! Tested locally and labels have a unique id of `label-field-${Id}` with a mathing `aria-labelledby`. All tests are also passing.

PR #280, opened 2022-11-02
chore: Remove circular dependency caused by undefinedDropdownValue

jonathonherbert:
## What does this change?

Removes a circular dependency caused by `undefinedDropdownValue` – this PR moves it to a `/constants` folder.

## How to test

A ✅ in the CI should be enough to prove this is good.

Additional circular dependencies exist that mean we can't yet include a linting rule. This will be added in a follow up PR, #267.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -9,7 +9,8 @@ module.exports = {
       rules: {
         "@typescript-eslint/unbound-method": 0,
         "@typescript-eslint/explicit-module-boundary-types": 0,
-        "@typescript-eslint/no-unused-vars": 2
+        "@typescript-eslint/no-unused-vars": 2,
+        "import/no-cycle": 2
       },
     },
   ],
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -9,8 +9,7 @@ module.exports = {
       rules: {
         "@typescript-eslint/unbound-method": 0,
         "@typescript-eslint/explicit-module-boundary-types": 0,
-        "@typescript-eslint/no-unused-vars": 2,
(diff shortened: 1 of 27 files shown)


.eslintrc.cjs
```diff
@@ -9,7 +9,8 @@ module.exports = {
       rules: {
         "@typescript-eslint/unbound-method": 0,
         "@typescript-eslint/explicit-module-boundary-types": 0,
-        "@typescript-eslint/no-unused-vars": 2
+        "@typescript-eslint/no-unused-vars": 2,
+        "import/no-cycle": 2
```
akash1810, line 13: Any reason to not use [`@guardian/eslint-config-typescript`](https://github.com/guardian/csnx/tree/main/libs/%40guardian/eslint-config-typescript)? Appreciate it'll likely be a bigger change than desired for now.

I'd recommend adding `ignoreExternal` to the `import/no-cycle` config to improve execution time. See https://github.com/guardian/csnx/pull/60.

jonathonherbert: IIUC we extend that config in the line above – wonder if we're using a version that predates us adding the rule? Be nice to update it in a separate PR.

akash1810: > IIUC we extend that config in the lines above

Ah! I'll admit to being lazy, and not checking prior to the suggestion!

jonathonherbert: Ace. We can't yet add that rule in this PR, so I'll resolve this now in anticipation of #267.

dskamiotis (approved):
This looking good. CI building successfully, however a couple of conflicts to address before merging

PR #281, opened 2022-11-02
chore: Rename fieldName to fieldType

jonathonherbert:
## What does this change?

Our codebase has two sorts of things called `fieldName`:

- The name of the field as defined by devs making elements
- The type of the field, e.g. text, richText

This PR renames the latter to `fieldType`, and renames the _existing_ `fieldType` (referring to the internal modelling prosemirror-elements does to represent values) to `fieldContentType`.

## How to test

Passing tests should indicate the refactor is OK.

Does the naming make sense? Anything to change?


dskamiotis (approved):
This name change makes perfect sense, previously could easily be misinterpreted and misleading. A really useful change :)

All test passing and running locally

PR #284, opened 2022-11-15
fix: Bump prosemirror

jonathonherbert:
~_Depends on #267. In draft until that's merged and this is rebased._~ Now ready for review.

## What does this change?

Bumps prosemirror to latest. This includes a [fix to prosemirror-view](https://github.com/ProseMirror/prosemirror-view/pull/147) that will let us properly handle combinations of decorations in nested prosemirror fields

## How to test

- Internal to this module, this is a no-op – the sandbox should work as expected, and the tests should pass.
- Use `yarn yalc` to add this to a consuming application (e.g. Composer.) You should notice that when toggling hidden characters, or using Typerighter, decorations should appear in nested fields.


rebecca-thompson (approved):
I have tested this locally and everything seems to be working as expected

PR #286, opened 2022-11-22
feat: Cartoon element

rebecca-thompson:
## What does this change?
This adds a cartoon element with similar fields to the image element. The main difference is the cartoon element allows for multiple desktop and mobile images to be added and includes fields for basic styling when multiple images are present.

## How to test
You can run prosemirror-elements locally and see how the element works. Alternatively, there is a testing branch in composer (https://github.com/guardian/flexible-content/tree/bt/cartoon-element). Bear in mind, you will need to run `yalc` to pull in and depend on these changes (see this repo's README for details on how to do this). The work in Composer is still ongoing, so not everything on the test branch above will work exactly as expected.

We have a cartoon element that we can build against in composer.

This PR contains minimal/no risk to existing operations. The cartoon element will be put behind a feature flag so can be thoroughly tested before release. There's a good chance this element will change (particular with regards to the data in/out transformers as we complete the work to support this element in composer.

## Images
Screenshot 2022-12-20 at 09 30 31


jonathonherbert (commented):
Looks great! A couple of comments.

demo/helpers.ts
```diff
@@ -17,7 +17,7 @@ type GridResponse = {
         metadata: {
           description: string;
           suppliersReference: string;
-          source: string;
+          credit: string;
```
alinaboghiu, line 20: just for my understanding, what's the reason for renaming here?

src/editorial-source-components/SvgCrossRound.tsx
```diff
@@ -0,0 +1,20 @@
+// This has been taken from the Guardian Source design system:
+// https://github.com/guardian/csnx/blob/63a2a71585cf4f98b1d84e3a40aa1e8fa19f70c4/libs/%40guardian/source-react-components/vendor/icons/SvgCrossRound.tsx
+// Not available on our current version of Source, but can be pulled in when we upgrade.
```
alinaboghiu, line 3: It sounds like there's an active desire not to use Source because we don't want that coupling, so maybe we don't need this comment?

jonathonherbert:
When I recrop in desktop, the behaviour is a bit buggy – cannot capture a good GIF atm as I'm on a tethered connection, but when I add an image and then attempt to recrop the first image, I get a duplicate in the UI that disappears on refresh.

I cannot see a way to recrop in mobile, and can only recrop the first desktop image – should it be possible to recrop each crop? (If we want to avoid repetition in the UI, we could introduce e.g. recrop/remove buttons on hover)

src/elements/cartoon/CartoonForm.tsx
```diff
@@ -0,0 +1,311 @@
...
+import { cartoonFields } from "./CartoonSpec";
+
+export const cartoonElement = (
+  imageSelector: ImageSelector,
+  createCaptionPlugins: (schema: Schema) => Plugin[]
+) => {
+  return createReactElementSpec(
+    cartoonFields(imageSelector, createCaptionPlugins),
+    ({ fields }) => {
+      const [desktopImages, setDesktopImages] = useState<MainImageData[]>(
```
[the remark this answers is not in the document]
rebecca-thompson: In this scenario using the update method does not trigger a component re-render for me, but if you have it working I'd be interested to see what you've done and why I'm getting different behaviour

demo/helpers.ts
rebecca-thompson: I think this was broken before. It may be that the grid's api changed underneath

src/elements/cartoon/CartoonForm.tsx
```diff
@@ -0,0 +1,311 @@
...
+  required = false,
+  mainMediaId,
+}) => {
+  return (
+    <div>
+      <InputHeading headingLabel={label} />
+      <FieldLayoutHorizontal>
+        {images.map((image, index) => (
+          <ImageThumbnail
+            key={index}
```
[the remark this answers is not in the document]
rebecca-thompson: How odd, I can't replicate this when I run the project locally. Let's put a session in and you can show me how you're getting this

src/editorial-source-components/SvgCrossRound.tsx
rebecca-thompson: I've removed the comment as you both seem to agree (https://github.com/guardian/prosemirror-elements/pull/286/commits/35bac7740caea55e01cdea3500ed64c48fde20b6), however my personal thoughts are if we take code from a different project we should reference the source (not the Source, in this case, just where the code was lifted from)

demo/helpers.ts
paperboyo: Oh, oh, oh. This hits at our Grave Confusion. Basically, Grid’s `credit` is what CAPI (and maybe FlexAPI?) calls `source`. Can’t comment on code, but can bore to death with details…

src/elements/cartoon/CartoonForm.tsx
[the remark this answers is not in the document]
rebecca-thompson: https://github.com/guardian/prosemirror-elements/pull/286/commits/da69ef70b7eb92c590cfdcd97081aa7c59896086

src/elements/cartoon/CartoonForm.tsx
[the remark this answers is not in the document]
rebecca-thompson: https://github.com/guardian/prosemirror-elements/pull/286/commits/a37084bafca0e14544b4bdf0b5ce5ddb944b105c

rebecca-thompson:
> When I recrop in desktop, the behaviour is a bit buggy – cannot capture a good GIF atm as I'm on a tethered connection, but when I add an image and then attempt to recrop the first image, I get a duplicate in the UI that disappears on refresh.
> 
> I cannot see a way to recrop in mobile, and can only recrop the first desktop image – should it be possible to recrop each crop? (If we want to avoid repetition in the UI, we could introduce e.g. recrop/remove buttons on hover)

I'm not sure what this means, and I can't replicate the issue. Can you try again with the latest change and if you're still having problems ping me/attach a gif?

I decided against recropping for all images as it would a) make the UI busier and b) unlike with the image element it is possible to remove all but the 'primary' image in a cartoon element without having to start again from scratch, so recropping each image becomes less necessary as you can just remove it and add a new one to the element. That said, if it was useful we could add it in future. There are other features I'd like to add, such as being able to reorder the crops.

jonathonherbert (approved):

PR #288, opened 2022-12-05
feat: add error handling to callout

abeddow91:
## What does this change?
Displays a non-blocking error message if one of the following occurs:

1. The callout has not been selected. This is handled out-of-the-box and displays as a warn in composer. 
2. The callout has an active until that has expired.
3. The callout no longer exists in the list of available callouts. 

## How to test
Error 1 - The callout has not been selected
- This can be tested in prosemirror or in composer by adding a callout but not selecting anything from the drop down.
- In composer,  there should also be a warn message when hovering over the publish button.
- The article should be publishable with this error.

Error 2 - The callout has not been selected
- This can be tested in prosemirror by selecting the expired callout from the list. It is more difficult to test in composer as we filter out expired callouts from the list. Commenting out this line will reveal expired callouts in composer. 
- The article should be publishable with this error.

Error 3 - The callout no longer exists in the list of available callouts
- This can be tested in prosemirror by forcing a callout into the dropdown that is not available in the campaign list. This can be done by adding the following code.
-  Alternatively, It is more difficult to test in composer as we filter out expired callouts from the list. Commenting out this line will reveal expired callouts in composer. 
- The article should be publishable with this error.

## Images

| none selected     | 
|-------------|
| ![ none-selected ][] | 

[none-selected]: https://user-images.githubusercontent.com/20416599/206424051-f8929456-552b-4697-bc39-33cc72dcf22b.png

| callout expired     | 
|-------------|
| ![ expired ][] | 

[expired]: https://user-images.githubusercontent.com/20416599/206424780-b0acc47e-ce8d-4b3c-87ab-470b84e10ce4.png

| callout unavailable     | 
|-------------|
| ![ unavailable ][] | 

[unavailable]: https://user-images.githubusercontent.com/20416599/206424156-f7533d17-3a86-4ca4-912e-76ccc81b9798.png

## Accessibility

### DIFF
--- a/src/elements/callout/Callout.tsx
+++ b/src/elements/callout/Callout.tsx
@@ -8,7 +8,8 @@ import type { FieldNameToValueMap } from "../../plugin/helpers/fieldView";
 import { dropDownRequired } from "../../plugin/helpers/validation";
 import { createReactElementSpec } from "../../renderers/react/createReactElementSpec";
 import { CustomDropdownView } from "../../renderers/react/customFieldViewComponents/CustomDropdownView";
-import { CalloutError, calloutStyles } from "../embed/Callout";
+import { calloutStyles } from "../embed/Callout";
+import { CalloutError } from "./CalloutError";
 import { CalloutTable } from "./CalloutTable";
 
 export type Fields = {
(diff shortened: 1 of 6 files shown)


src/elements/callout/Callout.tsx
```diff
@@ -8,34 +11,9 @@ import type { FieldNameToValueMap } from "../../plugin/helpers/fieldView";
...
-export type Campaign = {
-  id: string;
-  name: string;
-  fields: Fields;
-  rules: Rules[];
-  priority: number;
-  displayOnSensitive: boolean;
-  activeFrom?: number;
-  activeUntil?: number;
-};
```
abeddow91, line 38: we use these types in a few different places. In order to keep them synced, and to prevent any circular dependencies, i've extracted them into a types file within the callout folder. I couldn't find a current convention for this so let me know if this is not idiomatic to the project.

jonathonherbert (approved):
Tested locally with the three conditions, looks great 👍

PR #291, opened 2023-01-31
feat: add tagId field to callout and remove the applyTag & onRemove handlers 

marjisound:
Paired on this work with @jonathonherbert & @rebecca-thompson 

## What does this change?

This PR adds a new field `tagId` for the Callout element. This fields doesn't exist in view and is only added for the purpose of data. The field will be populated by value when the user chooses a callout from the drop down. 

This PR also removes the `applyTag` & `onRemove` handlers from the `Callout` element. These handlers were being used in composer client to add/remove the relevant callout tag to the furniture. 

The approach of adding/removing the tags on the client was found to be flaky and unreliable, so we decided to take a different approach by updating the tags server side when saving the updated content to database.  The composer PR can be found here https://github.com/guardian/flexible-content/pull/4287

## How to test

This was tested locally with composer.

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -181,8 +181,6 @@ const {
     callout: createCalloutElement({
       fetchCampaignList: () => Promise.resolve(sampleCampaignList),
       targetingUrl: "https://targeting.code.dev-gutools.co.uk/",
-      applyTag: (tag: string) => console.log(`Apply ${tag} tag`),
-      onRemove: (fields) => console.log("Remove callout", fields),
     }),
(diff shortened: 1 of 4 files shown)


marjisound:
[Composer - Add/Remove tags at database level when callout is added/removed](https://trello.com/c/aTl8BGpr/691-composer-add-remove-tags-at-database-level-when-callout-is-added-removed)

rebecca-thompson (approved):
Tested locally and all seems to work well

PR #295, opened 2023-03-17
feat: Make callout fields editable

sophie-macmillan:
Editorial have requested to be able to update a callout's prompt, title and description on an article by article basis. This PR is one of many, which will allow that.

## What does this change?

- This updates the callout embed, so that users can edit the Title and Description.
- Adds a new field, Prompt, which is also editable
- Updates the text by the callout checkbox from `Show as non-collapsable` -> `Tick for stand alone callouts only` based on editorial feedback
- Updates the callout title, so rather than being `CALLOUT: ${title}`, it is now just the callout name

## Images

| Desc  | Image |
| ------------- | ------------- |
| Using default values | image  |
| Overriding prompt with some text  | image  |
| Overriding title in order to hide it | image  |

## Accessibility

- I struggled to test any of the prosemirror elements with voiceover. If someone is free to take a look with me that would be great!

### DIFF
--- a/src/elements/callout/CalloutError.tsx
+++ b/src/elements/callout/CalloutError.tsx
@@ -43,7 +43,7 @@ export const CalloutError = ({
         <p>This callout has expired and will not appear in the article.</p>
       </Error>
       <CalloutTableHeader
-        title={callout.fields.callout}
+        name={callout.name}
         tagName={callout.fields.tagName}
         targetingUrl={targetingUrl}
         calloutId={calloutId}


src/elements/callout/calloutDataTransformer.ts
```diff
@@ -38,14 +59,37 @@ export const transformElementOut: TransformOut<
...
+  useDefaultDescription,
+  overrideDescription,
 }: FieldNameToValueMap<typeof calloutFields>): ExternalCalloutData => {
+  const cleanupDescription = (desc: string): string => {
+    // Empty richtext fields have a paragraph tag, so we should check if its empty
+    const noTags = desc.replace(/(<([^>]+)>)/gi, "");
+    return noTags.trim() ? desc : "";
+  };
+
+  const getOptionalFields = () => ({
```
jonathonherbert, line 75: Any reason not to do this inline? (I'm super happy either way, I'm just curious as to your reasoning, perhaps b/c it makes the intent clearer?)

src/elements/callout/calloutDataTransformer.ts
```diff
@@ -38,14 +59,37 @@ export const transformElementOut: TransformOut<
...
+    // If we are using the default value, don't include the override field
+    // Otherwise, cleanup the override field and include it
+    ...(!useDefaultPrompt && { overridePrompt: overridePrompt.trim() }),
```
jonathonherbert, line 78: I did _not_ know you could do this in objects (although did I know you cannot do this in arrays, hence some confusion 🙃) – TIL, ta!

jonathonherbert (approved):
This looks great. Tested, works as expected.

(Aware this is in draft, but looks g2g from what I can see. Happy to re-review as and when.)

Presume we're adding e.g. `useDefaultPrompt` because we need to model two states (overridden with the given string, even if empty | not overriden), and our text field is always of type `string`. Would permitting an `undefined` value for string fields be useful here, in that you could do away with the additional fields?

(Hypothesis: perhaps not, b/c it's useful to keep the previous value in state, even if it's not currently used, so you can revert to it. I presume this is expected behaviour, e.g. resetting the prompt retains the underlying value but does not use it, and editing the prompt reinstates that value?)

jonathonherbert:
👍  complexity seems fine given that it supports the feature you mention. Totally fair to leave in draft while we're waiting for things to happen downstream. I've seen a 'do-not-merge' tag used for this purpose, too.

PR #297, opened 2023-04-06
feat: Remove unnecessary fields from cartoon form

simonbyford:
## What does this change?

The Cartoon element form has two fields ("Vertical padding" and "Background colour") that were migrated across from the old Comic generator tool, but after speaking with users we're pretty sure we don't need them. This PR removes them. It also moves the "Weighting" field to the row above so it's not sitting on its own. Finally, it renames `cartoonElement` to `createCartoonElement` for the sake of consistency. 

## How to test

Run `prosemirror-elements` locally and click the "Add Cartoon" button. Note the fields are gone.

I've also updated the [testing branch](https://github.com/guardian/flexible-content/compare/bt/cartoon-element) in composer with the new name (`createCartoonElement`). ~~Annoyingly I wasn't actually able to get this working. I think my `yalc` might be broken.~~ (Update: @rebecca-thompson helped me fix this; I was running `yalc` in the parent directory by mistake)

Mateusz looks really happy.

There is no risk as this element is not currently consumed anywhere.

## Images

| Before      | After      |
|-------------|------------|
|  Screenshot 2023-04-06 at 12 00 27 | Screenshot 2023-04-06 at 12 00 17 |

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -11,10 +11,10 @@ import { EditorState } from "prosemirror-state";
 import { EditorView } from "prosemirror-view";
 import {
   buildElementPlugin,
-  cartoonElement,
   codeElement,
   commentElement,
   createCalloutElement,
+  createCartoonElement,
   createContentAtomElement,
   createDemoImageElement,
   createEmbedElement,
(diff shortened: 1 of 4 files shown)


rebecca-thompson (approved):
LGTM! 👍

PR #299, opened 2023-04-24
RFC: pipeable element transformers

jonathonherbert:
## What does this change?

At the moment, we've lots of ad-hoc functions for transforming our external elements into internal elements, and vice-versa. This leads to a fair amount of repetition, as we must define input and output transformers for each element – even when they're doing very similar things! How might we DRY things up a little?

### 1. Pipeable functions that manage both `in` and `out`, producing a transformer.

One way of reducing the boilerplate we're writing, and reducing the likelihood of bugs: implement a set of transformers that we can chain into a single function to process both `in` and `out` operations.

For example, we spend ~50 lines transforming interactives, which might be expressed more concisely as

```ts
const transformer = pipe(
  moveAssetsAndFieldsInline(),
  stringToBool("isMandatory"),
  optional(["alt", "caption")
);

// Which might give us

transformer.in // ExternalElement => Element
transformer.out // Element => ExternalElement
```

We can imagine composing the rest of our transformers in a similar way.

### 2. Pipeable functions useable within `in` and `out` transformers.

Another option would be to write transformer functions as usual, but provide methods for use internally. In this case, the interactives transformer might look like: 

```ts
// I've used the existing `TransformIn` type here, but perhaps we can imagine making something a little neater
export const transformElementIn: TransformIn<
  PartialInteractiveData,
  ReturnType<typeof createInteractiveFields>
> = pipe(
  moveAssetsAndFieldsInline(),
  stringToBool("isMandatory"),
  optional(["alt", "caption")
);

export const transformElementOut: TransformOut<
  ExternalInteractiveData,
  ReturnType<typeof createInteractiveFields>
> =  pipe(
  optional(["alt", "caption")
  boolToString("isMandatory"),
  moveAssetsAndFieldsToElement(),
);
```

This results in less type complexity than 1., but does mean some repetition (`out` will always be the fns composed for `in` in reverse order.)

### Pros and cons

There are some tradeoffs that I think are the same for both 1. and 2., but differ in degree – here's a list of the ones I've thought of:

- ✅ Declarative – it's easier to see what each transformer is doing.
- ✅ More concise.
- ✅ Less likely to ship bugs when creating new transformers with these functions.
- ✅ Relatively low risk to swap, as existing transformers are well tested with fixture data.
- ❌ More complicated to write pipeline functions – the type signatures can be challenging to get right without a good knowledge of mapped types and generics.
- ❌ Debugging when things go wrong is more difficult, as the type errors are long-winded and abstract.

This has been on my mind for a while, but I'm on the fence as to whether it's worth paving this road, or how far we should go. Are devs comfortable writing transformers ad-hoc as they stand? Or did the repetition and boilerplate give them a bit of an itch?

To give an idea of the complexity, the PR contains an example implementation of a few pipeline functions, but does not yet include the code that would compose pipeline functions into the object `{ in, out }` in a single pass. 

(The `pipe` function is stolen from Dotcom 😀 )

### DIFF
--- a/src/elements/helpers/__tests__/transformHelpers.spec.ts
+++ b/src/elements/helpers/__tests__/transformHelpers.spec.ts
@@ -1,0 +1,96 @@
+import type { Asset } from "../defaultTransform";
+import {
+  moveAssetsAndFieldsInline,
+  pipe,
+  stringToBool,
+} from "../transformHelpers";
+
+describe("moveAssetsAndFieldsInline", () => {
+  const { in: inFn, out: outFn } = moveAssetsAndFieldsInline<{
+    assets: Asset[];
+    fields: { something: string };
+  }>();
+  it("should move assets and fields inline - in", () => {
+    const result = inFn({
+      assets: [],
+      fields: { something: "hai" },
+    });
+
+    expect(result).toEqual({ assets: [], something: "hai" });
+  });
+
+  it("should move assets and fields inline - out", () => {
+    const result = outFn({ assets: [], something: "hai" });
+
+    expect(result).toEqual({ assets: [], fields: { something: "hai" } });
+  });
+});
+
+describe("stringToBool", () => {
+  const { in: inFn, out: outFn } = stringToBool("isMandatory");
+
+  it("should transform field properties to boolean values - in", () => {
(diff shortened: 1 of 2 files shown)


jonathonherbert:
Closing as this code is now in our CMS.

PR #302, opened 2023-06-21
feat: Hide the title for standalone callouts as this will never be shown

sophie-macmillan:
## What does this change?

This updates the callout element, so that the callout title is not editable when the standalone checkbox is checked.
This is because the title is always hidden in the front end, and is confusing to let editorial update something that won't be shown.

Also updates stand alone -> standalone

## Images

| Change | Before | After |
|--------|--------|--------|
| Standalone callout (title hidden) | image | image|
| Not standalone callout (text change only) | image | image | 

## Accessibility

- [x] [Tested with screen reader](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#screen-readers)

### DIFF
--- a/src/elements/callout/CalloutTable.tsx
+++ b/src/elements/callout/CalloutTable.tsx
@@ -70,6 +70,9 @@ const bodyStyle = css`
 const strongStyle = css`
   font-weight: 700;
 `;
+const italicStyle = css`
+  font-style: italic;
+`;
 
 const rightHeaderStyle = css`
   float: right;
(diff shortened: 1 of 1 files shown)


src/elements/callout/CalloutTable.tsx
```diff
@@ -202,41 +205,48 @@ export const CalloutTable = ({
             }
           />
         )}
-
-        {useDefaultTitle.value ? (
+        {isNonCollapsible.value && (
```
rebecca-thompson, line 208: Non-blocking: this could be refactored into a ternary operator which might make the logic easier to follow.

Also non-blocking: it would be nice to have some comments explaining the props. `isNonCollapsible` was not immediately apparent to me.

rebecca-thompson (approved):
A couple of optional refactors, but otherwise it looks good and works as expected

jonathonherbert:
Closing as #313 moves the relevant code to our CMS, and I hadn't spotted this change! Sorry about that. Happy to help lift and shift this code, if still needed, to the appropriate place.

PR #303, opened 2023-06-27
feat: Stop adding the thumbnail role to all image elements

rebecca-thompson:
## What does this change?
The image element currently derives its weighting drop down options (also referred to as 'role' in some places) from two places:

1. hard-coded values in the React component (thumbnail)
2. values that are passed in when the element is created - such as [here](https://github.com/guardian/flexible-content/blob/main/composer/src/js/prosemirror-setup/helpers/image-element.ts)

Up to this point, having the thumbnail weighting automatically added hasn't been a problem as all image elements allow for an image to be a thumbnail, however... the Picture content editor does not currently permit this weighting, so we need to move thumbnail from a hard-coded value and start passing it in along with the other weighting options.

## How to test
This has been tested locally by running `yarn yalc` and importing the local build into flexible-content. You should see the thumbnail option continue to appear in the image element in the article editor.

## Images
Screenshot 2023-06-27 at 17 46 10

### DIFF
--- a/src/elements/image/ImageElementForm.tsx
+++ b/src/elements/image/ImageElementForm.tsx
@@ -158,8 +158,6 @@ export const createImageElement = (options: ImageElementOptions) => {
   return { element, updateAdditionalRoleOptions };
 };
 
-const thumbnailOnlyOptions = [thumbnailOption];
-
 const RoleOptionsDropdown = ({
   field,
   mainImage,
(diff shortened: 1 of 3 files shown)


src/elements/image/ImageElementForm.tsx
```diff
@@ -171,13 +169,15 @@ const RoleOptionsDropdown = ({
 }) => {
   // We memoise these options to ensure that this array does not change identity
   // and re-trigger useEffect unless our additionalRoleOptions have changed.
-  const allOptions = useMemo(
-    () => [...additionalRoleOptions, thumbnailOption],
-    [additionalRoleOptions]
-  );
+  const allOptions = useMemo(() => additionalRoleOptions, [
+    additionalRoleOptions,
```
jonathonherbert, line 173: Fair to say that `additionalRoleOptions` is now just `roleOptions`? Might be worth a rename.

jonathonherbert (approved):
Looks great! One non-blocking comment.

I hope you don't mind – in testing I noticed that the thumbnail option hadn't been added to the demo, which makes it hard to test the behaviour w/ small crops locally. Added in eb0c1358d567fb7dd3522ec5d8e7527f886429a2 as part of testing.

src/elements/image/ImageElementForm.tsx
rebecca-thompson: I started doing this, but `roleOptions` is also used as a local variable in the `RoleOptionsDropdown` to define the final role options actually used once adjustments have been made, so it seemed to make more sense to keep it as is.

PR #305, opened 2023-06-28
fix: Simplify-actions-node-version

Mark-McCracken:
There's a simpler way to setup node with the version specified in the .nvmrc file

## What does this change?
Rather than cat'ing a file and using the output of steps, you can just list the file in the setup-node action
Reduces the number of steps, and makes the intention clearer in the workflow file


rebecca-thompson (approved):
I can see this option in the usage documentation for `actions/setup-node@v3` - https://github.com/actions/setup-node#usage

Thanks for the refactor!

PR #307, opened 2023-07-04
fix: model cartoon element booleans as booleans, not strings

rebecca-thompson:
## What does this change?
This changes the `displayCredit` boolean values in the cartoon data transformer to be correctly modelled as a boolean, rather than a string.

It also moves the definition of the `Image` type to be co-located with the other Media type definitions and exports this type, so we can use it in flexible-content (important for getting re-crop and add image buttons working).

## How to test
This has been checked by building the libs locally using `yalc` and pulling them into the flexible-content project.


simonbyford (approved):
LGTM. Co-location makes sense too 👍

PR #311, opened 2023-08-15
fix: formatting of cross to remove images in cartoon element

rebecca-thompson:
## What does this change?
Adds the remove image cross after the image rather than before. This prevents images of certain dimensions from masking the cross.

| Before      | After      |
| ----------- | ---------- |
| Screenshot 2023-08-15 at 09 50 46 | Screenshot 2023-08-15 at 09 50 28 |

## How to test
You should be able to run the demo and add/remove images from the cartoon element with no trouble.


simonbyford (approved):
Nice

PR #313, opened 2023-10-19
feat: Export element helpers, but not elements

jonathonherbert:
## What does this change?

Exports the necessary helpers to create elements with `prosemirror-elements` – but not the library elements themselves.

We think it's better to define these closer to our CMS to improve the feedback loop when we're iterating on element structure and UI changes.

What we do with the elements as they stand in the repo can be addressed in a follow-up PR. I'd be tempted to keep a few definitions around as they're useful in the sandbox – but we might clearly label them to make it clear they're not used in production. (Not that they could be, as they are no longer exported?)

## How to test

- Take a look at the new exports in [index.ts](https://github.com/guardian/prosemirror-elements/blob/f095a75916f5a5cb153c3b20de14674c3012fafb/demo/index.ts) – all make sense?
- Take a look at the [companion PR](https://github.com/guardian/flexible-content/pull/4410) in Composer. Following the instructions for element testing in local applications in the README, import a local build of this branch into Composer's branch (`jsh/elements-in-composer`). Does the application work as intended?

Happy to pair on the above!

### DIFF
--- a/cypress/helpers/editor.ts
+++ b/cypress/helpers/editor.ts
@@ -6,7 +6,7 @@ import {
   trimHtml,
 } from "../../src/plugin/helpers/test";
 import { elementWrapperTestId } from "../../src/renderers/react/ElementWrapper";
-import { getFieldViewTestId } from "../../src/renderers/react/FieldView";
+import { getFieldViewTestId } from "../../src/renderers/react/FieldComponent";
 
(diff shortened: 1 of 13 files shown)


rebecca-thompson (approved):
I like this approach. It will make developing elements much quicker and easier in Composer 🎉

PR #315, opened 2023-11-01
Use changesets for releases

joecowton1:
## What does this change?
Use Changesets rather than semantic-release for npm publishing.

## How to test
Merge to main and check build and release workflow works.

## Context

As discussed at this week's Client Side Meetup, Dev X are in the process of consolidating ownership of the Guardian NPM account by reducing the number of engineers with permission to use it.

With this in mind we’d like to revoke access to the Guardian NPM account to all but a small number of superusers on 1st December 2023.
 
Our current recommendation is that we use [Changesets](https://github.com/changesets/changesets) to release NPM packages, as this uses a single org level NPM token.


jonathonherbert (approved):
PR looks good, thanks for the detailed description 🙏

PR #318, opened 2023-11-16
feat: Export repeater field

jonathonherbert:
## What does this change?

Exports `createRepeaterField`, which should really have been in #313. 

## How to test

You should be able to use `createRepeaterField` in consuming projects.


tonytw1 (approved):
This is an export of an existing thing.
The name looks unique enough that we're not concerned about it clashing with anything.

PR #324, opened 2023-12-18
Add NestedElementField

rhystmills:
Co-authored with @davidfurey and @jonathonherbert - first few commits were David's but I unintentionally rebased his name out of the commit history.

image

## What does this change?
This PR adds the nested element field. This field is a rich text editor that is itself able to contain elements. This will enable us to build larger, 'structural' elements within Composer that can contain elements like images and videos.

There is some further work to do in future PRs:
- Add some helper functions to provide a content expression for a nested element field that disallows certain elements (for example, we probably don't want elements that contain nested element fields within other nested element fields).
- There is a known bug in repeater fields involving applying Decorations . Trying to insert an element (e.g. via pasting) into any node other than the initial one in a repeater field now works, but the Decorations for notes and Typerighter matches within nestedElements aren't currently working.
- Make sure that the element controls are usable within the field. (e.g. move, delete, select)

This PR could do with some tests for nested element fields - I'll add some when I'm back from holiday.

## How to test
1. Run the demo app for this branch with `yarn start`.
2. Add the 'Alt Style' element.
3. Can you past other elements into the 'Content' field? (You should be able to).

You may also want to test this PR in Composer via yalc with: https://github.com/guardian/flexible-content/pull/4492

### DIFF
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -47,6 +47,7 @@ import {
   sideEffectPlugin,
 } from "./helpers";
 import {
+  sampleAltStylesElement,
   sampleAudio,
   sampleCallout,
   sampleCampaignCalloutList,
@@ -432,7 +433,11 @@ const createEditor = (server: CollabServer) => {
       values: sampleInteractiveAtom,
     },
     { label: "Comment", name: commentElementName, values: sampleComment },
-    { label: "Alt Style", name: altStyleElementName, values: altStyleElement },
+    {
+      label: "Alt Style",
+      name: altStyleElementName,
+      values: sampleAltStylesElement,
+    },
   ] as const;
 
   buttonData.map(({ label, name, values }) =>
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -250,7 +250,7 @@ const {
         type,
         value: true,
         tags,
-      }),
+    })
   }
 );
 
@@ -261,8 +261,19 @@ const strike: MarkSpec = {
   },
 };
 
+const docNodeSpec = {
+  doc: {
+    /*
+    * We must allow 2 text elements to exists side by side
+    * so that when we delete an element we can then join it.
+    */
+    content: "(element|block)*",
+    toDOM: () => ["div", 0],
+  }
+}
+
 const schema = new Schema({
-  nodes: basicSchema.spec.nodes.append(nodeSpec),
+  nodes: basicSchema.spec.nodes.append(docNodeSpec).append(nodeSpec),
   marks: { ...omit(marks, "code"), strike },
 });
 
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -250,7 +250,7 @@ const {
         type,
         value: true,
         tags,
-    })
+      }),
   }
 );
 
@@ -264,13 +264,13 @@ const strike: MarkSpec = {
 const docNodeSpec = {
   doc: {
     /*
-    * We must allow 2 text elements to exists side by side
-    * so that when we delete an element we can then join it.
-    */
+     * We must allow 2 text elements to exists side by side
+     * so that when we delete an element we can then join it.
+     */
     content: "(element|block)*",
     toDOM: (): [string, ...any[]] => ["div", 0],
-  }
-}
+  },
+};
 
 const schema = new Schema({
   nodes: basicSchema.spec.nodes.append(docNodeSpec).append(nodeSpec),
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -267,7 +267,7 @@ const docNodeSpec = {
      * We must allow 2 text elements to exists side by side
      * so that when we delete an element we can then join it.
      */
-    content: "(element|block)*",
+    content: "block*",
     toDOM: (): [string, ...any[]] => ["div", 0],
   },
 };
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -268,7 +268,7 @@ const docNodeSpec = {
     * so that when we delete an element we can then join it.
     */
     content: "(element|block)*",
-    toDOM: () => ["div", 0],
+    toDOM: (): [string, ...any[]] => ["div", 0],
   }
 }
 
--- a/demo/index.ts
+++ b/demo/index.ts
@@ -263,10 +263,6 @@ const strike: MarkSpec = {
 
 const docNodeSpec = {
   doc: {
-    /*
-     * We must allow 2 text elements to exists side by side
-     * so that when we delete an element we can then join it.
-     */
     content: "block*",
     toDOM: (): [string, ...any[]] => ["div", 0],
   },
--- a/src/editorial-source-components/Editor.ts
+++ b/src/editorial-source-components/Editor.ts
@@ -28,7 +28,7 @@ export const Editor = styled.div<{ hasValidationErrors: boolean, useAlternateSty
     padding: 10px;
     &:focus-within {
       ${focusHalo}
-      outline: 1px solid black
+      border: 1px solid ${background.inputChecked};
     }
   }
 `;
--- a/src/editorial-source-components/Editor.ts
+++ b/src/editorial-source-components/Editor.ts
@@ -14,7 +14,7 @@ export const Editor = styled.div<{
   }
   .ProseMirrorElements__RichTextField,
   .ProseMirrorElements__TextField,
-  .ProseMirrorElements__NestedField {
+  .ProseMirrorElements__NestedElementField {
     ${inputBorder}
     &:active {
       border: 1px solid ${background.inputChecked};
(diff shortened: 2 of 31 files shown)


davidfurey:
As discussed before, I'm unsure on the best name for this. Perhaps `ElementsField`?

rhystmills:
> As discussed before, I'm unsure on the best name for this. Perhaps `ElementsField`?

I went with `nestedElement` rather than just `nested` in the end - going for just 'element' or 'elements' felt confusing in some places in the codebase. Does that seem alright to you?

src/editorial-source-components/Editor.ts
```diff
@@ -4,18 +4,35 @@ import { focusHalo } from "@guardian/src-foundations/accessibility";
...
+    ${({ hasValidationErrors }) =>
+      !!hasValidationErrors && `border-color: ${border.error};`}
+  }
+  .ProseMirrorElements__NestedElementField .ProseMirror-focused {
+    outline: none;
+  }
+  .ProseMirrorElements__NestedElementField {
+    padding: 10px;
+    &:focus-within {
+      ${focusHalo}
```
Fweddi, line 34:
```suggestion
      ${focusHalo}
```

src/plugin/fieldViews/NestedFieldView.ts
```diff
@@ -0,0 +1,102 @@
...
+    { placeholder, isResizeable }: NestedFieldDescription,
+    disallowedPlugins: PluginKey[] = []
+  ) {
+    super(
+      node,
+      outerView,
+      getPos,
+      offset,
+      decorations,
+      // Allow plugins without a plugin key, but exclude those that are explicitly blocked
```
Fweddi:
```suggestion
Disallow plugins which have a plugin key, or are explicitly blocked.
```

src/plugin/fieldViews/NestedElementFieldView.ts
```diff
@@ -0,0 +1,119 @@
...
+    super(
+      node,
+      outerView,
+      getPos,
+      offset,
+      decorations,
+      // Allow plugins without a plugin key, but exclude those that are explicitly blocked
+      outerView.state.plugins.filter(
+        (plugin) =>
+          !plugin.spec.key || !disallowedPlugins.includes(plugin.spec.key)
```
Fweddi, line 96:
```suggestion
Disallow plugins which have a plugin key, or are explicitly blocked.
```

src/plugin/fieldViews/NestedElementFieldView.ts
```diff
@@ -0,0 +1,119 @@
...
+    { placeholder, isResizeable }: NestedElementFieldDescription,
+    disallowedPlugins: PluginKey[] = []
+  ) {
+    super(
+      node,
+      outerView,
+      getPos,
+      offset,
+      decorations,
+      // Allow plugins without a plugin key, but exclude those that are explicitly blocked
```
Fweddi, line 93:
```suggestion
Disallow plugins which have a plugin key, or are explicitly blocked.
```

Fweddi: I think this is clearer, but not sure, feel free to ignore.

demo/index.ts
```diff
@@ -261,8 +261,19 @@ const strike: MarkSpec = {
   },
 };
 
+const docNodeSpec = {
+  doc: {
+    /*
+     * We must allow 2 text elements to exists side by side
+     * so that when we delete an element we can then join it.
+     */
```
Fweddi, line 269: Not entirely sure I understand this. Is this just a general bug that's being fixed here, or is it related to the nested elements?

Also, the comment here mentions two text elements but the content field is talking in terms of blocks, just thought I would flag in case that's unintended.

src/plugin/__tests__/nodeSpec.spec.ts
```diff
@@ -261,5 +262,35 @@ describe("nodeSpec generation", () => {
...
+        ).toMatchObject({
+          content: "element+",
+        });
+      });
+
+      it("should create a NodeSpec for the nestedElement using a value defined by the user if specified", () => {
+        const nodeSpec = getNodeSpecForField(
+          "exampleElement",
+          "exampleNestedElementField",
+          createNestedElementField({ content: "element*" })
```
Fweddi, line 285: Elsewhere this is defined with content as `block*`, is this an inconsistency?

I would have thought that a nested element would contain `element*`, but I suppose the nested elements are also nested in blocks?

I've said nested too many times... 🪆

src/plugin/field.ts
```diff
@@ -55,6 +55,8 @@ export const getFieldsFromNode = <FDesc extends FieldDescriptions<string>>({
     const fieldDescription = fieldDescriptions[fieldName];
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- strictly, we should check.
     if (!fieldDescription) {
+      console.log({ fieldNode, fieldDescriptions, fieldName });
```
Fweddi, line 58: Might not be needed?

src/plugin/helpers/__tests__/fixtures.ts
```diff
@@ -2,13 +2,19 @@ import { Schema } from "prosemirror-model";
...
+import { createNestedElementField } from "../../fieldViews/NestedElementFieldView";
 import { createRepeaterField } from "../../fieldViews/RepeaterFieldView";
 import { createTextField } from "../../fieldViews/TextFieldView";
 import { RepeaterFieldMapIDKey } from "../constants";
 import { createEditorWithElements, createNoopElement } from "../test";
 import { maxLength, required } from "../validation";
 
 export const elements = {
+  nestedElement: createNoopElement({
+    content: createTextField({
```
Fweddi, line 14: Interested in this as it seems to be doing something similar but different to the nestedElementField below. Also it's called nestedElement, but seems to be a text no-op element with a text field?

Fweddi:
I can't copy / cut an element that is nested - I can select it, but the copy and cut functions don't seem to work. I don't know if this is related to the repeater fields bug, or is something separate.

Fweddi:
I also can't reorder nested elements using the single arrows (I can using the double arrows).

Fweddi:
This might be an issue with the demo tool, but if I delete all the text in a Nested Element Field, and there's another element nested, I can't edit the field (or can't click into it to type / add elements).

Fweddi:
> I also can't reorder nested elements using the single arrows (I can using the double arrows).

I see you've mentioned this as work for a follow-up PR. 👍

src/plugin/helpers/__tests__/fixtures.ts
rhystmills: Unclear naming on my part I think - this is essentially a test element, separate from the 'example' element used in the tests, to put in the `nestedElementField` in the tests.

I think `exampleElementToNest` might be better

demo/index.ts
rhystmills: Whoops - I think I unintentionally committed that after getting the node spec from Composer's editor.

Definitely not relevant to the PR!

src/plugin/__tests__/nodeSpec.spec.ts
rhystmills: 🪆

The top level `block*` content expression is specific to the demo in `prosemirror-elements` - `element+`/`element*` is closer to what we'll see in flexible-content where the top level editor has an `element+` content expression. At the moment there's a bit of a compromise to allow Composer and the Demo to both work.

It would be good to align the demo more closely with what we see in flexible-content so we don't have this discrepancy. I could add that as a follow-up card to follow this one, or we could fix in this PR?

The demo could do with some love in general, e.g. having far fewer elements.

src/plugin/field.ts
rhystmills: Tis gone.

src/plugin/fieldViews/NestedElementFieldView.ts
rhystmills: Currently more plugins exist without keys than with keys in Composer, and they are likely to be doing important work. That's why I've gone with a blocklist rather than an allowlist.

src/plugin/__tests__/nodeSpec.spec.ts
Fweddi: That makes sense and sounds good to me. The difference between demo and Composer does confuse me from time to time (the Alt Styles element in this PR for example threw me).

src/plugin/fieldViews/NestedElementFieldView.ts
Fweddi: @rhystmills and I paired on this briefly to clarify the code (the TS checker was confusing me).

src/plugin/fieldViews/NestedElementFieldView.ts
```diff
@@ -0,0 +1,123 @@
...
+  ) {
+    super(
+      node,
+      outerView,
+      getPos,
+      offset,
+      decorations,
+      // We should expect to encounter plugins without keys.
+      // We use a disallow list (rather than an allow list) because we expect most plugins to work,
+      // and only want to disallow those that cause problems.
```
Fweddi, line 96: We'll need to remove this comment now we've changed to an allowlist.

Fweddi (approved):
Subsequent nodes of the repeater field now work as expected 👍 

There are still some small issues around decorations but I think this puts us in a good place to merge.

PR #330, opened 2024-01-18
Add minimum rows option for nestedElementField

rhystmills:
## What does this change?
This adds an option to specify the minimum rows for a `nestedElementField`.  This will allow us to hint to Composer users that a field is intended for multiline text. `minRows` for the `nestedElementField` applies a `min-height` style to the relevant field using a calculation based on the browser's `line-height`.

I've added this to the `NestedElementField` rather than the `ProseMirrorFieldView` it extends because we haven't got a use case for `minRows` in the `RichTextField` yet (which also extends `ProseMirrorFieldView`) - additionally, the `TextField` (which already supports rows) also includes `maxRows` and `isMultiline` which I was reluctant to add as options for all text fields. I've avoided these questions by just adding `minRows` to the place where we definitely want it.

## How to test
1. Run the `prosemirror-elements` demo (https://prosemirror-elements.local.dev-gutools.co.uk/) locally, by running `yarn start`.
2. Add the 'Alt Style' element (one of the many buttons in the demo)
3. Does the content field appear to have a min-height of around six rows?
4. Does it expand to fit the content you type if you go beyond six rows? (It should)

### DIFF
--- a/src/plugin/fieldViews/NestedElementFieldView.ts
+++ b/src/plugin/fieldViews/NestedElementFieldView.ts
@@ -90,7 +90,6 @@ export class NestedElementFieldView extends ProseMirrorFieldView {
     { placeholder, isResizeable, minRows }: NestedElementFieldDescription,
     // Specify plugins of which the field should have its own copy
     allowedPlugins: PluginKey[] = []
-    // Initial height of the field, in rows
   ) {
     super(
       node,
--- a/src/plugin/fieldViews/NestedElementFieldView.ts
+++ b/src/plugin/fieldViews/NestedElementFieldView.ts
@@ -4,10 +4,10 @@ import type { DecorationSource, EditorView } from "prosemirror-view";
 import type { FieldValidator } from "../elementSpec";
 import { pluginKey } from "../helpers/constants";
 import type { PlaceholderOption } from "../helpers/placeholder";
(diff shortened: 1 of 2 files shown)


src/plugin/fieldViews/NestedElementFieldView.ts
```diff
@@ -82,9 +87,10 @@ export class NestedElementFieldView extends ProseMirrorFieldView {
     offset: number,
     // The initial decorations for the FieldView.
     decorations: DecorationSource,
-    { placeholder, isResizeable }: NestedElementFieldDescription,
+    { placeholder, isResizeable, minRows }: NestedElementFieldDescription,
     // Specify plugins of which the field should have its own copy
     allowedPlugins: PluginKey[] = []
+    // Initial height of the field, in rows
```
Georges-GNM, line 93: Not sure if this comment should be here?

Georges-GNM (approved):
Looks good to me, confirmed it was working locally 👍

src/plugin/fieldViews/NestedElementFieldView.ts
rhystmills: Definitely not, thanks for spotting!

rhystmills, line 93:
```suggestion
```

PR #331, opened 2024-01-19
Support Typerighter in repeater nodes by default

rhystmills:
## What does this change?
This adds typerighter support in repeater nodes by default. This means that the key takeaways element will support Typerighter.

image

## How to test
1. Publish this branch locally with `yalc publish`
2. Use in the composer directory of flexible-content with `yalc add @guardian/prosemirror-elements`
3. Activate the key takeaways feature switch
4. Add some incorrect text in the content field of the key takeaways element
5. Run Typerighter (you may need to visit a CODE domain to be authorised)
6. Are the mistakes highlighted by Typerighter?

### DIFF
--- a/src/plugin/nodeSpec.ts
+++ b/src/plugin/nodeSpec.ts
@@ -1,6 +1,7 @@
 import OrderedMap from "orderedmap";
 import type { Node, NodeSpec, NodeType, Schema } from "prosemirror-model";
 import { DOMParser } from "prosemirror-model";
+import { useTyperighterAttrs } from "../elements/helpers/typerighter";
 import { FieldContentType } from "./fieldViews/FieldView";
 import type { RepeaterFieldDescription } from "./fieldViews/RepeaterFieldView";
 import {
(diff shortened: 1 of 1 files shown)


twrichards (approved):
code change LGTM - great detective work finding the solution here.

tested locally and seems to work (good testing instructions btw)

jonathonherbert:
As @twrichards says, great find! Non-blocking: suspect there will be many cases where we repeat nodes that we don't want Typerighter to see, e.g. parts of [the Recipe element](https://github.com/guardian/flexible-content/blob/6d155949005adfa0c3622597c135107ce0fd62ca/composer/src/js/prosemirror-setup/elements/recipe/RecipeElementSpec.tsx#L106) – we might end up highlighting odd input text or amounts etc. as typos.

If that is the case, one solution might be to pass attrs to [createRepeaterField](https://github.com/guardian/prosemirror-elements/blob/26d04205d4f03654c8375f15aecdc8c90d269441/src/plugin/fieldViews/RepeaterFieldView.ts#L18) as `attrs`, like [createTextField](https://github.com/guardian/prosemirror-elements/blob/26d04205d4f03654c8375f15aecdc8c90d269441/src/plugin/fieldViews/TextFieldView.ts#L52). Also has the benefit of keeping Typerighter's specifics in userland, away from core plugin code.

Aware that in this case we'd either need to duplicate attributes across parent and children, or enable separate attributes for each node. Former arguably preferable until we've a use case that requires separate attrs?

PR #332, opened 2024-01-22
Fix typerighter elements bug, get decorations working better in repeater fields containing nested fields

rhystmills:
This PR makes sure more than one typerighter decoration is displayed in elements (currently only one shows in Composer).

image

It also gets decorations working better in `nestedElement` fields contained in `repeater` fields. Decorations will work in `nestedElement` fields in `repeater` fields until an element is added to one of the `nestedElement` fields. After an element is added to one of the `nestedElement` fields, on refresh - decorations won't work in any field in the repeater **_preceding_** the field that contains the element (on page refresh). They **_do_** work in the element that contains the element. 

The easiest way to see this is by using this branch in Composer, adding a note and some text that will match something in Typerighter. For example, "typerighter" is an amend match and "Ava Gardner" is an OK match.

| Before adding the element (all decos work) | After adding element and refreshing (decos don't work in repeater nodes before the one containing the element) |
| --- | --- |
|  image | image |

It also solves the double placeholder bug, which was caused by the `NestedElementField` receiving a `prosemirror-elements` specific placeholder plugin, in addition to the placeholder plugin in the parent editor in Composer. Now, only the `TextField` and `RichTextField` provide their own placeholder plugin. This plugin is designed for fields represent text, whereas the one from Composer is specifically designed to display in place of an empty `TextElement`

## Why is bug happening?
`applyDecorationsToOuterEditor` in the `ProsemirrorField` throws an error when we add element to a repeated `nestedElement` field, and perhaps some decorations are not being applied when this happens. I will continue to look into this issue, but wanted to get this PR out because it will solve the production bug with Typerighter in elements.

## How to test
1. Use this branch in Composer locally using `yalc`.
2. Add a key takeaways element using the feature switch.
3. Try Typerighter and notes in the fields.
4. Try multiple typerighter fields in the pullquote. Do all the matches get decorated?


twrichards (approved):
Code changes LGTM, glad to see some of the stuff we added previously when trying to resolve repeater bug(s) is going here and this change is a net reduction in CODE 🎉 

I've tested locally and its looking pretty good 
image

PR #333, opened 2024-01-22
Serialise elements within nestedElements equivalently to top level element.

rhystmills:
## What does this change?
This PR serialises the nestedElement field as an array of 'elements', when we call the function produced by `createGetElementDataFromNode` in the consumer of `prosesmirror-elements` (i.e. Composer).

If you use this branch in `flexible-content` locally via `yalc`, `flexible-content` should fail to save key takeaways articles, because the shape of what the backend (`apiv2`?) receives will not be what it expects. I imagine saves will fail there until we do the work in `apiv2` to handle the new data structure properly, which will involved representing the contents of the keyTakeaways `nestedElement` field as a `List[Element]` or something similar.

This PR should provide the functionality required to do that work, but it doesn't add any tests, which we'll want to do before merging.

## How to test

TBD


rhystmills:
Using this in `flexible-content`, it looks like the client is sending most `elements` to the backend, but not `textElements`. Will investigate.

src/plugin/field.ts
```diff
@@ -184,24 +226,28 @@ export const updateFieldsFromNode = <FDesc extends FieldDescriptions<string>>({
...
                 offset: accumulatedOffset,
                 innerDecos,
+                getElementDataFromNode,
+                transformElementOut,
               });
 
         const newFieldsForChild = updateFieldsFromNode({
           node: childNode,
-          fields,
+          fields: fields as FieldNameToField<FDesc>,
```
twrichards, line 243: why is a cast needed here?

src/plugin/nodeSpec.ts
```diff
@@ -365,21 +374,82 @@ export const createNodesForFieldValues = <
...
-        return [
-          field.type === "nestedElement"
-            ? createContentNodeFromRichText(
-                schema,
-                content,
-                nodeType.create({ type: field.type })
-              )
-            : createContentNodeFromText(content, field, nodeType),
-        ];
+        const content = fieldValue as unknown[];
```
twrichards, line 377: why is a cast needed here?

src/plugin/nodeSpec.ts
```diff
@@ -365,21 +374,82 @@ export const createNodesForFieldValues = <
...
+          content,
+          field,
+          nodeType,
+          schema,
+          getNodeFromElementData,
+          transformElementIn
+        );
+
+        if (!node) {
+          console.log({ content, field, nodeType, nodeName });
```
twrichards, line 388: did you mean to leave this in? if so, in that it throws after, perhaps...
```suggestion
          console.error({ content, field, nodeType, nodeName });
```

src/plugin/nodeSpec.ts
```diff
@@ -365,21 +374,82 @@ export const createNodesForFieldValues = <
...
+  elementsArray: unknown[],
+  fieldDesc: FieldDescription,
+  nestedElementFieldNodeType: NodeType,
+  schema: Schema,
+  getNodeFromElementData: GetNodeFromElementData,
+  transformElementIn?: TransformElementIn
+): Node | null | undefined => {
+  const childNodes = elementsArray
+    .map((element) => {
+      const externalElement = element as ExternalElementData;
```
twrichards, line 410: is this cast safe? would an `is` predicate function be better?

twrichards (approved):
code LGTM, important change, nice work 🙌  

I've left some questions about casting, hopefully they can be eliminated.

I've not tested locally as I'm in the midst of repeater bug debugging, perhaps worth someone doing that (can be someone in the pair/mob who worked on this)

src/plugin/__tests__/element.spec.ts
```diff
@@ -402,13 +402,15 @@ describe("buildElementPlugin", () => {
...
-          `),
+          nestedField1: [
+            {
+              elementType: "testElement",
+              fields: {
+                richTextField1: "<p>Content 1</p>",
+              },
+              assets: [],
+            },
+          ],
```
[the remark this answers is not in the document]
simonbyford: +1

simonbyford (approved):
This looks good and I've been able to verify everything works locally!

Apologies for the accidental `git push` 🤦

rhystmills:
Noticed that a branch based on this one was having a save failure with callout elements - going to try adding all the elements to the doc and see how they fare, and get to the bottom of the callout element problem. I suspect it's not being transformed properly.

src/plugin/field.ts
rhystmills: (We talked through the type casts off Github and decided to work on them in a follow-up PR)

PR #334, opened 2024-01-24
Add more control over repeater child nodes

Fweddi:
Paired on with @twrichards 

## What does this change?
Previously we were able to add child nodes at the end of the repeater, and remove child nodes at a given index.

This PR allows us to add child nodes at a given index, and move child nodes up or down an index. It also refactors the remove method to have greater guards against invalid invocation (in line with the new methods), and renames the methods to be more precise.

## How to test
You can play about with these controls by creating an Alt Style element in the demo:
image

Check that the controls work as expected, and are enabled / disabled in the right places.

There are also new tests for these methods which can be inspected and run to give confidence in the PR's intent.

### DIFF
--- a/cypress/tests/ImageElement.cy.ts
+++ b/cypress/tests/ImageElement.cy.ts
@@ -597,19 +597,6 @@ describe("ImageElement", () => {
           })
         );
       });
-      it("should add nested repeater elements", () => {
-        addImageElement();
-        clickButton(AddRepeaterButtonId);
-        clickButton(AddNestedRepeaterButtonId);
-        getElementRichTextField("nestedRepeaterText").should("exist");
-      });
-      it("should remove nested repeater elements", () => {
-        addImageElement();
-        clickButton(AddRepeaterButtonId);
-        clickButton(AddNestedRepeaterButtonId);
-        clickButton(RemoveNestedRepeaterButtonId);
-        getElementRichTextField("nestedRepeaterText").should("not.exist");
-      });
       it("should accept values in nested repeater elements", () => {
         addImageElement();
         clickButton(AddRepeaterButtonId);
--- a/cypress/tests/ImageElement.cy.ts
+++ b/cypress/tests/ImageElement.cy.ts
@@ -578,15 +578,27 @@ describe("ImageElement", () => {
 
     describe("Repeater behaviour", () => {
       it("should add repeater elements", () => {
-        addImageElement();
-        clickButton(AddRepeaterButtonId);
-        getElementRichTextField("repeaterText").should("exist");
+        addImageElement({
+          repeater: [
+            {
+              repeaterText: "Example repeater text 1",
+            },
+          ],
+        });
+        getElementRichTextField("repeaterText").its("length").should("eq", 1);
+        const addRepeaterButtons = getButton(AddRepeaterButtonId);
+        addRepeaterButtons.first().click();
+        getElementRichTextField("repeaterText").its("length").should("eq", 2);
       });
       it("should remove repeater elements", () => {
-        addImageElement();
-        const addRepeaterButtons = getButton(AddRepeaterButtonId);
-        addRepeaterButtons.click();
-        addRepeaterButtons.first().click();
+        addImageElement({
+          repeater: [
+            {
+              repeaterText: "Example repeater text 1",
+            },
+            { repeaterText: "Example repeater text 2" },
+          ],
+        });
         getElementRichTextField("repeaterText").its("length").should("eq", 2);
         getButton(RemoveRepeaterButtonId).first().click();
         getElementRichTextField("repeaterText").its("length").should("eq", 1);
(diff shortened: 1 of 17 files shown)


src/plugin/fieldViews/RepeaterFieldView.ts
```diff
@@ -85,28 +115,100 @@ export class RepeaterFieldView extends FieldView<unknown> {
       );
       return;
     }
-    tr.replaceWith(endOfRepeaterNode, endOfRepeaterNode, newNode);
+    let positionToAddFrom;
+    if (maybeIndex === undefined) {
+      // If no index, add to end of repeater node
+      positionToAddFrom = this.getPos() + this.offset + this.node.nodeSize;
+    } else {
```
Fweddi, line 122: We still add child nodes at the end of the repeater when no index is supplied - we could deprecate this function.

cypress/tests/ImageElement.cy.ts
```diff
@@ -597,19 +597,6 @@ describe("ImageElement", () => {
...
-        clickButton(AddNestedRepeaterButtonId);
-        getElementRichTextField("nestedRepeaterText").should("exist");
-      });
-      it("should remove nested repeater elements", () => {
-        addImageElement();
-        clickButton(AddRepeaterButtonId);
-        clickButton(AddNestedRepeaterButtonId);
-        clickButton(RemoveNestedRepeaterButtonId);
-        getElementRichTextField("nestedRepeaterText").should("not.exist");
-      });
```
Fweddi, line 612: These tests are now covered in `RepeaterFieldView.spec.ts` - we could have both unit and integration tests but this seems like overkill.

src/plugin/fieldViews/RepeaterFieldView.ts
twrichards: @fredex42 is 'adding at the end' something that's going to be needed by recipes use case... if not we should remove to avoid premature abstraction

cypress/tests/ImageElement.cy.ts
twrichards: I've re-instated these tests (but fixed) in https://github.com/guardian/prosemirror-elements/pull/334/commits/5c48fed79d9548b2f50e55fc617d5c4169224408, as these integration tests more closely test the experience of the user and the unit tests test the building blocks - a worthy distinction IMO.

https://github.com/guardian/prosemirror-elements/pull/334/commits/5c48fed79d9548b2f50e55fc617d5c4169224408 also fixes the outstanding test that was broken in https://github.com/guardian/prosemirror-elements/pull/334/commits/de60847f0e1fe9417134f8019cea148f1959328f.

twrichards: The fixes mentioned above relate to handling the fact we no longer allow removal of repeater children when there's only one. @fredex42 can I confirm this is desired behaviour for all uses of the repeater in recipes? If not, then we'll need to make `minChildren` a prop of repeater to support both use cases (and the tests will need to be updated accordingly).

rhystmills: I think we'll want to support repeater fields that have no children in general, though the Key Takeaways will always want at least one child.

Fweddi: Interesting - would a repeater field have any functionality without a child though?

rhystmills: I'm struggling to think of a useful example, but I can imagine a field where we might want to support 0-many of something.

rhystmills (commented):
This is looking great, and nice to have this functionality. I've added a couple of inline comments, otherwise:

- Worth bearing in mind that renaming the `remove` and `add` methods of the repeater will be a breaking change for the recipe element so we'd need to update that element in Composer when we start using whatever release includes this change; that's worth running by the people working on that element. 
I think the new names are better, and more descriptive.
- I'd remove the hover effect for the element delete button when it's disabled, in line with the movement controls.

src/plugin/fieldViews/RepeaterFieldView.ts
fredex42: @twrichards I am OK to remove this.... I figure that "insert" is a better case for us as it means less scrolling in the UI

cypress/tests/ImageElement.cy.ts
fredex42: @Fweddi @rhystmills @twrichards I am afraid that in recipes we definitely do need to support 0 repeater children; this is to handle the case where the structuriser fails to find something, e.g. serving quantities or timings.  The form in that case needs to show 0 of that thing and allow the user to then add it manually (as there is no sensible default that can be applied in that case)

fredex42:
Really glad to see this coming through, I'm looking forward to being able to remove my rather ugly move up/down code in https://github.com/guardian/flexible-content/pull/4528 and the ability to insert a node at a given index will improve our UI in the recipe form

src/plugin/fieldViews/RepeaterFieldView.ts
```diff
@@ -85,28 +115,100 @@ export class RepeaterFieldView extends FieldView<unknown> {
...
+      this.node,
+      index,
+      this.getPos(),
+      this.offset
+    );
+    const endOfNodeToMove = startOfNodeToMove + nodeToMove.nodeSize;
+    const previousNode = this.node.child(index - 1);
+    tr.deleteRange(startOfNodeToMove, endOfNodeToMove);
+    tr.insert(startOfNodeToMove - previousNode.nodeSize, nodeToMove);
+    this.outerView.dispatch(tr);
```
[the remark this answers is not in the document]
Fweddi: @twrichards and I chatted about this, there are different index restrictions, and the transactions need to be applied in the opposite order, so they have some core differences. If this was a strike of the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), I think we should refactor, but here I think it's fine (and is line with the other move up / down methods in this repo). 

(Also, you can argue `moveChildNPositions` would be a premature abstraction, as we only need the ability to move up or down one position at the moment.)

src/elements/alt-style/AltStyleElementForm.tsx
```diff
@@ -1,34 +1,90 @@
...
+          <LeftRepeaterActionControls
+            removeChildAt={() => fields.repeater.view.removeChildAt(index)}
+            numberOfChildNodes={children.length}
+            minChildren={1}
           />
-          <FieldWrapper
-            headingLabel="Key Takeaway Content"
-            field={repeater.content}
-            useAlternateStyles={true}
+          {/*Use field index as key to avoid React render conflicts*/}
```
twrichards, line 64: think this comment line needs to move up above 57 (or remove it since, `key` in react is a fairly well know/understood concept)

src/plugin/fieldViews/RepeaterFieldView.ts
```diff
@@ -85,28 +112,101 @@ export class RepeaterFieldView extends FieldView<unknown> {
...
+    tr.insert(positionToAddFrom, newNode);
     this.outerView.dispatch(tr);
   }
 
   /**
    * Remove a child from this repeater at the given index.
+   * Do not remove if we are at the minimum threshold for number of children.
    */
-  public remove(index: number) {
+  public removeChildAt(index: number, minChildren = 0) {
```
twrichards, line 132: is there anyway for the RepeaterFieldView to hold the `minChildren` so it doesn't need to be passed in here, I can see bugs creeping in if the call site forgets to specify and it falls back to the default

twrichards (commented):
really nice PR - two comments, one minor, one a bit more important

src/elements/alt-style/AltStyleElementForm.tsx
Fweddi: yep, needs to move up, I thought I had but missed it, thanks for spotting.

the comment is about that bug we encountered at the beginning of this work (it used to use the map index rather than the field index which caused the render conflicts on certain movements). perhaps I could clarify it? it's not saying use keys in React, it's saying field index is the valid key not map / node index.

src/plugin/fieldViews/RepeaterFieldView.ts
twrichards: I've implemented my suggestion in https://github.com/guardian/prosemirror-elements/pull/334/commits/b2dfc0cb16addda1c1491ddc60a9654c04b88c8d - let me know what you think?

twrichards (approved):
This is great - nice work @Fweddi! Approving but also worked on it a bit so if @rhystmills and/or @fredex42 want a final review.

fredex42 (approved):
I'm happy with this and looking forward to being able to use it 😁

PR #335, opened 2024-01-26
Fix high vulnerabilities

rebecca-thompson:
## What does this change?
Bumps changeset version and moves it to devDependencies. This should resolve high Snyk vulnerabilities both in this project and in flexible-content


rhystmills (approved):
Looks sensible!

There are some other yet-to-be-released changes that will go out with this package bump - I've mentioned them in the changeset.

PR #339, opened 2024-01-31
Tweak validation to account for empty key takeaway content field

Georges-GNM:
## Intro note

This looks like a really small change in comparison with the number of commits and amount of time spent, but this is due to eventually pivoting to wanting to make an initial small change (this PR), before handling a more complex consideration later (as described below). 

## What does this change?

The current validation that checks if a required field is filled is not quite working when there is only an empty text field, which gets inserted by composer. This adds a condition which should only get picked up when the content field of a nested element field (e.g. a key takeaway) has that empty element.

Also updates the altStyleElementSpec used by the demo to include validation on the content field.

## How to test
Load this branch, run the demo and add a new alt style element - both fields should come up as required. Adding any text in the content field will still show it as required, but adding a pullquote (or any other element) will get correctly registered. 

To check the validation in composer is working, [link it to composer](https://github.com/guardian/prosemirror-elements#testing-locally-in-applications-using-prosemirror-elements) running locally with [this branch](https://github.com/guardian/flexible-content/pull/4551). The content field should be shown as required, even with the standard "Enter text..." placeholder: 
image
... until any text or element is inserted into the field. 

## Next steps

This is a small PR adding some simple logic to account for an empty content field in a nested element. A couple of considerations: 

* This will appear to fully work in composer, but not in the pme demo. Adding text content won't get registered, but inserting a pullquote element (for instance) will. This seems to be due to differences on how composer and the pme demo handle the text value in a content field. AFAICT, just adding text in pme doesn't add data to the array of objects that normally constitutes the content field, but validating against that array feels like the relevant change to make. Open to suggestions though!

* The bottom border of the title field feels like it clashes with the top border of the content field, but this should be resolved by a [separate PR](https://github.com/guardian/prosemirror-elements/pull/359).

* This only checks the "top level" of a content field, as in whether it's empty or not, but it doesn't look at the validation of the nested elements themselves, which is being left to a subsequent PR.

### DIFF
--- a/src/plugin/helpers/validation.ts
+++ b/src/plugin/helpers/validation.ts
@@ -1,3 +1,6 @@
+
+import { useState } from "react";
+import { useTyperighterAttrs } from "../../elements/helpers/typerighter";
 import type {
   ErrorLevel,
   FieldValidationErrors,
@@ -5,6 +8,8 @@ import type {
   ValidationError,
   Validator,
 } from "../elementSpec";
+import { createCustomDropdownField } from "../fieldViews/CustomFieldView";
+import { createTextField } from "../fieldViews/TextFieldView";
 import type { FieldDescriptions } from "../types/Element";
 import { undefinedDropdownValue } from "./constants";
 import type { FieldNameToValueMap } from "./fieldView";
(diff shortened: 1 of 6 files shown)


Fweddi (commented):
Nice work.

In the ProseMirror demo, if you write some content and then delete it, leaving it empty, the validator thinks it's valid. This isn't an issue in Composer, but perhaps worth tightening up the logic so it's valid for both.

Sidebar: it's unclear to me how this validation logic works. I can see in this PR there are two references to a message, are these ever displayed? Is the logic from these two files linked together? 

We might want to consider how closely we tie the validation message to the field if we want to keep the field as 'content' but label it as 'body' (https://github.com/guardian/flexible-content/pull/4560).

Georges-GNM:
> In the ProseMirror demo, if you write some content and then delete it, leaving it empty, the validator thinks it's valid. This isn't an issue in Composer, but perhaps worth tightening up the logic so it's valid for both.
>
Good point - I'm not 100% sure yet what's happening with the empty field in prosemirror, but I've got the impression it's different than in composer. I do imagine there's a way to get the logic working for both though, so going to explore that 👍  

> Sidebar: it's unclear to me how this validation logic works. I can see in this PR there are two references to a message, are these ever displayed? Is the logic from these two files linked together?
> 
So from what I can tell the actual text that's in this pme PR or in other element spec files in pme doesn't actually get displayed anywhere in the demo itself - the message that's returned in pme's `validation.ts` file though does get displayed in the composer pre-publication check section, but actually specified in the element spec file in composer.

image

src/plugin/helpers/validation.ts
```diff
@@ -162,7 +162,24 @@ export const required = (
   customMessage: string | undefined = undefined,
   level: ErrorLevel = "ERROR"
 ): FieldValidator => (value, field) => {
-  if (typeof value !== "string" && value !== undefined) {
+  const requiredError = {
```
Fweddi, line 165: I would use ternaries and some slight rephrasing to simplify it further:

```suggestion
  const requiredError = {
    error: "Required",
    message: customMessage ?? `${field} is required`,
    level,
  };
  const typeError = {
    error: "Invalid field",
    message: `Checking if ${field} is required: value is invalid`,
    level,
  };

  const isArray = Array.isArray(value);
  const isString = typeof value === "string";

  if (isArray || isString) {
    return value.length === 0 ? [requiredError] : [];
  }

  if (value === undefined) {
    return [requiredError];
  }

  return [typeError];
```

Fweddi: (the suggestion might not be covering the correct lines so would recommend against using the `Commit suggestion` button if you wanted to use this code)

Fweddi (approved):

Fweddi:
LGTM - let's add the changeset in this PR?

PR #344, opened 2024-02-08
Widen alt style element

Fweddi:
## What does this change?
Before:
image

After:
image

We want more real estate for this type of element as it is closer to a document level item than an element level item. It also helps us deal with fitting our nested elements inside this element.

We now omit the headings for these fields because we want it to look closer to a document. We should consider adding better placeholders for these fields in a future PR (not trivial because placeholder logic is a bit of a muddle).

### DIFF
--- a/src/elements/alt-style/AltStyleElementForm.tsx
+++ b/src/elements/alt-style/AltStyleElementForm.tsx
@@ -1,5 +1,5 @@
 import styled from "@emotion/styled";
-import { neutral } from "@guardian/src-foundations";
+import { neutral, space } from "@guardian/src-foundations";
 import React from "react";
 import { FieldLayoutVertical } from "../../editorial-source-components/FieldLayout";
 import { FieldWrapper } from "../../editorial-source-components/FieldWrapper";
@@ -21,22 +21,20 @@ import { keyTakeawaysFields } from "./AltStyleElementSpec";
 export const AltStyleElementTestId = "AltStyleElement";
 
 const RepeaterChild = styled(Body)`
-  padding: 8px 8px 16px 8px;
-  &:not(:last-child) {
-    border-bottom: 1px dashed ${neutral[60]};
+  &:first-child {
+    margin-top: ${space[3]}px;
   }
+  margin-bottom: ${space[3]}px;
   position: relative;
 `;
 
 const RepeatedFieldsWrapper = styled("div")`
   width: 100%;
-  padding: 0 8px;
 `;
 
 const ChildNumber = styled("div")`
-  position: absolute;
-  top: 16px;
-  left: 8px;
+  box-sizing: border-box;
+  background-color: ${neutral[100]};
   color: ${neutral[46]};
   font-family: "Guardian Agate Sans", sans-serif;
   line-height: 1;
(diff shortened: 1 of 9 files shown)


src/elements/alt-style/AltStyleElementForm.tsx
```diff
@@ -21,22 +21,20 @@ import { keyTakeawaysFields } from "./AltStyleElementSpec";
...
+  &:first-child {
+    margin-top: ${space[3]}px;
   }
+  margin-bottom: ${space[3]}px;
   position: relative;
 `;
 
 const RepeatedFieldsWrapper = styled("div")`
   width: 100%;
-  padding: 0 8px;
```
twrichards, line 33: will this affect the recipes use case?

src/elements/alt-style/AltStyleElementForm.tsx
```diff
@@ -106,16 +107,8 @@ export const keyTakeawaysElement = createReactAltStylesElementSpec(
...
-        field={repeaterChild.title}
-        useAlternateStyles={true}
-      />
-      <FieldWrapper
-        headingLabel="Key Takeaway Content"
-        field={repeaterChild.content}
-        useAlternateStyles={true}
-      />
+      <AltFieldWrapper field={repeaterChild.title} />
+      <AltFieldWrapper field={repeaterChild.content} />
```
twrichards, line 111: does this meet rule of 3, and should the abstraction not live in this file rather than FieldWrapper.tsx

src/elements/alt-style/AltStyleElementForm.tsx
```diff
@@ -45,10 +43,13 @@ const ChildNumber = styled("div")`
...
   align-items: center;
-  height: 24px;
+  height: 32px;
   padding: 2px;
-  width: 24px;
+  width: 32px;
   border: 1px solid ${neutral[60]};
+  position: absolute;
+  top: 0;
+  left: -32px;
```
twrichards, line 52: Should this number be inside `LeftRepeaterActions` to avoid repetition (and not having to maintain the negative positioning in more than one place?

twrichards (approved):
LGTM, left a few comments? 

I presume placing the controls above & below instead will be in a follow-up PR?

Any chance we could increase the left padding inside the inputs and increase the negative margin so the boxes are actually full width but text still aligns with stuff outside the repeater? (perhaps ignore if this doesn't affect composer, because no outer grey border) to essentially avoid this uncomfortable alignment on the left...
image

src/elements/alt-style/AltStyleElementForm.tsx
Fweddi: I had it in there to start with, but wanted the Child Number to always be visible so moved it out (if it's in actions we can only see it on hover).

src/elements/alt-style/AltStyleElementForm.tsx
Fweddi: Shouldn't do as this form isn't used by the recipe element

src/elements/alt-style/AltStyleElementForm.tsx
Fweddi: I don't see how it meets the rule of three, can you clarify? Happy to move the logic around

twrichards: I meant, does it need an abstraction if there are only two?

src/elements/alt-style/AltStyleElementForm.tsx
twrichards: makes sense, thanks for clarifying. perhaps worth moving the -32 to a `const`?

PR #349, opened 2024-02-14
FIX 'NestedElement' IN repeater bug

twrichards:
https://trello.com/c/ZRa7FLPg/1979-solve-repeater-nested-field-bug

### Background 
Throughout the development of Key Takeaways (part of the templates/formats OKR) we've been plagued by a rather pernicious bug which occurs when the new `NestedElements` (i.e. instances of `ProseMirrorFieldView` i.e. inner prosemirror editors) are used inside the relatively new 'repeater' functionality. Specifically, `RangeErrors` are thrown by the internals of prosemirror, ultimately invoked by the function `applyDecorationsFromOuterEditor` in `ProseMirrorFieldView.ts` (which AFAICT was originally implemented following @jonathonherbert's [discussion with ProseMirror maintainer here](https://discuss.prosemirror.net/t/is-it-possible-to-get-hold-of-decorations-that-cover-the-range-managed-by-a-nodeview-in-that-nodeview/3544)) - which works fine for `ProseMirrorFieldView`s outside of repeaters or indeed normal `FieldViews` inside repeaters.

After intensive interactive debugging sessions with @rhystmills we discovered that the each repeater child was receiving all the decorations for the whole repeater and attempting apply all decorations to each repeater child and hence RangeErrors were thrown (e.g. tried to apply decoration from child C to child A... 💥 ). 

Decorations are essential for features such as noting, Typerighter etc. so we really needed to solve this problem.

Temporary workarounds were put in place (primarily try/catch around the offending parts of `applyDecorationsFromOuterEditor`) to unblock other work from proceeding while we got to the bottom of this.

Spent too long (in hindsight) thinking this issue was down to offsets being slightly wrong, before thinking to add an instance of NestedElement outside the repeater and observing the shape of the decorations that got passed into `applyDecorationsFromOuterEditor` then using the debugger tracing back through the call-stack into the internals of prosemirror where I discovered the 'internal' function `forChild(offset: number, childNode: Node): DecorationSource` available on `DecorationSource` interface (`DecorationSet` & `DecorationGroup` are the implementations) which was the key to this fix.

Becuase `field.ts` has various conditional blocks to deal with repeaters the decorations being passed into the repeater children were essentially one level too high up the tree of decorations.

### Changes

- Added a cypress test to cover the `NestedElement` IN 'repeater` scenario (in the `AltStylesElement.cy.ts` collection of tests)
- Abstracted the retrieval of the repeater decorations into a function `getRepeaterDecorations` in `field.ts` so it could be used in the various repeater specific places in the file, this does two main things...
  - 'steps down' the missing level (i.e. inside the repeater parent) of the decoration tree using the `forChild` method mentioned above
  - offsets the decorations by `1` to account for the repeater parent, to counter the `-1` offset in `applyDecorationsFromOuterEditor` of `ProseMirrorFieldView`

The `forChild` method was marked as internal in https://github.com/prosemirror/prosemirror-view, so I raised https://github.com/ProseMirror/prosemirror-view/pull/165 and https://github.com/ProseMirror/prosemirror-view/pull/166. to expose it for use in this fix. Upgrade to `prosemirror-view` `1.33.1` ended up requiring upgrading all the prosemirror libs - which I've encapsulated in https://github.com/guardian/prosemirror-elements/pull/350.

### Further Notes
I did notice that `RepeaterFieldView` does have things like...
image
...so I have wondered whether some of the conditional logic for repeaters in `field.ts` was temporary located there and planned to be encapsulated in `RepeaterFieldView` (@jonathonherbert will likely know) - if so I think worth revisiting now we have various PRODish use cases for repeaters now.

### Testing
Using `yalc` (see README) release the library locally and import into local composer, you should be able to create a Key Takeways article (behind feature switch) and then add multiple key takeaways, each with some content, then try noting out that text and also try running Typerighter (be sure to log into CODE composer first) with some intentional mistakes in the content to see that everything looks correct 🤞


jonathonherbert:
Brilliant spot, great work on a fix! Back next Monday if a review is still needed.

Re: update and destroy lifecycle methods – `update` would be stomping in a new array of repeater values, so if there's a use case that'd be worth implementing. I think `destroy` is always be a no-op for repeaters, as the repeater doesn't directly manage fieldViews, and has no view to manage itself (thus nothing to destroy) – if that's right, we could note that more clearly in the method and remove the logging.

Re: `field.ts`, architecturally there's plenty of logic spread in ad-hoc methods that might be better colocated with field definitions! Might be a good candidate with a refactor, alongside (from memory) the node <-> value transforms, maybe there are more things.

(The ideal from an extensibility perspective would I think be that adding a new field type would mean implementing an interface, rather than patching in translations in a few different places that won't be clear to new devs.)

davidfurey (approved):
I haven't reviewed in detail (hopefully Rhys or Jon are better placed to do so), but I'd be happy for you to merge it as is if it unblocks other work.

Georges-GNM:
Tested this locally and it's working as expected, fantastic 🎉 

One thing to flag and which @twrichards and I investigated a bit was that the notes in a key takeaway content field can be added, but are missing some functionality (e.g. clicking on the wrapper of a note doesn't expand/collapse it as it normally should). 

It seems this is due to the fact that the key takeaway is getting its noting functionality from the outerDecorations logic, but in order for it to fully work as intended it requires some prosemirror-noting plugin to be enabled in the `allowedPlugin` array in the [KeyTakeawaySpec.tsx](https://github.com/guardian/flexible-content/blob/055aed6a0244284925092d04615a0a8e54684246/composer/src/js/prosemirror-setup/elements/key-takeaways/KeyTakeawaysSpec.tsx#L15). The relevant logic seems to sit in prosemirror-elements [NestedElementFieldView.ts](https://github.com/guardian/prosemirror-elements/blob/73e41d8eeb4cf4b516f1d69fcee2fa055f695a5e/src/plugin/fieldViews/NestedElementFieldView.ts#L100) file. 

But this is something that can be solved in a subsequent PR.

Georges-GNM (approved):
I could verify the functionality running this locally - typerighter getting triggered and notes being enabled in a second key takeaway: 
image

The underlying logic with the offsetting makes sense to me (although feels worth caveating I'm not a prosemirror pro like Rhys or Jon), so happy to approve.

PR #353, opened 2024-02-23
Display element controls vertically when elements are nested

Fweddi:
## What does this change?
When using nested elements within repeaters, we found there was often a clash between them and we were worried that users might get the two control sets confused. 

This PR displays the controls differently depending on whether the controls are nested or not:
* Regular Element => Horizontal Controls (to the sides of the element) - as before
* Nested Element => Vertical Controls (on top or on the bottom of the element) - **new**

### Before:
image

### After:
https://github.com/guardian/prosemirror-elements/assets/40991816/d882c408-6eae-4422-9fe7-4509833b24e0

## How to test
See if the controls work as expected and render as before for regular elements. You might want to also test it in a consuming application (e.g. Composer) using `yalc`.

Note that using the nested controls exposes the placeholder bugs we are experiencing more keenly - but that's not in the purview of this PR.

## Notes
It could be argued this doesn't improve the situation as much as we would have hoped. The vertical spacing between elements is much larger than before (to accommodate the vertical controls).

If we had we increased this spacing just a little, without making the nested controls display vertically, we would have eliminated most of the clash between the control sets. This PR is definitely better to help distinguish between the two, but we should ask ourselves if this is worth the trade-off of increasing the vertical spacing.

I'm also open to other suggestions and ideas.

### DIFF
--- a/src/renderers/react/ElementWrapper.tsx
+++ b/src/renderers/react/ElementWrapper.tsx
@@ -14,13 +14,19 @@ export const Container = styled("div")`
 
 export const Body = styled("div")`
   display: flex;
-  :hover,
-  :focus-within {
-    .actions {
-      opacity: 1;
-    }
-  }
   min-height: 134px;
+  &:not(:hover) .actions,
+  &:not(:focus-within) .actions,
+  .nested &:not(:hover) .actions,
+  .nested &:not(:focus-within) .actions {
+    opacity: 0;
+  }
+  &:hover .actions,
+  &:focus-within .actions,
+  .nested &:hover .actions,
+  .nested &:focus-within .actions {
+    opacity: 1;
+  }
 `;
 
 const Panel = styled("div")<{
--- a/src/renderers/react/ElementWrapper.tsx
+++ b/src/renderers/react/ElementWrapper.tsx
@@ -3,13 +3,13 @@ import { space } from "@guardian/src-foundations";
 import { neutral } from "@guardian/src-foundations/palette";
 import type { ReactElement } from "react";
 import React, { useContext, useState } from "react";
-import { actionSpacing } from "../../plugin/helpers/constants";
 import type { CommandCreator } from "../../plugin/types/Commands";
 import { TelemetryContext } from "./TelemetryContext";
 import { LeftActionControls, RightActionControls } from "./WrapperControls";
 
 export const Container = styled("div")`
-  margin: ${space[3]}px -${actionSpacing}px;
+  margin: ${space[3]}px 0;
+  position: relative;
 `;
 
 export const Body = styled("div")`
--- a/src/renderers/react/ElementWrapper.tsx
+++ b/src/renderers/react/ElementWrapper.tsx
@@ -15,17 +15,31 @@ export const Container = styled("div")`
 export const Body = styled("div")`
   display: flex;
   min-height: 134px;
-  &:not(:hover) .actions,
-  &:not(:focus-within) .actions,
-  .nested &:not(:hover) .actions,
-  .nested &:not(:focus-within) .actions {
+  .wrapper:not(:hover) .actions,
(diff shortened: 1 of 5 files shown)


src/renderers/react/WrapperControls.tsx
```diff
@@ -285,9 +305,24 @@ export const RightActionControls = ({
...
+        onClick={() => {
+          sendTelemetryEvent?.(CommandTelemetryType.PMEDownButtonPressed, {
+            jump: false,
+          });
+          moveDown(true);
+        }}
+        aria-label="Move element down"
+      >
+        <SvgArrowDownStraight />
+      </Button>
```
Fweddi, line 322: This button is re-ordered to account for some column / column-reverse shenanigans.

src/renderers/react/WrapperControls.tsx
```diff
@@ -32,15 +32,14 @@ const Button = styled("button")<{ expanded?: boolean }>`
...
-  border-top: 1px solid ${neutral[100]};
   color: ${neutral[0]};
   cursor: pointer;
   flex-grow: ${({ expanded }) => (expanded ? "1" : "0")};
-  ${({ expanded }) => !expanded && `height: ${buttonWidth}px;`};
   font-size: 15px;
   line-height: 1;
   padding: ${space[1]}px;
   width: ${buttonWidth}px;
+  height: ${buttonWidth}px;
```
Fweddi, line 42: This might cause problems, I couldn't see what the use cases for 'expanding' were but would be good to test.

src/renderers/react/WrapperControls.tsx
```diff
@@ -141,6 +124,54 @@ const Tooltip = styled("div")`
...
 
+type VerticalPosition = "top" | "bottom";
+type HorizontalPosition = "left" | "right";
+
+const Actions = styled("div")`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  width: ${actionSpacing}px;
+  z-index: 11;
```
Fweddi, line 135: This z-index is needed (in PME at least) to avoid clashing with the PME demo rich text controls.

src/renderers/react/ElementWrapper.tsx
```diff
@@ -9,18 +9,30 @@ import { TelemetryContext } from "./TelemetryContext";
...
 export const Body = styled("div")`
   display: flex;
-  :hover,
-  :focus-within {
-    .actions {
-      opacity: 1;
-    }
-  }
   min-height: 134px;
+  .nested & {
```
Fweddi, line 19: `&`: https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector

src/renderers/react/WrapperControls.tsx
Georges-GNM: I think `expanded` gets used to set the height of the buttons depending on the size of the element, but it feels like the `flex-grow` property is maybe doing the job now?

Either way can't see any of the buttons looking out of the ordinary, so seems to be working.

src/renderers/react/ElementWrapper.tsx
```diff
@@ -9,18 +9,28 @@ import { TelemetryContext } from "./TelemetryContext";
...
```
Georges-GNM, line 19: I think the extra vertical space is not too problematic, but we can maybe shave a few pixels by adding some negative margin here.

Georges-GNM (approved):
Tested in prosemirror and composer, looks good to me 👍

Georges-GNM (approved):
Tested in prosemirror and composer, looks good to me 👍

Fweddi:
On surveying a few people I have decided against implementing this PR in the described way. I am going to reraise a PR with this branch to reflect the changes I have made. The headline is we are _not_ nesting element controls.

PR #359, opened 2024-03-04
collapse border between adjacent alt-style fields

twrichards:
This double line was grating on me whilst working on something else. 
CSS is a bit fiddly given the layers of DOM between the div with the existing `border` css and the the thing which would respond to `:not(:first-of-type)` selector.

Also, nudged the alt-style `ChildNumber` over a pixel, again to avoid double/adjacent borders.

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/src/editorial-source-components/FieldWrapper.tsx
+++ b/src/editorial-source-components/FieldWrapper.tsx
@@ -1,3 +1,4 @@
+import { css } from "@emotion/react";
 import type { ValidationError } from "../plugin/elementSpec";
 import type { FieldView } from "../plugin/fieldViews/FieldView";
 import type { Field } from "../plugin/types/Element";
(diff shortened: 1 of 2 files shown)


simonbyford:
Change looks great and it would be fab to get this fixed!

It might be an issue with my local environment/yalc setup but I can't see this working locally for some reason, with or without a rebase. The style doesn't come through. Anyone else able to verify?

rhystmills:
> Change looks great and it would be fab to get this fixed!
> 
> It might be an issue with my local environment/yalc setup but I can't see this working locally for some reason, with or without a rebase. The style doesn't come through. Anyone else able to verify?

Yeah I'm seeing the same - the double border is gone between the number and the title but not between the title and the body content.

Fweddi:
Think this is now covered in https://github.com/guardian/prosemirror-elements/pull/382.

PR #362, opened 2024-03-11
propagate click events upwards from inner editors (e.g. `NestedElementFieldView`) up to the top-level editor so that plugins work

twrichards:
https://trello.com/c/MhfcXb4t/2168-ensure-plugins-work-across-the-nestedelement-boundary-eg-noting-dragdrop-media-etc

Whilst reviewing #349 @Georges-GNM observed that when used in composer, notes weren’t individually clickable when inside a 'NestedElement` (i.e. an inner ProseMirror instance). This is because the [noting plugin](https://github.com/guardian/prosemirror-noting) operates on the top-level editor (and we don't re-register the plugin for inner editors as they instances of the plugin will essentially fight over the content).

Digging deeper, by utilising the debugger to view the call stack when interacting with notes in the top-level editor and comparing to interacting with notes in inner editors, the default `handleClick` for the inner editor was effectively eating the event and preventing it from travelling up to outer editor (where the plugin could do its thing). 

## What does this change?
This PR address the above (for more than just plugins) by defining a `handleClick` for all `ProseMirrorFieldView` (i.e. inner editors) and explicitly calls the `handleClick` for the 'outer view' (i.e. top-level editor) and crucially passing that function the outer view for the `view` argument to ensure when handled (in plugins etc.) it has access to the right view and state to perform as it should.

## How to test
Yalc it into composer local and add a note inside a nested element (e.g. key takeaways) and observe that you can click the left/right side of a note and collapse & expand it individually.

More behaviour that users will naturally expect to work should indeed work.

### DIFF
--- a/src/plugin/fieldViews/NestedElementFieldView.ts
+++ b/src/plugin/fieldViews/NestedElementFieldView.ts
@@ -97,10 +97,11 @@ export class NestedElementFieldView extends ProseMirrorFieldView {
       getPos,
       offset,
       decorations,
-      outerView.state.plugins.filter((plugin) =>
-        plugin.spec.key === undefined
-          ? true
-          : [pluginKey, ...allowedPlugins].includes(plugin.spec.key)
+      outerView.state.plugins.filter(
+        (plugin) =>
+          !plugin.spec.key ||
+          pluginKey === plugin.spec.key ||
+          allowedPlugins.includes(plugin.spec.key)
       ),
       placeholder,
       isResizeable
--- a/cypress/tests/EventPropagationFromInnerEditors.cy.ts
+++ b/cypress/tests/EventPropagationFromInnerEditors.cy.ts
@@ -1,0 +1,30 @@
+import type { WindowType } from "../../demo/types";
+import {
+  addAltStyleElement,
+  getElementRichTextField,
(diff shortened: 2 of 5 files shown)


Georges-GNM (approved):
Makes sense and seems to be working as intended 👍

src/plugin/fieldViews/NestedElementFieldView.ts
```diff
@@ -97,11 +97,14 @@ export class NestedElementFieldView extends ProseMirrorFieldView {
...
       decorations,
-      outerView.state.plugins.filter((plugin) =>
-        plugin.spec.key === undefined
-          ? true
-          : [pluginKey, ...allowedPlugins].includes(plugin.spec.key)
-      ),
+      [
+        ...outerView.state.plugins.filter(
+          (plugin) =>
+            !plugin.spec.key ||
```
davidfurey, line 103: I presume this is just a refactor to make the code easier to read?

But I'm confused about why you need to spread the array into a new array.

twrichards: good spot, this spead is left-over from when experimenting with a few explicitly added plugins, will address now

PR #364, opened 2024-03-13
upgrade cypress

twrichards:
upgraded cypress to latest in the hope of fixing some tests which fail in CI (but work locally) in #363 - didn't help but no harm in releasing this upgrade stuff to save us a job in future


twrichards:
Closing as this doesn't help with #363 (as state already) and in fact breaks some cypress tests.

PR #365, opened 2024-03-14
Always update fieldView of nestedElement field

rhystmills:
## What does this change?
To be used in conjunction with https://github.com/guardian/flexible-content/pull/4645

This PR makes sure `nestedElement` fieldViews are always updated, as a workaround for a bug with the zipRoot plugin, that causes the document to be uneditable when an element is deleted between two text elements.

## How to test
1. Use this branch locally with Composer using yalc.
2. Do all the things suggested in the other PR: https://github.com/guardian/flexible-content/pull/4645

### DIFF
--- a/.changeset/neat-guests-study.md
+++ b/.changeset/neat-guests-study.md
@@ -1,0 +1,8 @@
+---
+"@guardian/prosemirror-elements": patch
+---
+
+Always update fieldViews for nestedElement fields, in order to fix
+a problem related to our zipRoot plugin in Composer, where joining
+text elements within a nestedElement causes the document to enter
+an invalid state, unless the fieldViews are updated again.
--- a/src/plugin/fieldViews/NestedElementFieldView.ts
+++ b/src/plugin/fieldViews/NestedElementFieldView.ts
@@ -38,12 +38,12 @@ export interface NestedElementFieldDescription
 export const anyDescendantFieldIsNestedElementField = (node: Node) => {
   let descendantFieldIsNestedElementField = false;
   node.descendants((node) => {
-    if (node.type.spec.content = "element+"){
-      descendantFieldIsNestedElementField = true
+    if (node.type.spec.content === "element+") {
+      descendantFieldIsNestedElementField = true;
     }
-  })
+  });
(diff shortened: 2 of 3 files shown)


davidfurey (commented):
Seems reasonable to me

jonathonherbert:
If we've got time, I'd recommend taking a little more time to figure out exactly what's going wrong here, and either
- dropping the optimisation if it's proving to be too aggressive, rather than working around a special case
- discovering that something's not right downstream from this update. Top suspects include the [node diffing in ProsemirrorFieldView](https://github.com/guardian/prosemirror-elements/blob/b5cbb4aef54e5aaee23fc2ad043ed010a335a895/src/plugin/fieldViews/ProseMirrorFieldView.ts#L149), which tries to update the minimal part of the doc. If this code was the culprit, another approach that would not degrade performance for general documents would be to make `updateInnerEditor` protected, and override it to remove the diffing from `NestedFieldView`.

If there's not time for that, then 👍  we can always circle back to fix.

davidfurey:
> If we've got time, I'd recommend taking a little more time to figure out exactly what's going wrong here, and either
> 
> * dropping the optimisation if it's proving to be too aggressive, rather than working around a special case
> * discovering that something's not right downstream from this update. Top suspects include the [node diffing in ProsemirrorFieldView](https://github.com/guardian/prosemirror-elements/blob/b5cbb4aef54e5aaee23fc2ad043ed010a335a895/src/plugin/fieldViews/ProseMirrorFieldView.ts#L149), which tries to update the minimal part of the doc. If this code was the culprit, another approach that would not degrade performance for general documents would be to make `updateInnerEditor` protected, and override it to remove the diffing from `NestedFieldView`.
> 
> If there's not time for that, then 👍 we can always circle back to fix.

I think our current preference would be to circle back to fix, so that we can get key takeaways out to users

Georges-GNM (approved):
Have tested with the appropriate composer branch and can confirm we're seeing the desired behaviour; makes sense to me to be using this approach in prod for the moment, considering the significant amount of time that's already been spent on it!

PR #372, opened 2024-03-25
Scroll to focused field, not top of element, when calling history commands in text editor fields

jonathonherbert:
_co-authored-by_: @Divs-B 

## What does this change?

Ensure that the editor scrolls to the focused field, not the top of the element, when undoing or redoing in text editor fields.

This makes use of a change that [@Divs-B requested in prosemirror-history](
https://github.com/ProseMirror/prosemirror/issues/1451) to provide a way of triggering a history command without a scroll side effect 🎉 

|Before|After|
|--|--|
|scroll-incorrect|scroll-correct|

## How to test

Running locally, have a play with undoing text changes within a text or rich text field. Does the scroll behaviour work as expected?

It'd be nice to include an integration test for this behaviour. We timeboxed writing a test for this, but ran out of time. 

Less user confusion when hitting Mod-Z!

### DIFF
--- a/.changeset/loud-years-care.md
+++ b/.changeset/loud-years-care.md
@@ -1,0 +1,5 @@
+---
+"@guardian/prosemirror-elements": patch
+---
+
+Fix scrolling to top of element on undo in element field
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "prosemirror-commands": "^1.5.2",
     "prosemirror-dev-tools": "^4.0.0",
     "prosemirror-example-setup": "^1.2.2",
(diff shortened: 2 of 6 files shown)


Divs-B:
Thanks for pairing on this @jonathonherbert, great to learn with you, starting from analysis on `undo` problem that has scroll to view in it, then to post the issue to Prosemirror to suggest solution such as undo without Scroll, and now testing new introduced `undoNoScroll` (method in Prosemirror) by using it in this PR.

Divs-B:
Tested locally and fix is working as expected 👍

twrichards (approved):
makes sense and looks like you've tested 🎉

PR #374, opened 2024-03-25
Do not coerce HTML nodes without field attributes to a field value of `{}`

jonathonherbert:
## What does this change?

Do not coerce HTML nodes without field attributes to a field value of `{}`. This has been causing problems in flexible-content with recipe elements, which often contain optional fields that should be `undefined`. When they instead become `{}`, the rendering code throws an error.

Also removes a redundant code path where we deserialise field values from an element node. Element nodes don't contain fields, so this step isn't required. Because the `fields` attr is not specified in the Element `NodeSpec`, they were never added anyway.

## How to test

The integration test should verify the change:
  - Before the fix in 81bccb5f534b3b442aa64000b97f153e5bb14b84, a field that is undefined by default will receive a value of `{}` when it is deserialised from HTML, and the test with the additional field `undefinedByDefault` will fail.
  - 81bccb5f534b3b442aa64000b97f153e5bb14b84 fixes by only providing a value to `fields` if the relevant attr is defined.

There remains the unhappy path of `JSON.parse` attempting to serialise invalid JSON and blowing up. We've never caught it in production, but the path is there. That's probably another PR.

### DIFF
--- a/src/plugin/nodeSpec.ts
+++ b/src/plugin/nodeSpec.ts
@@ -298,9 +298,7 @@ const getDefaultParseDOMForLeafNode = (nodeName: string) => [
       }
 
       const maybeFieldAttrs = dom.getAttribute("fields");
-      const fields = maybeFieldAttrs
-        ? (JSON.parse(maybeFieldAttrs) as unknown)
-        : undefined;
+      const fields: unknown | undefined = maybeFieldAttrs && JSON.parse(maybeFieldAttrs)
 
       return { fields };
     },
--- a/src/plugin/nodeSpec.ts
+++ b/src/plugin/nodeSpec.ts
@@ -299,11 +299,12 @@ const getDefaultParseDOMForLeafNode = (nodeName: string) => [
         return false;
       }
 
-      const attrs = {
-        fields: JSON.parse(dom.getAttribute("fields") ?? "{}") as unknown,
-      };
+      const maybeFieldAttrs = dom.getAttribute("fields");
+      const fields = maybeFieldAttrs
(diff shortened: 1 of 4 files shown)


twrichards (approved):
the code (and justification in PR description) LGTM 👍 

Please could you double check this changes doesn't adversely affect key takeaways etc. (behind a feature switch)

src/plugin/nodeSpec.ts
```diff
@@ -299,11 +297,12 @@ const getDefaultParseDOMForLeafNode = (nodeName: string) => [
         return false;
       }
 
-      const attrs = {
-        fields: JSON.parse(dom.getAttribute("fields") ?? "{}") as unknown,
-      };
+      const maybeFieldAttrs = dom.getAttribute("fields");
+      const fields = maybeFieldAttrs
+        ? (JSON.parse(maybeFieldAttrs) as unknown)
+        : undefined;
```
twrichards, line 303:
```suggestion
      const fields: unknown | undefined = maybeFieldAttrs && JSON.parse(maybeFieldAttrs)
```

twrichards: feel free to drop the last two commits if this suggestion is breaking stuff (tests look to be failing)

jonathonherbert: Ah yep, this changes `undefined` to `null`, null values are serialised to html as the string `null` 🤔

jonathonherbert:
@twrichards copying and pasting takeaways with content of various types works as expected 👍 

If we would like more safety around serialisation of nested elements let's add more tests, v. happy to pair.

PR #375, opened 2024-03-26
Make list element title and content fields align with body in Composer

rhystmills:
## What does this change?
Makes some minor tweaks to `nestedElement` field and `altStyle` wrapper to make sure the list element title and content align with the main editor text in Composer. They were once aligned but there has been some drift.

| Before | After |
| --- | --- |
| image | image |

## How to test
1. Use `yarn yalc` in this repo to publish prosemirror-elements locally.
2. Use it in the composer directory of flexible-content with `yalc add @guardian/prosemirror-elements`
3. Create one of the list elements and check out the alignment.


simonbyford (approved):
This was very jarring, thank you

PR #376, opened 2024-03-27
Fix Prosemirror tests which pass in CI, but fail locally

davidfurey:
There are some element movement tests in ElementWrapper which are a bit flaky - they fail locally but pass in CI.

The helper function getElementType in editor.ts seems to fail to find the elementWrapperTestId for an element, which means its returns unknown and fails the test.

Fixed by bumping Cypress version


Fweddi:
> The Cypress PR has been merged and is awaiting release. We will probably see a release in the next two weeks (based on the average cadence of the [release schedule](https://github.com/cypress-io/cypress/releases)).

What do you know, it's now out: https://github.com/cypress-io/cypress/releases/tag/v13.7.2

Fweddi:
Tests now passing locally 👍

Fweddi (approved):

PR #382, opened 2024-04-10
Remove double border between nested fields and text fields

rhystmills:
Co-authored with @simonbyford.

We thought we'd have another pop at this to see if we could avoid the side effects in https://github.com/guardian/flexible-content/pull/4704

This time around, we've used outlines rather than borders to avoid layout shifts when we apply the styles, and to make sure validation and focus states are always 'above' the existing borders.

### Before:
image

### After:
image
image
image
image
image

## How to test
Pull this branch into composer locally with `yalc` and check that this is working properly in one of the list elements.

### DIFF
--- a/.changeset/chilled-ravens-run.md
+++ b/.changeset/chilled-ravens-run.md
@@ -1,0 +1,5 @@
+---
+"@guardian/prosemirror-elements": patch
+---
+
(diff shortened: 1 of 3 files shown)


simonbyford (approved):
🥳

PR #386, opened 2024-04-16
Support alt styles for Timelines

rhystmills:
## What does this change?
This exports the `AltStyleElementWrapper` for use in the `TimelineForm` in Composer. We may want to move this to Composer in the long run, but I'm leaving that for a future optimisation.

It also adds a `flex-grow` style to ProsemirrorFields, i.e. text, rich text and nested. This won't affect most uses but will cause the text fields to fill their horizontal space in a flex layout.

## How to test
This should be a no-op, but you can test locally with my current branch on Composer:
https://github.com/guardian/flexible-content/compare/main...rm/alt-style-timeline


Georges-GNM (approved):
👍 
Tried to spot where the `flex-grow` might have an impact, but couldn't see it affecting the main elements we've got at the moment, so seems fine to me.

PR #394, opened 2024-05-01
Fix element movement near list elements

Fweddi:
## The Problem
Before this PR, moving elements **up** past a list element was a dangerous pursuit, sometimes even deadly. 

1. If your list element had content, the element moving up would jump into the list element. This is not desired: I think we want the element to move _past_ the list element[^1]. 

https://github.com/guardian/prosemirror-elements/assets/40991816/caff134f-b9f2-4424-bd4e-5876641fb0b0

2. If your list element was empty, the element moving up would get deleted. Also not desired, you would think.

https://github.com/guardian/prosemirror-elements/assets/40991816/0dcfa31e-f84e-4c5a-a983-79367e479615

So what's going on? And why is it only affecting elements moved up?

## Explanation

When moving an element, we search for the next node along in the document to help calculate where to move our element. 

When moving down past a list element, it finds the next node correctly: the list element. It then using the list element's position and size to calculate our moved element's next position.

When moving up past a list element, it thinks that the next node isn't the list element itself, but a child node within the list element, and so breaks in the patterns described above.

This is a product of how the ProseMirror document hierarchy functions. At the "top" of a list element is the list element itself (parent), and at the "bottom" is some a nested element (child).

This is why moving down works, and moving up fails - the direction you approach the node from affects whether you are looking at the list element as a whole (down) or a child node (up).

#### Moving Down

* OurElement (If we move this element down...)
* ListElement (...it would find this node...)
	* NestedElement1
	* NestedElement2
* _OurElement (...and move to here.)_

___

#### Moving Up

* ListElement 
	* NestedElement1
	* _OurElement (...and move to here.)_
	* NestedElement2 (...it would find this node...)
* OurElement (If we move this element up...)

Or (when list element has no content):

* ListElement (...and break - deleting OurElement.)
	* NestedElement1 (...it would find this node...)
* OurElement (If we move this element up...)

## The Solution

The solution is to take advantage of the underlying ProseMirror method, [nodesBetween](https://prosemirror.net/docs/ref/#model.Node.nodesBetween) (used by our own `nodesBetween` method, which is used in `nextPosFinder` in `moveNode`):
> Invoke a callback for all descendant nodes recursively between the given two positions that are relative to start of this node's content. The callback is invoked with the node, its position relative to the original node (method receiver), its parent node, and its child index. **When the callback returns false for a given node, that node's children will not be recursed over**. The last parameter can be used to specify a starting position to count from.

I've highlighted the relevant part in bold. We can choose not to recurse in a given node, which would allow us to ignore the child nodes of a list element, and treat the whole element as if it were one node:

#### Moving Up (Take 2)
* _OurElement (...and move to here.)_
* ListElement (...it would find this node...)
	* NestedElement1 (...ignoring this node...)
	* NestedElement2 (...ignoring this node...)
* OurElement (If we move this element up...)

This fixes our problem, allowing us to move elements up and down past the list element as intended![^2]

## How to test
Publish to Composer locally using yalc.

Move top-level elements past a list element. Does this now work?

Does moving nested elements within list elements still work as before?

Are there any other movement related issues generated by this PR?

[^1]: I think the right approach is to move elements past list elements, not within them, but happy to hear arguments either way!

[^2]: I initially prevented recursion only when looking at list nodes, but discovered that preventing recursion everywhere is non-breaking, so went with that approach. The advantage of this is we don't need to check for list element types (the definitions for which are downstream in Composer), or nested-ness.

### DIFF
--- a/src/plugin/helpers/prosemirror.ts
+++ b/src/plugin/helpers/prosemirror.ts
@@ -17,16 +17,15 @@ const nodesBetween = (state: EditorState, _from: number, _to: number) => {
   state.doc.nodesBetween(from, to, (node, pos, parent, index) => {
     arr.push([node, pos, parent, index]);
     /*
-      Returning false from state.doc.nodesBetween prevents recursion into the node's children.
-      We don't want to recurse.
+      Returning false from state.doc.nodesBetween prevents recursion into the node's children. We want
+      to allow one level of recursion (for textElements, which contain paragraphs), but no further.
 
-      This allows us to treat list elements and their nested contents as a single entity for movement purposes.
-      In other words, we can move top-level elements up and down past a list element.
-      Before, with recursion, top-level elements would move inside the list element, often breaking the document structure.
-
-      Note this still allows us to move nested elements **within** list elements.
+      This allows us to treat list elements and their contents as a single entity for movement purposes.
+      This means we can move top-level elements up and down past a list element. Before, with
+      recursion, top-level elements would move inside the list element, often breaking the document
+      structure. Note this still allows us to move nested elements **within** list elements.
     */
-    return false;
+    return parent?.type.name === "doc";
   });
   if (dir < 0) {
     arr.reverse();
--- a/src/plugin/helpers/prosemirror.ts
+++ b/src/plugin/helpers/prosemirror.ts
@@ -13,13 +13,21 @@ const nodesBetween = (state: EditorState, _from: number, _to: number) => {
   const range = [_from, _to];
   range.sort((a, b) => a - b);
   const [from, to] = range;
-
   const arr: NodesBetweenArgs[] = [];
-  state.doc.nodesBetween(
-    from,
-    to,
-    (node, pos, parent, index) => !!arr.push([node, pos, parent, index])
-  );
+  state.doc.nodesBetween(from, to, (node, pos, parent, index) => {
+    arr.push([node, pos, parent, index]);
+    /*
+      Returning false from state.doc.nodesBetween prevents recursion into the node's children.
+      We don't want to recurse.
+
+      This allows us to treat list elements and their nested contents as a single entity for movement purposes.
+      In other words, we can move top-level elements up and down past a list element.
+      Before, with recursion, top-level elements would move inside the list element, often breaking the document structure.
+
+      Note this still allows us to move nested elements **within** list elements.
+    */
+    return false;
+  });
   if (dir < 0) {
     arr.reverse();
   }
--- a/src/plugin/helpers/prosemirror.ts
+++ b/src/plugin/helpers/prosemirror.ts
@@ -4,6 +4,7 @@ import type { EditorState, Transaction } from "prosemirror-state";
 import { AllSelection, NodeSelection, TextSelection } from "prosemirror-state";
 import type { EditorView } from "prosemirror-view";
 import { Decoration, DecorationSet } from "prosemirror-view";
+import { anyDescendantFieldIsNestedElementField } from "../fieldViews/NestedElementFieldView";
 
 type NodesBetweenArgs = [Node, number, Node | null, number];
 export type Commands = ReturnType<typeof buildCommands>;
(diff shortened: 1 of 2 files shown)


src/plugin/helpers/prosemirror.ts
```diff
@@ -13,13 +13,21 @@ const nodesBetween = (state: EditorState, _from: number, _to: number) => {
   const range = [_from, _to];
   range.sort((a, b) => a - b);
   const [from, to] = range;
-
   const arr: NodesBetweenArgs[] = [];
-  state.doc.nodesBetween(
-    from,
-    to,
-    (node, pos, parent, index) => !!arr.push([node, pos, parent, index])
```
Fweddi, line 21: The !! in `!!arr.push([node, pos, parent, index])` was to achieve a boolean typing, I'm not sure we were really using the logic in practice. See 15e19d1.

simonbyford (commented):
Firstly, superb PR description, so descriptive and helpful, it is very much appreciated 🙏 

I've played around and have verified this has fixed the bug!

One regression seems to be that you can no longer skip over individual paragraphs:

https://github.com/guardian/prosemirror-elements/assets/7423751/7df36da7-84ba-4ec2-913d-1b53af970128

I think this is because paragraphs are stored as child elements of a `textElement`. So recursion is probably desirable in this case.

Would it make sense to go with your initial approach of having a special case for list elements?

Fweddi:
> Firstly, superb PR description, so descriptive and helpful, it is very much appreciated 🙏
> 
> I've played around and have verified this has fixed the bug!
> 
> One regression seems to be that you can no longer skip over individual paragraphs:
> 
>  no-para-hop.mp4 
> I think this is because paragraphs are stored as child elements of a `textElement`. So recursion is probably desirable in this case.
> 
> Would it make sense to go with your initial approach of having a special case for list elements?

Good spot - thanks for this. My solution for this is to allow just the one level of recursion (for the textElement -> paragraph case), but otherwise prevent recursion. This means you won't recurse into a list element's fields, but you also won't recurse into a pullquote's fields, a callout's fields and so on. I think this is acceptable and fixes all the movement cases.

simonbyford:
> > Firstly, superb PR description, so descriptive and helpful, it is very much appreciated 🙏
> > I've played around and have verified this has fixed the bug!
> > One regression seems to be that you can no longer skip over individual paragraphs:
> > no-para-hop.mp4
> > I think this is because paragraphs are stored as child elements of a `textElement`. So recursion is probably desirable in this case.
> > Would it make sense to go with your initial approach of having a special case for list elements?
> 
> Good spot - thanks for this. My solution for this is to allow just the one level of recursion (for the textElement -> paragraph case), but otherwise prevent recursion. This means you won't recurse into a list element's fields, but you also won't recurse into a pullquote's fields, a callout's fields and so on. I think this is acceptable and fixes all the movement cases (but worth testing thoroughly!).

Thanks for this! I've been able to verify it is now possible to skip over paragraphs, but unfortunately it is no longer possible to do so within list elements:

nested

Fweddi:
Okay, I've reverted back to my original solution, which hard-codes the list element types. Shall we see if this works and then think about any possible refactors?

Hard-coding the list element types is not ideal, as (1) the list element types are defined downstream in Composer, and (2) we would need to remember to add new list element types here when we create them.

rhystmills:
> Okay, I've reverted back to my original solution, which hard-codes the list element types. Shall we see if this works and then think about any possible refactors?
> 
> Hard-coding the list element types is not ideal, as (1) the list element types are defined downstream in Composer, and (2) we would need to remember to add new list element types here when we create them.

You might be able to use `anyDescendantFieldIsNestedElementField` to identify the list element nodes - it's in https://github.com/guardian/prosemirror-elements/blob/main/src/plugin/fieldViews/NestedElementFieldView.ts

simonbyford (approved):
Tested locally and everything LGTM 👍 _Conditional recursion_ short-circuits my brain so nice work on finding such an elegant solution and working through the edge cases!

PR #395, opened 2024-05-06
Add basic animations to list elements

simonbyford:
## What does this change?

I've noticed it can be quite difficult to visually keep track of list elements as they are created, deleted or moved around in the editor. These operations happen instantaneously, it felt like some animations would help a lot.

I experimented with CSS solutions (/hacks) to this problem without any success, then I came across `@formkit/auto-animate` which seemed to promise exactly what we wanted:

https://auto-animate.formkit.com

You specify a container element and it adds basic transitions to its direct descendants. 

In my view this improves the usability of the editor:

| Before | After |
|--------|--------|
| before2  | after2 |

Unfortunately I wasn't able to use the recommended React implementation of `@formkit/auto-animate`, I left a [comment](https://github.com/guardian/prosemirror-elements/blob/00390275c789bce31dc8fbdea74fd0dc91b8c80b/src/hooks/useAutoAnimate.ts#L1-L11) explaining this.

## How to test

Via the demo (`yarn start`) or linking with composer using `yarn yalc`.

Is it easier to keep track of list elements when they are added/removed/moved?


rhystmills:
Incredible work!

Fweddi:
This looks awesome! Perhaps if this is a success we could roll it out to all element types?

PR #400, opened 2024-05-14
Update inner selection when outer selection updates selection within a text field

rhystmills:
## Background
In prosemirror-elements, elements are a collection of fields representing user-editable data. Most logic managing these fields is in the relevant `FieldView`. Text fields are all based on `ProseMirrorFieldView`, via implementations that extend it, e.g. `TextFieldView`. These contain instances of ProseMirror ('inner editors'), and the 'outer editor' (the editor that contains the element) delegates control of certain actions to the 'inner editor' managed by the FieldView.

Currently, two kinds of changes from the outer editor can cause changes in the field view:
  1. Changes to the nodes that the inner editor manages:
  2. Changes to `Decorations` that affect the range covered by the field view.

However, if the selection is updated by the outer editor, the selection within a relevant field is not updated.

This hasn't been obvious in the past, because usually the selection is updated by the field's 'inner' editor. With notes being supported in nested element fields, when a user hits 'F10' in Composer with text selected inside the nested element, the selection is updated by the outer editor (to which the menu belongs), and should be moved to the end of the note as users expect. Because the selection is not passed to the inner editor, the note content remains selected, so when the user types, they overwrite the content of the note.

## What does this change?

This PR updates field views when the selection changes, and the text fields (extensions of `ProsemirrorFieldView`) update their inner editor selection to match the incoming selection. This means that the usual F10 behaviour for notes also works inside our test fields.

Kapture 2024-05-14 at 15 49 59

This happens as follows:
1. The outer selection changes
2. Update` in `plugin.ts`, a function at the boundary of the plugin, receives state changes from the outer editor, which included the updated `Selection`.
3.  We pass this into the `FieldViews` via some intermediate functions.
4. The only `FieldView` that needs to handle the selection change is the abstract `ProseMirrorFieldView`, which all text `FieldViews` extend.
5. Every text `FieldView` checks if the selection covers its own range. If it does, it updates its internal selection.

## How to test
1. Use prosemirror-elements locally with flexible-content using `yalc`.
6. Create a list format document, like Key Takeaways
7. Type some text in the nested element field, highlight some of it, and hit F10. The cursor should move to the end of the note.

**n.b. - due to an unrelated bug with menu clicks causing the editor to lose focus, this only works with the F10 shortcut, not the notes menu button**

I have updated existing tests to match the new update behaviour and added a test to reflect the selection behaviour itself, making sure that the inner selection is properly updated when there is a relevant incoming selection.

### DIFF
--- a/src/plugin/fieldViews/ProseMirrorFieldView.ts
+++ b/src/plugin/fieldViews/ProseMirrorFieldView.ts
@@ -1,6 +1,6 @@
-import { AttributeSpec, Node } from "prosemirror-model";
 import { DOMParser } from "prosemirror-model";
-import { Plugin, Selection, TextSelection, Transaction } from "prosemirror-state";
+import type { AttributeSpec, Node } from "prosemirror-model";
+import type { Plugin, Selection, Transaction } from "prosemirror-state";
 import { EditorState } from "prosemirror-state";
 import { Mapping, StepMap } from "prosemirror-transform";
 import type { DecorationSource, EditorProps } from "prosemirror-view";
@@ -200,39 +200,42 @@ export abstract class ProseMirrorFieldView extends FieldView<string> {
       const incomingHeadPos = selection.$head.pos;
 
       const fieldStart = this.offset + this.getPos() + 2;
-      const fieldEnd = this.offset + this.getPos() + this.innerEditorView.state.doc.nodeSize;
+      const fieldEnd =
+        this.offset + this.getPos() + this.innerEditorView.state.doc.nodeSize;
 
-      const selectionAnchorIsWithinThisField = incomingAnchorPos > fieldStart && incomingAnchorPos < fieldEnd;
-      const selectionHeadIsWithinThisField = incomingHeadPos > fieldStart && incomingHeadPos < fieldEnd;
-          
+      const selectionAnchorIsWithinThisField =
+        incomingAnchorPos > fieldStart && incomingAnchorPos < fieldEnd;
+      const selectionHeadIsWithinThisField =
+        incomingHeadPos > fieldStart && incomingHeadPos < fieldEnd;
+
       if (selectionAnchorIsWithinThisField && selectionHeadIsWithinThisField) {
         // The inner editor's selection will be offset relative to the start of this field, compared to the incoming selection
-        const currentAnchorPos = this.innerEditorView.state.selection.$anchor.pos;
+        const currentAnchorPos = this.innerEditorView.state.selection.$anchor
+          .pos;
         const currentHeadPos = this.innerEditorView.state.selection.$head.pos;
         const offsetIncomingAnchorPos = incomingAnchorPos - fieldStart;
         const offsetIncomingHeadPos = selection.$head.pos - fieldStart;
 
-        if (currentAnchorPos !== offsetIncomingAnchorPos || currentHeadPos !== offsetIncomingHeadPos){
-          const offsetMap = StepMap.offset(
-            -fieldStart
-          );
+        if (
+          currentAnchorPos !== offsetIncomingAnchorPos ||
+          currentHeadPos !== offsetIncomingHeadPos
+        ) {
+          const offsetMap = StepMap.offset(-fieldStart);
           const mappedSelection = selection.map(state.tr.doc, offsetMap);
 
           this.innerEditorView.dispatch(
             state.tr
-              .setSelection(
-                mappedSelection
-              )
+              .setSelection(mappedSelection)
               .replace(
                 diffStart,
                 endOfInnerDiff,
                 node.slice(diffStart, endOfOuterDiff)
               )
               .setMeta("fromOutside", true)
-          )
-          return
+          );
+          return;
         }
-      };
+      }
     }
     this.innerEditorView.dispatch(
       state.tr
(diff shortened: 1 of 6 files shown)


src/plugin/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -145,62 +147,98 @@ export abstract class ProseMirrorFieldView extends FieldView<string> {
...
+          currentAnchorPos !== offsetIncomingAnchorPos ||
+          currentHeadPos !== offsetIncomingHeadPos
+        ) {
+          const offsetMap = StepMap.offset(-fieldStart);
+          const mappedSelection = selection.map(state.tr.doc, offsetMap);
+          shouldDispatchTransaction = true;
+          tr = tr.setSelection(mappedSelection);
+        }
+      }
+    }
```
simonbyford, line 190: I've tried refactoring this a bit by adding a couple of comments and attempting to simplify the `if` logic, removing the nested one. I don't know if it's any more readable now. `incomingSelectionDiffersFromCurrentSelection` is pretty long...

Anyway - feel free to ignore this suggestion, just food for thought.

```suggestion
    if (selection) {
      // Absolute positions of the incoming selection
      const incomingAnchorPos = selection.$anchor.pos;
      const incomingHeadPos = selection.$head.pos;

      // Relative positions of the current selection in the inner editor
      const currentAnchorPos = this.innerEditorView.state.selection.$anchor.pos;
      const currentHeadPos = this.innerEditorView.state.selection.$head.pos;

      // Absolute position of the field in the document
      // Note: we must offset to account for a few things:
      //  - getPos() returns the position directly before the parent node (+1)
      //  - the node we will be altering is a child of its parent (+1)
      const contentOffset = 2;
      const fieldStart = this.offset + this.getPos() + contentOffset;
      const fieldEnd =
        this.offset + this.getPos() + this.innerEditorView.state.doc.nodeSize;

      const incomingSelectionIsWithinThisField =
        incomingAnchorPos > fieldStart && incomingHeadPos < fieldEnd;

      // The inner editor's selection will be offset relative to the start of this field,
      // compared to the incoming selection
      const incomingSelectionDiffersFromCurrentSelection =
        currentAnchorPos !== incomingAnchorPos - fieldStart ||
        currentHeadPos !== incomingHeadPos - fieldStart;

      if (
        incomingSelectionIsWithinThisField &&
        incomingSelectionDiffersFromCurrentSelection
      ) {
        const offsetMap = StepMap.offset(-fieldStart);
        const mappedSelection = selection.map(state.tr.doc, offsetMap);
        shouldDispatchTransaction = true;
        tr = tr.setSelection(mappedSelection);
      }
    }
```

simonbyford (approved):
This looks good and seems to work fine locally! One suggestion you are free to ignore.

I appreciate the time taken to go over this with me 👍

src/plugin/fieldViews/ProseMirrorFieldView.ts
rhystmills: This is much nicer, cheers!

src/plugin/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -145,62 +147,99 @@ export abstract class ProseMirrorFieldView extends FieldView<string> {
...
-    //
-    // This happens because `findDiffEnd` finds the first point at which the content differs,
-    // starting from the end of the nodes. So if we encounter identical content, the diff will
-    // be shorter by the length of the identical content we encounter – or the overlap between
-    // the two nodes from the point of view of the diff.
-    const overlap = diffStart - Math.min(endOfOuterDiff, endOfInnerDiff);
-    if (overlap > 0) {
-      endOfOuterDiff += overlap;
-      endOfInnerDiff += overlap;
+    if (shouldDispatchTransaction) {
```
jonathonherbert, line 238: More a note to self: shame we can't ask the transaction itself if it's changed! (There's a `this.updated` field, but it's private.) We might be able to avoid this state with ('transaction selection has changed' && 'transaction.steps.length > 0), but it's a bit of a weird one as it relies on the dev knowing that `.setSelection` mutates the selection, and `.replace` adds steps. So this is fine.

src/plugin/plugin.ts
```diff
@@ -280,16 +283,23 @@ const createNodeView = <
...
-        // Only update our FieldViews if their content or decorations have changed.
+        // Only update our FieldViews if their content or decorations have changed, or the selection has changed.
         // nestedElement FieldViews are always updated as a workaround for a bug
         // with merging text elements in the zipRoot plugin after an intermediate element is
         // deleted.
         if (
           fieldValuesChanged ||
           innerDecosChanged ||
-          anyDescendantFieldIsNestedElementField(newNode)
+          anyDescendantFieldIsNestedElementField(newNode) ||
```
jonathonherbert, line 293: Do we have any tests for this logic?

jonathonherbert (approved):
Looks great, top work 👏 Two non-blocking questions about tests.

jonathonherbert:
(hah – too slow!)

rhystmills:
Whoops, apologies!

jonathonherbert:
(There was already an approval, I was just sticking my oar in ⛵ 👍)

rhystmills:
> (There was already an approval, I was just sticking my oar in ⛵ 👍)

I'll try and address these in another PR.

PR #403, opened 2024-05-20
Fix issues with decorations in nested and repeater fields

jonathonherbert:
_co-authored-by: @rhystmills_

## What does this change?

#349 fixes a problem with decorations in `NestedFieldView` elements, but introduces an issue with repeater fields – the problem is illustrated in 14586e2f7524c60817a73083fe40b021bf59ee4e.

The likely cause of this problem is that prosemirror-elements has always passed decorations directly to child editor instances without altering their depth.

This seems to have worked while the decorations and the document have been in sync – by mapping Decorations to ensure that the relevant field always indexes into the correct position, child editors pick up the correct decoration set and ignore other decorations. (I think this is an artefact of the way Decorations are structured, as they are indexed with a start and end position, and as long as these line up with the relevant `Node`, things work as they normally would.)

But as we traverse the boundary between parent and child editors supporting elements, as we do in `NestedElementFieldView`, we introduce a shift in the document structure that introduces out-of-bounds errors. A detailed write up of this case to come.

The solution is likely to be to only pass exactly the decorations that are needed to each view, building on the useful API that was introduced in implementing #349.

## How to test

- [ ] The automated tests should pass. They should cover nested, repeated, and nested repeated/repeated nested elements.

Problems with decorations in complex documents are resolved downstream.

### DIFF
--- a/cypress/helpers/editor.ts
+++ b/cypress/helpers/editor.ts
@@ -1,3 +1,4 @@
+import type { insertElement } from "../../demo";
 import type { WindowType } from "../../demo/types";
 import { getFieldHeadingTestId } from "../../src/editorial-source-components/DemoInputHeading";
 import { placeholderTestAttribute } from "../../src/plugin/helpers/placeholder";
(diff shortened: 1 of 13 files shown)


jonathonherbert:
Closed in favour of #424.

PR #406, opened 2024-05-30
Fix recently introduce bug affecting elements

rhystmills:
## Background
Developers and users have noticed some unusual behaviour in elements recently:
- Attempting to switch the position of two elements of the same type did not appear to update their text fields, so the elements didn't appear to switch. 
- Certain changes to text fields in image elements appeared garbled/disordered in preview after save and close
- Saving and closing then reopening list formats with an image element in them caused placeholders not to appear, and certain changes weren't being saved

Through testing we identified that version `9.6.1` (the most recent release of prosemirror-elements) introduced the change. The biggest and most relevant change in this release was https://github.com/guardian/prosemirror-elements/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed

That change was reverted in Composer and the issues seem to be resolved there now: https://github.com/guardian/flexible-content/pull/4803

## What does this PR change

This PR is intended to solve the bugs reported above.

The bug was caused by a subtle change in our logic during the refactor in the [selection PR](https://github.com/guardian/prosemirror-elements/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed).

Previously, we would check for a 'diff' between the previous state of the node represented by a text view, and its updated state, as follows:

```js
if (diffStart === null) {
      return this.maybeRerenderDecorations();
    }
```

I had changed this check to:
```js
    if (diffStart && diffEnd) {

```

However, a diff could often start at position 0 relative to the `innerEditorView` represented by the text field, if the entire text field had changed, e.g. as part of wider changes to an entire element.

Changing this check to `if (diffStart !== null && diffEnd)` seems to have resolved the issues mentioned above when testing locally.

## Accessibility

### DIFF
--- a/src/plugin/fieldViews/ProseMirrorFieldView.ts
+++ b/src/plugin/fieldViews/ProseMirrorFieldView.ts
@@ -197,7 +197,7 @@ export abstract class ProseMirrorFieldView extends FieldView<string> {
     const diffStart = node.content.findDiffStart(state.doc.content);
     const diffEnd = node.content.findDiffEnd(state.doc.content);
 
-    if (diffStart && diffEnd) {
+    if (diffStart !== null && diffEnd) {
       let { a: endOfOuterDiff, b: endOfInnerDiff } = diffEnd;
       // This overlap accounts for a situation where we're diffing nodes where we encounter
       // identical content.
--- a/src/plugin/__tests__/plugin.spec.tsx
+++ b/src/plugin/__tests__/plugin.spec.tsx
@@ -85,13 +85,14 @@ describe("createPlugin", () => {
     createEditorWithSingleElementPresent(defaultEditorOptions);
 
   describe("Response to content changes", () => {
-    it("should not update consumers or fieldViews when the element content has not changed", () => {
+    it("should not update consumers or fieldViews when the element content and selection have not changed", () => {
       const { view, exampleText } = createDefaultEditor();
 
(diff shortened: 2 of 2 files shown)


src/plugin/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -197,7 +197,8 @@ export abstract class ProseMirrorFieldView extends FieldView<string> {
     const diffStart = node.content.findDiffStart(state.doc.content);
     const diffEnd = node.content.findDiffEnd(state.doc.content);
 
-    if (diffStart && diffEnd) {
+    // Check for null specifically, rather than falsiness, because a diff starting at pos 0 is a valid diff
+    if (diffStart !== null && diffEnd) {
```
davidfurey, line 201: Good spot! Might be nice to also check diffEnd is not `null`? Who knows, maybe at some point negative positions will be valid, and `diffEnd` could be zero but still greater than `diffStart`.

Georges-GNM (approved):
Tested locally and all the desired behaviour seemed to be working 🥳

src/plugin/fieldViews/ProseMirrorFieldView.ts
rhystmills: Done, `!== null` does seem like what we really want to check for there.

PR #408, opened 2024-05-31
Lint strict boolean expressions

davidfurey:
## What does this change?

Disallows numbers being coerced into boolean expressions. It is really easy to introduce subtle bugs by testing for non-nullness and inadvertently also  excluding `0`. Hopefully being more explicit about `null` and `0` also makes the intent of the code more clear.

`@typescript-eslint/strict-boolean-expressions` allows us to selectively disallow specific types from coercion to `boolean`. To start with, I've just excluded numbers, since these are typically the highest risk of introducing bugs. But it would be nice in future to expand this to exclude strings and objects too.

## How to test / review

Please carefully check every change, since it is easy to switch a `!variable` into a `variable !== undefined` when it should be `variable === undefined`.


src/plugin/helpers/prosemirror.ts
```diff
@@ -175,7 +177,7 @@ const removeNode = (getPos: () => number | undefined) => (
   dispatch: ((tr: Transaction) => void) | false,
   view?: EditorView
 ) => {
-  if (!dispatch) {
+  if (dispatch == false) {
```
simonbyford, line 180:
```suggestion
  if (dispatch === false) {
```

src/plugin/helpers/prosemirror.ts
```diff
@@ -84,7 +84,9 @@ const nextPosFinder = (consumerPredicate: Predicate) => (
       const [nextNode = null, nextNodePos = null] =
         nodesBetween(state, pos, all.to).find(predicate) ?? [];
 
-      return nextNodePos && nextNode && nextNodePos + nextNode.nodeSize;
+      return nextNodePos !== null && nextNode !== null
+        ? nextNodePos + nextNode.nodeSize
+        : null;
```
simonbyford, line 89: This one was particularly cursed, thanks for fixing!

src/elements/helpers/getImageSrc.ts
```diff
@@ -13,7 +13,10 @@ export const getImageSrc = (assets: Asset[], desiredWidth: number) => {
     widthDifference(stringOrNumberToNumber(assetB.fields.width));
 
   const sortedAssets = assets
-    .filter((asset) => !asset.fields.isMaster)
+    .filter(
+      (asset) =>
+        asset.fields.isMaster === false || asset.fields.isMaster === undefined
```
simonbyford, line 18: It looks like `isMaster` can also be a [string](https://github.com/guardian/prosemirror-elements/blob/ef306e9a94a3983452a9bdac5d671843f08224fd/src/elements/helpers/defaultTransform.ts#L13). Should we include another condition?

```suggestion
        asset.fields.isMaster === "false" || asset.fields.isMaster === false || asset.fields.isMaster === undefined
```

simonbyford (approved):

src/plugin/helpers/prosemirror.ts
davidfurey: Checked twice for `==` instead of `===`, but clearly one still got through. Thanks for spotting!

src/elements/helpers/getImageSrc.ts
davidfurey: Not sure if that ever actually happens, but safer to check for it!

PR #410, opened 2024-06-10
Fix selection error when backwards selections cover fields

jonathonherbert:
## What does this change?

Fixes an error where backwards selections that extend beyond a field are passed into the field view, causing range errors.

The errors are largely harmless – they're a no-op on the state and do not halt the outer editor – but they're causing some noise in [our downstream CMS's error reporting.](https://the-guardian.sentry.io/issues/5384237869/?project=26737&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=0&utc=true)

There's a question as to whether fields should be aware of selections that cover them entirely. I don't think we've seen a use case yet, but we could map and clamp the selection values, rather than just ignoring, if we wanted that behaviour.

## How to test

- The new unit tests should pass. The commit history shows the failure, then the pass.
- Try adding a backwards selection across an element. You should not see red ink in the console:

error-selection

### DIFF
--- a/src/plugin/fieldViews/ProseMirrorFieldView.ts
+++ b/src/plugin/fieldViews/ProseMirrorFieldView.ts
@@ -176,7 +176,10 @@ export abstract class ProseMirrorFieldView extends FieldView<string> {
         this.offset + this.getPos() + this.innerEditorView.state.doc.nodeSize;
 
       const incomingSelectionIsWithinThisField =
-        incomingAnchorPos > fieldStart && incomingHeadPos < fieldEnd;
+        incomingAnchorPos > fieldStart &&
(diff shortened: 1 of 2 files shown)


rhystmills:
Think this may need a rebase now but I am happy to review as soon as possible.

jonathonherbert:
Thanks @rhystmills, just done that. (Now we no longer use semantic-release in this library, perhaps it's worth turning off squash and merge, which IIUC makes conflicts more likely.)

rhystmills (approved):
Makes sense as a fix, and all tests passing 👍

PR #417, opened 2024-06-25
Bump prosemirror-transform to solve bug

rhystmills:
## What does this change?
This bumps `prosemirror-transform` to 1.9.0, pulling in a bug fix that has affected our implementation in Composer.

Specifically, transformation to bullet lists doesn't work when it includes the first paragraph in a rich text field.

The underlying bug is mentioned here: https://github.com/ProseMirror/prosemirror-transform/releases/tag/1.9.0

## How to test
Check the tests pass.


simonbyford (approved):
Great find

PR #424, opened 2024-08-14
Fix decorations bug, with inline and widget decorations supported in all text fields

rhystmills:
## What does this change?
Since the introduction of the `nestedElement` field, we have noticed various problems related to decorations. This PR: 

- Aims to solve the decorations problems once and for all. The core fix is in 'ProseMirrorFieldView.ts'. I've removed some now unnecessary code related to the previous attempted fix from 'field.ts'.

- I've added some tests, building on tests from @jonathonherbert in https://github.com/guardian/prosemirror-elements/pull/403. The new element specs and forms ('NestedElement' and 'RepeaterElement') and helpers in 'editor.ts' and 'test.ts' are all related to testing.

- I've added a plugin for testing `Widget` decorations in the demo - any time 'widget' appears in text, it will be marked up with two red line `Widget` decorations on either side. This makes the feedback loop for manually testing `Widget` decorations a bit faster. It's also used in some of the new Cypress integration tests.

(Note - I came across an issue with [Typescript being unable to have union types with 25 or more items in them](https://github.com/microsoft/TypeScript/issues/40803), affecting compilation when we had more than 24 elements in the demo elements list. This already could do with a trim - using a handful of elements that demonstrate library features, rather than replicating everything we have in Composer. I've removed the witness element for now, but we should do a general tidy up. If we do need more than 24, we'll need a different approach for handling the types).

## Background

The most obvious errors occurred when adding elements to `nestedElement` fields within `repeater` fields - usually out of bounds errors, crashing the editor and forcing a page reload and preventing the saving of content.

Inline decorations (e.g. our Typerighter matches) have generally worked better in these contexts than `Widget` decorations (like our prosemirror-noting start/end decorations and placeholder decorations).

There have been a few attempts to fix this problem (e.g. https://github.com/guardian/prosemirror-elements/pull/332), but we hadn't quite got to the bottom of it - so have introduced [workarounds](https://github.com/guardian/flexible-content/pull/4867) in certain places and delayed affected features like a main media field in the Timelines element. The [most recent fix](https://github.com/guardian/prosemirror-elements/pull/349) seemed promising, but it seems to break decorations for `repeater` nodes containing standard flat text fields, and we found a few cases it didn't fix:
- `nestedElement` fields would only work if they were the first field in the `repeater`, alphabetically
- An element or `repeater` couldn't contain more than one `nestedElement` field
- Repeaters containing `nestedElement` fields, within another repeater

I'm quite confident that this PR will fix decorations in these contexts.

I think the problems with decorations were not specifically caused by repeaters or nested elements, but our existing decoration code was not really capable of handling all decorations properly in those contexts. The addition of those new field types exposed more complicated structures for us to apply decorations to, which exposed some of the problems with our existing approach.

We've not really had to handle `Widget` decorations within elements containing content until the introduction of the `nestedElement` field, at which point external placeholders and notes were being added within the context of a prosemirror-elements field - both of which use `Widget` decorations rather than inline decorations.  Adding to this, most fields in most elements have been 'flat text' (containing text without wrapping `p` blocks, sometimes with `br` tags for line breaks) - presenting quite a simple structure to apply decorations to.

I also suspect we are passed `DecorationGroup` more often than before the introduction of those fields.

## How does this PR fix our problem
I've fixed the problems with decorations by rewriting `applyDecorationsFromOuterEditor` in 'ProseMirrorFieldView'. 

### Handling `DecorationGroup` and `DecorationSet`
We sometimes receive a [DecorationGroup](https://github.com/ProseMirror/prosemirror-view/blob/cfa73eb969777f63bcb39972594fd4a9110f5a93/src/decoration.ts#L485) from the outer editor, rather than a [DecorationSet](https://prosemirror.net/docs/ref/#view.DecorationSet).

The type and class for `DecorationGroup` are not made available by `prosemirror-view`, so I've had to add a type declaration for that class. It would be good to make a change upstream so we can import that from the source.

The code now explicitly handles these two options. `DecorationGroup` contains a collection of `DecorationSet` in its `members` property, so we can handle them fairly similarly.

### Mapping decorations from the outer editor (the document) to the inner editor (the field)
This happens in `getMappedDecorationsFromSet` and I've added quite detailed comments to explain what I do there.

In plain text:

1. I find the relevant decorations in the context of the outer editor (the document containing the prosemirror-elements plugin) within the range of the field in that editor.
2. I create a new set of decorations, again in the context of the original document. I've done these two steps because mapping the decorations **may fail** (often silently, without an error) if any of the decorations will fall outside of the valid range during the mapping operation. If so, it gives us a `DecorationSet` where Widget and Node decorations won't be handled properly, though inline decorations might work.
3. I 'map' the decoration set with an offset so that they match the positions in the inner editor (which starts at 0 from the inner editor's perspective).

`DecorationSet` is a custom data structure containing a collection of `Decoration`. The key thing is - when you create one, it needs to be given a document to which the decorations can be logically applied. Inline decorations seem to be quite forgiving - the editor will try to apply them as long as they have relevant positions, but Widget and Node decorations need to map up properly with nodes in the document.

## How to test
With the demo application running (`yarn start`), run the integration tests -  (`test:integration`). There are some new ones that reflect inline and widget decorations in different contexts.

You may also want to run the demo app and test decorations manually. Insert different element types, and type "deco" and "widget" to test inline and widget decorations in different contexts. Do they work everywhere? (They should.)

You may also want to test the plugin in Composer using `yalc`. If you're feeling especially diligent, try renaming the Mini Profiles' elements `miniBio` back to `bio` - and removing the use of its transformer (reverting the additions in [this PR](https://github.com/guardian/flexible-content/pull/4867)). This used to fail due to the decorations bug.

## Images

Example in the context of the demo app, with inline and widget decorations in many contexts:

image

Example Mini Profile, with the workaround removed, working with inline and widget decorations in many contexts:

image


jonathonherbert (commented):
Brilliant. Not only fixes this The Right Way, but the tests look pretty comprehensive and are also super legible.

A few non-blocking comments.

There is one (1) bug I can see – we should see a nested decoration appear in the repeater element for every instance of `deco` when it is instantiated, but we're missing one. Strangely, it reappears once a widget deco is added:

decos

It looks like this is related to the code that is finding the relevant decorations, as adding text to the document changes which decos appear:

decos-mapping

Should the decoration find offset be scoped differently?
- the decoration find offset is scoped to the document (get me the decorations for this part of the document)
- the decoration mapping offset is scoped to the element (map the decorations I am creating to eliminate their offset to the element)

rhystmills:
> Brilliant. Not only fixes this The Right Way, but the tests look pretty comprehensive and are also super legible.
> 
> A few non-blocking comments.
> 
> There is one (1) bug I can see – we should see a nested decoration appear in the repeater element for every instance of `deco` when it is instantiated, but we're missing one. Strangely, it reappears once a widget deco is added:
> 
> decos [ decos ](https://private-user-images.githubusercontent.com/7767575/359044757-58513537-4896-4a5e-96ea-1f899aa95486.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQwNjk5ODEsIm5iZiI6MTcyNDA2OTY4MSwicGF0aCI6Ii83NzY3NTc1LzM1OTA0NDc1Ny01ODUxMzUzNy00ODk2LTRhNWUtOTZlYS0xZjg5OWFhOTU0ODYuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDgxOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA4MTlUMTIxNDQxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NDNkZTZhNTgyOTc4NmMzMWNiMDBiMWI3Mjg5YTY3MjkzOTRkOGE2MTUzZmRhZWNlM2Q4YjA1MzIwYmJkMjhiYyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.3HYDQ-JXKxV4WJSZFh4GSdLKJyp-62Vw6YReykIl4TM) [ ](https://private-user-images.githubusercontent.com/7767575/359044757-58513537-4896-4a5e-96ea-1f899aa95486.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQwNjk5ODEsIm5iZiI6MTcyNDA2OTY4MSwicGF0aCI6Ii83NzY3NTc1LzM1OTA0NDc1Ny01ODUxMzUzNy00ODk2LTRhNWUtOTZlYS0xZjg5OWFhOTU0ODYuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDgxOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA4MTlUMTIxNDQxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NDNkZTZhNTgyOTc4NmMzMWNiMDBiMWI3Mjg5YTY3MjkzOTRkOGE2MTUzZmRhZWNlM2Q4YjA1MzIwYmJkMjhiYyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.3HYDQ-JXKxV4WJSZFh4GSdLKJyp-62Vw6YReykIl4TM)
> 
> It looks like this is related to the code that is finding the relevant decorations, as adding text to the document changes which decos appear:
> 
> decos-mapping [ decos-mapping ](https://private-user-images.githubusercontent.com/7767575/359046834-f17c3517-9e25-49d3-bfc2-89334ffd35ab.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQwNjk5ODEsIm5iZiI6MTcyNDA2OTY4MSwicGF0aCI6Ii83NzY3NTc1LzM1OTA0NjgzNC1mMTdjMzUxNy05ZTI1LTQ5ZDMtYmZjMi04OTMzNGZmZDM1YWIuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDgxOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA4MTlUMTIxNDQxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NzZiNjk0YzU4ODhiZjk3NmFlNDE0NWM3NzQ3YzhlZmJhMDhhM2M1ZDI1OWQxNzZiMDFmYzBhYWI4NmY5NmQ0MSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.sLggHOsy3TDYYnb7yFkiWsarvSz8Wd3lHsGaN0EuCa0) [ ](https://private-user-images.githubusercontent.com/7767575/359046834-f17c3517-9e25-49d3-bfc2-89334ffd35ab.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQwNjk5ODEsIm5iZiI6MTcyNDA2OTY4MSwicGF0aCI6Ii83NzY3NTc1LzM1OTA0NjgzNC1mMTdjMzUxNy05ZTI1LTQ5ZDMtYmZjMi04OTMzNGZmZDM1YWIuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDgxOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA4MTlUMTIxNDQxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NzZiNjk0YzU4ODhiZjk3NmFlNDE0NWM3NzQ3YzhlZmJhMDhhM2M1ZDI1OWQxNzZiMDFmYzBhYWI4NmY5NmQ0MSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.sLggHOsy3TDYYnb7yFkiWsarvSz8Wd3lHsGaN0EuCa0)
> 
> Should the decoration find offset be scoped differently?
> 
> Edit: I've added a failing test for this behaviour in [b42a295](https://github.com/guardian/prosemirror-elements/commit/b42a29535aba0dbbbfece17db077234965b145ad).

Good spot there.

I feel like I tested for this situation earlier on working in this branch, and I suspect I've made it appear during a refactor. For the `DecorationSet`-handling logic, I removed some code where I was recreating the `DecorationSet` in the context of the inner editor's node (the field), rather than using the one from the outer editor (the document). I didn't find any problems at the time of making this refactor, but I was a bit worried it would introduce an issue. Recreating the `DecorationSet` in the context of the field seems to have solved the issue as far as I can see.

I wonder if the reason adding the widget decorations fixes this is because you end up with a `DecorationGroup` instead, where we already have a `DecorationSet` created in the context of the field node.

jonathonherbert:
Thanks for this explanation! I wondered whether it might be because we were searching the set too tightly and excluding results, but couldn't see a problem with the search bounds. That's a satisfying answer 👍

jonathonherbert (approved):

src/plugin/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -345,27 +357,89 @@ export abstract class ProseMirrorFieldView extends FieldView<string> {
...
-    ]);
-    this.decorations = localDecoSet.map(offsetMap, node);
-    this.decorationsPending = true;
+    // The incoming decorations are positioned relative to the parent NodeView, so we only
+    // need to consider the offset of this field from its parent NodeView, and can ignore
+    // the offset of the parent from the outer editor's document.
+    // The node we will be altering is a child of its parent, so we add +1 to the offset.
+    const fieldOffsetFromElement = this.offset + 1;
+
+    if ("find" in decorations) {
```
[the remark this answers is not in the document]
rhystmills: This would be much nicer. Working on it now!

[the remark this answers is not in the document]
rhystmills: (Now done)

src/plugin/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -23,6 +23,18 @@ export interface AbstractTextFieldDescription
   isResizeable?: boolean;
 }
 
+// The type of this class isn't made available by prosemirror-view - we should request a change there to make
```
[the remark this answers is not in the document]
rhystmills: Good idea, have nabbed that.

src/plugin/helpers/decorations.ts
```diff
@@ -0,0 +1,84 @@
+import type { Node } from "prosemirror-model";
+import { Mapping, StepMap } from "prosemirror-transform";
+import type { DecorationGroup, DecorationSource } from "prosemirror-view";
+import { DecorationSet } from "prosemirror-view";
+
+export const getMappedDecorationsFromSet = (
```
davidfurey, line 6: Does this need to be exported?

src/plugin/helpers/decorations.ts
```diff
@@ -0,0 +1,84 @@
...
+  const mappedDecorations = relevantDecosInOriginalDoc.map(
+    offsetMap,
+    fieldNode
+  );
+  return mappedDecorations;
+};
+
+export const getMappedDecorationsFromSource = (
+  decorations: DecorationSource | DecorationSet | DecorationGroup,
+  fieldOffsetFromElement: number,
```
davidfurey, line 46: Would it be helpful to have the same comments as for `getMappedDecorationsFromSet` here?

src/plugin/helpers/decorations.ts
```diff
@@ -0,0 +1,84 @@
...
+  if ("find" in decorations) {
+    // 'decorations' is a DecorationSet. Map them, then set them as the field's decorations.
+    const relevantDecorationSet = getMappedDecorationsFromSet(
+      decorations,
+      fieldOffsetFromElement,
+      fieldNode,
+      document
+    );
+    // Decorations may be lost if we don't recreate the DecorationSet in the context of the innerEditor
+    const decoArray = relevantDecorationSet.find();
```
davidfurey, line 59: `find()` with no parameters returns all decorations as a `Decoration[]`

src/plugin/helpers/decorations.ts
```diff
@@ -0,0 +1,84 @@
...
+      .flatMap((set) => set.find());
+    const decorationsAsSet = DecorationSet.create(
+      fieldNode,
+      relevantDecorations
+    );
+    return decorationsAsSet;
+  }
+  // The DecorationSource should be implemented as a DecorationSet or a DecorationGroup. If it isn't, return an empty
+  // DecorationSet
+  return DecorationSet.create(fieldNode, []);
```
davidfurey, line 83: I believe this is an unexpected code path. If so, should we log this so that we find out (eventually) about new implementations of `DecorationSource` in prosemirror?

src/plugin/helpers/decorations.ts
```diff
@@ -0,0 +1,84 @@
...
+  // The offset of the field from its containing element.
+  fieldOffsetFromElement: number,
+  // The node representing the current field in the ProseMirror document.
+  fieldNode: Node,
+  // The node representing the outer ProseMirror document containing the field.
+  document: Node
+) => {
+  // This field may receive decorations that will not apply to its range.
+  // Find the decorations in the context of the original document that should apply to this field, based on the position of
+  // the decorations in the original field.
```
davidfurey, line 18: I'm not clear on what "based on the position of the decorations in the original field." means

src/plugin/helpers/decorations.ts
```diff
@@ -0,0 +1,84 @@
...
+  // We must filter the decorations before we offset them via a 'map', otherwise there may be errors due to decorations
+  // being mapped outside of a valid range.
+  const relevantDecosFromOriginalDoc = decorationSet.find(
```
davidfurey, line 21: Explicit type annotation `Decoration[]` might make the comment on line 25 more clear

src/plugin/helpers/decorations.ts
```diff
@@ -0,0 +1,84 @@
+import type { Node } from "prosemirror-model";
+import { Mapping, StepMap } from "prosemirror-transform";
+import type { DecorationGroup, DecorationSource } from "prosemirror-view";
+import { DecorationSet } from "prosemirror-view";
+
+export const getMappedDecorationsFromSet = (
+  // A decoration set received, positioned relative to the outer editor. It may contain irrelevant decorations.
+  decorationSet: DecorationSet,
+  // The offset of the field from its containing element.
```
davidfurey, line 9: I'm confused about this, it is the offset of the field from the entire doc, or just from the containing element?

src/plugin/helpers/decorations.ts
```diff
@@ -0,0 +1,84 @@
...
+    fieldOffsetFromElement,
+    fieldOffsetFromElement + fieldNode.nodeSize
+  );
+  // We must recombine the Decoration[] into a DecorationSet because we can only 'map' a DecorationSet, not a Decoration[],
+  // and we want to reposition the decorations relative to the current field.
+  // We must do this in the context of the original document, because ProseMirror uses the structure of the document to structure
+  // the decorations - in particular Widget and Node decorations rely on the document structure for a correct DecorationSet to be
+  // formed.
+  const offsetMap = new Mapping([StepMap.offset(-fieldOffsetFromElement)]);
+  const relevantDecosInOriginalDoc = DecorationSet.create(
```
davidfurey, line 31: `const relevantDecoSetInOriginalDoc`?

src/plugin/fieldViews/ProseMirrorFieldView.ts
```diff
@@ -346,25 +351,29 @@ export abstract class ProseMirrorFieldView extends FieldView<string> {
...
-    // Offset because the node we are displaying these decorations in is a child of its parent (-1)
-    const localOffset = -1;
-    const offsetMap = new Mapping([
-      StepMap.offset(-elementOffset + localOffset),
-    ]);
-    this.decorations = localDecoSet.map(offsetMap, node);
+    // The incoming decorations are positioned relative to the parent NodeView, so we only
+    // need to consider the offset of this field from its parent NodeView, and can ignore
+    // the offset of the parent from the outer editor's document.
+    // The node we will be altering is a child of its parent, so we add +1 to the offset.
```
davidfurey, line 366: todo: ask Jon what this line means

src/plugin/helpers/decorations.ts
davidfurey: The comment is correct. The decorations are already offset from the beginning of the element.

src/plugin/helpers/decorations.ts
davidfurey: todo: Rhys to rewrite

davidfurey (commented):
Amazing work!

This review is a combination of suggestions and my own notes as I tried to understand what was going on

src/plugin/fieldViews/ProseMirrorFieldView.ts
davidfurey: @jonathonherbert - what does "The node we will be altering is a child of its parent" mean?

jonathonherbert: Maybe better to comment on the original comment, which is also a weird gnomic tautology 😅 : _Offset because the node we are displaying these decorations in is a child of its parent (-1)_

That's not clear. IIUC the whole offset situation looks something like:

> We are creating a DecorationSet local to the node that represents our FieldView, so we must add two offsets:
> - an offset for the position difference between the element received by the NodeView and this field (`elementOffset` – probably better called `elementToFieldOffset`, below `e2foffset`),
> - an offset to align the decorations within the child editor, because the node we are rendering in the child editor is now the `doc` node for that editor.

Why does _the node we are rendering in the child editor is now the `doc` node for that editor_ matter?

If the outer editor sees:

```xml
<doc>
    <!-- other nodes -->
    <element-node>
        <!-- maybe other nodes -->
        <field-node>
            <!-- field content -->
        </field-node>
    </element-node>
</doc>
```

which to ProseMirror's positioning system looks like (abbreviated)

```xml
<doc> <!-- snip --> <element-node> <field-node> <!-- field content, etc. -->
     |             |              |            |
     0             e-offset       e2f-offset   e2f-offset + 1
```

The inner editor sees

```xml
<doc> <!-- previously field-node -->
    <!-- field content -->
</doc>
```

which, if it's passed decorations straight from the parent editor, looks like 

```xml
<doc> <!-- field content, etc. -->
     |             
     e2f-offset + 1         
```

That can't be right, because the position after `doc` within an editor is always 0!

So we must subtract `e2f-offset` to account for the offset from element to field, and subtract -1 to account for the fact that as the `field-content` node becomes a `doc` node, it is removed from the positioning system (there is no 'before' `<doc>`.)

This is I think what I meant by the 'child of its parent' comment above – we're collapsing the child/parent position with this offset – but if the above is clearer, happy to add it to the code in a little PR.

jonathonherbert: (Could also add the renaming in that PR, I think the offset names account for some general confusion here.)

davidfurey: Sounds good to me. I prefer `elementToFieldOffset` over `e2foffset`. We don't pay Github per letter.

jonathonherbert: 👍  used `e2foffset` in diagrams b/c it fits in a single line – I'll go to the trouble of adding manual line breaks, but only because it's you, @davidfurey.

PR #441, opened 2025-01-02
Bump webpack-dev-server

rhystmills:
## What does this change?
Bump `webpack-dev-server` to solve vulnerability introduced via http-proxy-middleware

`webpack-dev-server` is a dev dependency used to run local dev servers.

## How to test
1. Run `./script/setup.sh` from the root directory to set up the application, including the nginx mappings that make the local https://prosemirror-elements.local.dev-gutools.co.uk/ url work.
2. Ensure the javascript dependencies are up to date by running `yarn`.
3. Run `yarn start` to start the local server. Does it work?

The dependency is bumped and the dev server still runs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "wait-on": "^5.3.0",
     "webpack": "5.94.0",
     "webpack-cli": "5.1.4",
-    "webpack-dev-server": "5.0.4",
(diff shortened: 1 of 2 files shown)


Fweddi (approved):
Tested locally 👍

PR #443, opened 2025-01-09
Improve performance for command and selection changes

jonathonherbert:
## What does this change?

We recently had a large article — 7000+ words and 242 elements, id `673f39b58f083ae557dd22af` — which was slow to load and edit in production.

Using that article as a test case, this PR improves the performance of the elements plugin in two ways:
- [f62abf6](https://github.com/guardian/prosemirror-elements/pull/443/commits/f62abf6d368e4193c78f303b1e7b3e7a9ef0c8b6): only calculate whether commands are valid once per transaction, storing the result in a plugin and using that to determine whether consumers need to rerender. Thanks @KaliedaRik for suggesting this approach as we paired to figure out what was slow!
- [7076cba](https://github.com/guardian/prosemirror-elements/pull/443/commits/7076cba9560399a73773c779d2b1790b6bb7b264): only update elements on selection change if the selection concerns them (i.e. the selection moves within the node, or moves in or out of the node)

To help measure performance when profiling these changes, [e857a1a](https://github.com/guardian/prosemirror-elements/pull/443/commits/e857a1ab8ef6e64d1025263995d58dafe6fc3dcf) adds a flag to disable performance-heavy features, which is useful for removing noise from serialisation, collab and dev tools code when profiling.

The table below gives the profiler script time for 30 keystrokes with the article above loaded in the demo page, with the above flag disabled. Tested on Chrome 131.0.6778.205  a 14-inch Macbook Pro, M1 Max, 2021:

| Iteration        | 40dded8f9abdc583090819c59bdea7df201db940 | [f62abf6](https://github.com/guardian/prosemirror-elements/pull/443/commits/f62abf6d368e4193c78f303b1e7b3e7a9ef0c8b6) | [7076cba](https://github.com/guardian/prosemirror-elements/pull/443/commits/7076cba9560399a73773c779d2b1790b6bb7b264) |   
|------------------|------------------------------------------|------------------------------------------|------------------------------------------|
|                1 |                                     2224 |                                     1353 |                                      555 |  
|                2 |                                     2230 |                                     1403 |                                      554 |  
|                3 |                                     2381 |                                     1378 |                                      569 |  
| Average          |                              2278.333333 |                                     1378 |                              559.3333333 |  
| ms per keystroke |                              75.94444444 |                              45.93333333 |                              18.64444444 |  

About 4x faster.

## How to test

- Reviewing commit-by-commit should make the changes as clear as possible. I'm really happy to pair on reviewing anything that isn't clear.
- This PR adds unit tests for the selection logic, and updates our rerender tests in https://github.com/guardian/prosemirror-elements/blob/main/src/plugin/__tests__/plugin.spec.tsx to account for the new behaviour. They should pass. Existing integration tests test element wrapper movement.
- Perhaps have a play with that large document on `main` and this branch, by copying and pasting the content from our CMS to the editor playground (`yarn start`.) Can you feel the difference? You can also test in Composer following the instructions in the readme — I've included examples of both in GIFs here:

|Location|Before|After|
|-|-|-|
|Sandbox|pme-slower|pme-faster|
|Composer|composer-slower|composer-faster|

Less pain editing large articles.

## Is there anything more we can do?

On modern hardward, large articles with large numbers of elements like the above are still slow to load (something this PR does not attempt to address), and still have appreciable jank.

It looks like iterating over decorations, which we use to ensure nodeViews always update, takes up a fair few cycles, too. That might be another good place to start looking for performance improvements.

Screenshot 2025-01-09 at 14 12 40

Loading improvements look like they're connecting to rendering. A few strategies we could try:
- Optimise the rendering of individual components.
- Optimise the rendering of components generally, perhaps by avoiding a React root per-element and mounting elements in [portals](https://react.dev/reference/react-dom/createPortal) instead.
- Implement a strategy for virtualising element rendering, so offscreen elements are not rendered until they're needed.

Addresses #444.


KaliedaRik (approved):
Tested in my local:
+ ran `yarn test:unit` - all tests passed
+ ran `./script/setup.sh` to install bumped dependencies, then `yarn start` to test sandbox - things seem to work

Note: have not tested in Composer as (I assume) that requires bumping the `prosemirror-elements` dependency in the `flexible-content` repo and I'm not sure how to achieve that locally before this PR merges into `main` 

Given the assumption that this will work as expected in Composer, happy to approve.

jonathonherbert:
> Note: have not tested in Composer as (I assume) that requires bumping the prosemirror-elements dependency in the flexible-content repo and I'm not sure how to achieve that locally before this PR merges into main

There's a bit in the main readme to facilitate testing in Composer that I mention above (very briefly — easy to miss!), but I'm quite confident in the test suite for this library, and I did some testing in Composer myself for this PR per the GIFs. I think we're fine to merge and test the new version in Composer when it's available.

PR #445, opened 2025-01-21
A few naming tweaks

jonathonherbert:
## What does this change?

Tweaks a few names after pairing with @KaliedaRik reminded me that some of our naming conventions aren't very clear:

- `updator` -> `updater` (just a typo)
- `validateWithFieldAndElementValidators` -> `createElementValidator`
- `createUpdator `-> `createUpdateElementViewFn` (a typo, and hopefully this is more descriptive)
- `render` -> `initElementView`
- `type Renderer` -> `type InitElementView`

Hopefully introducing the idea of an `ElementView`, the part of the element definition responsible for creating and updating its UI, makes this part of the code clearer.

As part of this work, I've refactored the `createUpdater` function, which implements a pub/sub pattern for state updates. It's now a class, ElementStateUpdatePublisher, which I hope better illustrates both its stateful nature, and its job.

## How to test

This should be a no-op. Automated tests should pass.

Marginally less confused developers when pairing on this codebase 😀

### DIFF
--- a/src/plugin/elementSpec.ts
+++ b/src/plugin/elementSpec.ts
@@ -16,22 +16,20 @@ type Subscriber<FDesc extends FieldDescriptions<string>> = (
   isSelected: boolean
 ) => void;
 
-type Updater<FDesc extends FieldDescriptions<string>> = {
-  update: Subscriber<FDesc>;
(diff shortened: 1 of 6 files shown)


KaliedaRik (approved):
Sensible changes. Approved!

PR #447, opened 2025-01-21
Do not render decorations unless necessary

jonathonherbert:
## What does this change?

At the moment, we add a decoration for every parent element node each time the document changes, to force our element NodeViews to rerender.  Adding decorations is the approach recommended by the library author [(here)](https://discuss.prosemirror.net/t/force-nodes-of-specific-type-to-re-render/2480/10).

Normally, NodeViews would only rerender when their content, internal selection or decorations had changed — but we need to respond to all document changes, regardless of whether they touch the relevant node, to update the element movement UI, the buttons of which are disabled when movement is not possible.

However, forcing rerenders for each element on each change isn't ideal, because in documents with large numbers of elements, ProseMirror has to manage large numbers of decorations, and subsequently call `update` for large numbers of nodes. This has a measurable effect on performance (see below.)

This PR uses the recent work optimising command validity detection in #443 to optimise decoration rendering, using the valid ranges to render decorations just for elements that fall outside valid element insertion ranges.

Both adding and removing decorations force a re-render: elements that fall outside the valid insertion range receive a decoration and rerender, and elements moving inside that range lose their decoration, which also forces a rerender.

We recently had a large article, id 673f39b58f083ae557dd22af, which was slow to load and edit in production. The table below gives the profiler script time for 30 keystrokes with the article above loaded in the demo page, with the `enableExpensiveFeatures` flag disabled. Tested on Chrome 131.0.6778.205 on a 14-inch Macbook Pro, M1 Max, 2021:

|Iteration|fdb8213c6eec6f9ba068c22bf9ff9e1f76c42595|2f045d0aa1ca1be2643aeb043854ca504a6e84cc
--|--|--
1|399|463
2|400|495
3|403|482
Average|401|479

About 1.2x faster.

## How to test

- The automated tests should pass. Our integration tests for element movement should test the relevant cases. I've added an additional check for the button states, to make sure they're enabled/disabled as necessary.
- Have a play with the element movement buttons. Do they become enabled/disabled as expected?

Improved performance in large documents with many elements.

### DIFF
--- a/.changeset/sixty-pumpkins-jam.md
+++ b/.changeset/sixty-pumpkins-jam.md
@@ -1,0 +1,5 @@
+---
+"@guardian/prosemirror-elements": patch
+---
+
+Do not render element decorations unless necessary, improving performance in large documents
--- a/cypress/helpers/editor.ts
+++ b/cypress/helpers/editor.ts
@@ -74,8 +74,7 @@ export const typeIntoElementField = (fieldName: string, content: string) =>
   getElementRichTextField(fieldName).focus().type(content);
 
 export const getArrayOfBlockElementTypes = () => {
-  // eslint-disable-next-line prefer-const -- it is reassigned.
-  let elementTypes = [] as string[];
+  const elementTypes = [] as string[];
   return new Cypress.Promise((resolve) => {
     cy.get(".Editor > .ProseMirror-menubar-wrapper > .ProseMirror")
       .children()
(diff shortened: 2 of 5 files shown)


cypress/downloads/downloads.html
simonbyford, line 1: Does this file serve a purpose?

simonbyford (approved):
LGTM, I've played around in the demo page and everything appears to work fine 🚢

cypress/downloads/downloads.html
jonathonherbert: No! Thanks for spotting.

PR #455, opened 2025-07-01
fix: Preserve whitespace in rich text nodes on parse

jonathonherbert:
_co-authored-with: @SHession_

## What does this change?

At the moment, we're dropping whitespace when we parse elements per HTML's rules ([handy MDN docs.](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace)) This puts us in a situation where we can persist whitespace into elements fields, but cannot remove it — because it's not present to remove.

This PR passes `preserveWhitespace: true` to the ProseMirror methods responsible for parsing element data into ProseMirror nodes, which does what it says on the tin. It was previously set to `false` as what I suspect was an unintended default in the PR that changed it (https://github.com/guardian/prosemirror-elements/pull/129).

## How to test

- The new unit test should pass.
- Open a document in the sandbox, add an element, and add whitespace. Refresh the page. The whitespace should remain:

pm-after

### DIFF
--- a/src/plugin/helpers/prosemirror.ts
+++ b/src/plugin/helpers/prosemirror.ts
@@ -324,19 +324,6 @@ const createParsers = <S extends Schema>(schema: S) => {
   return { parser, serializer };
 };
 
-const docToHtml = (serializer: DOMSerializer, doc: Node) => {
-  const dom = serializer.serializeFragment(doc.content);
-  const e = document.createElement("div");
-  e.appendChild(dom);
-  return e.innerHTML;
-};
-
-const htmlToDoc = (parser: DOMParser, html: string) => {
-  const dom = document.createElement("div");
-  dom.innerHTML = html;
-  return parser.parse(dom, { preserveWhitespace: "full" });
-};
-
 // Select all text within a node, as opposed to AllSelection
 const selectAllText = (
   state: EditorState,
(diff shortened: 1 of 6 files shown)


src/plugin/helpers/prosemirror.ts
```diff
@@ -334,7 +334,7 @@ const docToHtml = (serializer: DOMSerializer, doc: Node) => {
 const htmlToDoc = (parser: DOMParser, html: string) => {
   const dom = document.createElement("div");
   dom.innerHTML = html;
-  return parser.parse(dom);
+  return parser.parse(dom, { preserveWhitespace: true });
```
SHession, line 337: Should this be "full" too?

SHession (approved):
Tested this locally on a few elements, worked as expected!

src/plugin/helpers/prosemirror.ts
jonathonherbert: Yes! [091e816](https://github.com/guardian/prosemirror-elements/pull/455/commits/091e816fd125c253bf9dde53b2a6ee72896a7632)

jonathonherbert: Ah, there was a reason — explained in [65faa1b](https://github.com/guardian/prosemirror-elements/pull/455/commits/65faa1b8c3cc0ee65c59faccbe21ef625d326668), I've moved the parser helpers out of library code and added a line to the commit to explain why this is necessary.

Issue #279, opened 2022-11-01
Caret traversal should stop at external and internal element boundaries


paperboyo:
In Chrome it’s possible to traverse the caret via keyboard through all this:

![sdfsdfsdfsdf](https://user-images.githubusercontent.com/6032869/199266674-9cde1663-1883-4fce-9491-d7e132bb3934.gif)

While it may seem useful at first:
- I’m not sure it’s actually correct
- doesn’t happen in Firefox

I think tab should be the only way to get to the next editable “island”. Some apps use Alt+Enter, but not sure about that.

Something to ponder if nothing more. At least the inconsistency between the browsers should not happen.

PR #458, opened 2025-07-04
Create commands to restrict caret traversal inside a textblock

andrewHEguardian:
## What does this change?

Attempts to address https://github.com/guardian/prosemirror-elements/issues/279 - by preventing the caret traversing outside of the boundaries of text blocks. Previously (in Chrome) navigating with the arrow keys would take the caret outside of the bounds of the current text block and into adjacent blocks.

This is implemented as a set of key mapping commands. I've only applied it to `TextFieldView` to validate the concept, although it applies to other elements.

### Before

https://github.com/user-attachments/assets/18d79413-70f1-42fd-a797-fa92a5fc1d76

### After

https://github.com/user-attachments/assets/9f88261c-6861-4332-8ce0-f3ac591cffe9

## How to test

Add a new element, eg Image, and enter the cursor into a plain text block. You should not be able to leave the bounds of that block by using arrow keys, only by tabbing.

### DIFF
--- a/.changeset/old-peas-do.md
+++ b/.changeset/old-peas-do.md
@@ -1,0 +1,5 @@
+---
+"@guardian/prosemirror-elements": minor
+---
+
+Limit caret traversal with arrow keys to within boundaries of text blocks


andrewHEguardian:
I'm also very unfamiliar with prosemirror(-elements) so any input is very appreciated!

> If I understand right, this means the up arrow in some fields where the contents are wrapped over multiple lines will go straight to the start of the field. It’d be nicer for the user if it went up a line (and to the start if it’s the first line)

This is the case 👍  video might not have been clear

> and retained their previous left-right position so that pressing up followed by down always took them back to the previous position. 

Ah, I hadn't even realised that! That's a great spot

src/plugin/helpers/keymap.ts
```diff
@@ -26,3 +27,60 @@ export const createHistoryCommands = (
...
+  },
+  ArrowUp: (
+    state: EditorState,
+    dispatch?: (tr: Transaction) => void,
+    view?: EditorView
+  ) => {
+    if (view?.endOfTextblock("up", state)) {
+      // Move the cursor to the start of this block
+      const tr = state.tr.setSelection(
+        TextSelection.between(state.doc.resolve(0), state.doc.resolve(0))
```
jonathonherbert, line 61: This might be a little clearer as:

```suggestion
        TextSelection.atStart(state.doc)
```

(there's `atEnd` for the other case, too)

jonathonherbert (commented):
Love love love this! @emdash-ie has covered the product things. Great job on finding the appropriate parts of the PM API to achieve this.

As @emdash-ie mentions, preserving the state necessary to replicate the existing caret selection memory when moving up against a document boundary and then back down a line will be challenging. Feeling is that we either agree to keep the behaviour in this PR as-is, or keep the previous behaviour. @paperboyo may have thoughts.

A few non-blocking suggestions below. Before we ship, could we add tests for each case (up, down, left, right against boundaries)? We've integration tests that cover selection behaviour @ https://github.com/guardian/prosemirror-elements/blob/862ca6fc3be40bb5ad94c375c487b67accffa373/cypress/tests/ImageElement.cy.ts#L474 — be super happy to pair if they're proving tricky to write.

paperboyo:
Hi. Thanks all for working on this! Chatted to @emdash-ie a tad and IIUC, this should be fine when fixed as proposed. `Up` from middle of second line to the middle of first line, then `Up` again to beginning and then `Down` to the beginning of second line all sounds correct.

andrewHEguardian:
I've realised this doesn't prevent moving from a paragraph to the text fields (where you can't move back out agian with keyboard). IIUC the paragraph bit would be out of scope for this so we couldn't apply the keymapping there?

https://github.com/user-attachments/assets/a589aa62-b005-48bf-9d4a-b9a09696d40a

jonathonherbert (approved):

andrewHEguardian:
> Any objections to me merging this, @andrewHEguardian?

please do!

PR #472, opened 2026-03-25
Migrate to trusted publishing

SHession:
## What does this change?

Attempt to bump Pan Domain Node to trusted publishing. Following:
- https://github.com/guardian/pan-domain-node/pull/52
- https://github.com/guardian/workflow/issues/1285
- https://github.com/guardian/anghammarad/pull/325

## How to test

Testable by going through the release process and checking the results

We are able to successfully publish package updates.


emdash-ie:
The DevX guide quoted in https://github.com/guardian/workflow/issues/1285 suggests we should add `id-token: write` to the release workflow, and (possibly?) that we should change the url field to a full object – do we need to do those things here?

emdash-ie (approved):
Looks good except for my questions above!

SHession:
> The DevX guide quoted in [guardian/workflow#1285](https://github.com/guardian/workflow/issues/1285) suggests we should add `id-token: write` to the release workflow, and (possibly?) that we should change the url field to a full object – do we need to do those things here?

Good point about the permissions, I have updated this. 

It looks like the way we have set up repository is an [accepted shortcut](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#repository), so I expect it will be fine. It's easy to change if not. 

EDIT: this did show an error so I've changed it: https://github.com/guardian/prosemirror-elements/actions/runs/23584205609/job/68673854803#step:7:36

PR #479, opened 2026-05-06
nested elements can supply own plugins

dblatcher:
## What does this change?

Updated the constructor for the `NestedElementFieldView` to take an optional `Plugin` array. This will allow the Field to have plugins in addition to those inherited from the outer editor (the view which the`NestedElementFieldView` is in).

The main motivation for this change is to unblock migration work on Composer - when defining elementsSpecs in our library, we do not have access to the `pluginKey` of the plugins defined in Composer so cannot include them in `allowedPlugins`, but there may be other application where a `NestedElementFieldView` may need plugins with the rest of the document may not need.

Also exports some extra types  currently imported in the Composer Source code from the dist folder.

## How to test

TBC

Provides a more flexible constructor `NestedElementFieldView` 

This should be safe as only providing an extra option.

### DIFF
--- a/.changeset/perky-coins-kiss.md
+++ b/.changeset/perky-coins-kiss.md
@@ -1,0 +1,5 @@
+---
+"@guardian/prosemirror-elements": minor
+---
+
+NestedElementFieldView constructor accepts optional list of plugins
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,5 @@
 export type { FieldValidator, ValidationError } from "./plugin/elementSpec";
+export type { ExternalElementData } from "./plugin/helpers/element";
 export type { FieldNameToValueMap } from "./plugin/helpers/fieldView";
 export type {
   FieldDescriptions,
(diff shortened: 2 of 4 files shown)


jonathonherbert:
My initial take on this is that we shouldn't be expanding the surface area of the API here if this can be solved in the consuming code. Is it possible to do that — for example, by moving the plugin keys into flexible-frontend (where they presumably will live anyhow, eventually?)

The exported types are a great shout!

dblatcher:
> My initial take on this is that we shouldn't be expanding the surface area of the API here if this can be solved in the consuming code. Is it possible to do that — for example, by moving the plugin keys into flexible-frontend (where they presumably will live anyhow, eventually?)
> 
> The exported types are a great shout!

Makes sense - we could define and export the plugin key in flexible-frontend, but still define the plugin in Composer. I'll give that a try.

We have a somewhat artificial problem in that defining some code  like the plugins in flexible-frontend and importing into Composer will work in the application itself, but will cause the jest-based tests to fail (described in https://github.com/guardian/flexible-content/pull/6544 - TLDR I think its because jest doesn't cope with imported code referencing the Document, but haven't dug too deep into that) - we will ultimately solve this by migrating the whole of the editor (including the test coverage, either with jest or something else), but we have a few awkward cases of needing to duplicate code in the meantime.

jonathonherbert:
> > My initial take on this is that we shouldn't be expanding the surface area of the API here if this can be solved in the consuming code. Is it possible to do that — for example, by moving the plugin keys into flexible-frontend (where they presumably will live anyhow, eventually?)
> > The exported types are a great shout!
> 
> Makes sense - we could define and export the plugin key in flexible-frontend, but still define the plugin in Composer. I'll give that a try.
> 
> We have a somewhat artificial problem in that defining some code like the plugins in flexible-frontend and importing into Composer will work in the application itself, but will cause the jest-based tests to fail (described in [guardian/flexible-content#6544](https://github.com/guardian/flexible-content/pull/6544) - TLDR I think its because jest doesn't cope with imported code referencing the Document, but haven't dug too deep into that) - we will ultimately solve this by migrating the whole of the editor (including the test coverage, either with jest or something else), but we have a few awkward cases of needing to duplicate code in the meantime.

Ah, that's a pain — happy to pair if more 👀  on the weird Jest problem would be useful.

Would it be worth trimming this PR and altering the description so it exports those useful types? Happy to give that a ✅  once it's done!

dblatcher:
> > > My initial take on this is that we shouldn't be expanding the surface area of the API here if this can be solved in the consuming code. Is it possible to do that — for example, by moving the plugin keys into flexible-frontend (where they presumably will live anyhow, eventually?)
> > > The exported types are a great shout!
> > 
> > 
> > Makes sense - we could define and export the plugin key in flexible-frontend, but still define the plugin in Composer. I'll give that a try.
> > We have a somewhat artificial problem in that defining some code like the plugins in flexible-frontend and importing into Composer will work in the application itself, but will cause the jest-based tests to fail (described in [guardian/flexible-content#6544](https://github.com/guardian/flexible-content/pull/6544) - TLDR I think its because jest doesn't cope with imported code referencing the Document, but haven't dug too deep into that) - we will ultimately solve this by migrating the whole of the editor (including the test coverage, either with jest or something else), but we have a few awkward cases of needing to duplicate code in the meantime.
> 
> Ah, that's a pain — happy to pair if more 👀 on the weird Jest problem would be useful.
> 
> Would it be worth trimming this PR and altering the description so it exports those useful types? Happy to give that a ✅ once it's done!

Exporting the plugin key (but not the plugin) from flexible-frontend worked locally, so I think that's the answer to the immediate Composer problem (I'll be sure to include some comments about why as it would look odd without context). Thanks for the idea!

We might work around the jest problem by wrapping the whole editor up and/or moving the test coverage, but might ask for help if we need to tackle it head on.

I think I'll close this PR and open a new one with just the exports to preserve the decision and not have a merge PR with a misleading branch name in the history -  will tag you when done.

PR #483, opened 2026-05-19
Remove unused and broken ESM support

liamness:
## What does this change?

Follow up to #481, which did not have the benefits I was hoping for! It seems that the ESM build is not valid, because it does not include full file paths (i.e. ending with `.js`) in imports. As the ESM build is not currently used, and is also broken, I think the simplest thing to do is to remove it and also the `"type": "module"` from the package.json (which I think I must have done at some point when previously debugging vitest-related problems).

We could revisit ESM support later if we think the tooling around it is mature enough to justify this. I think it would also make sense to move completely from CJS to ESM if this were the case—as there is only one downstream project using this I don't think there is much justification for supporting both. 

## How to test

I think this might be difficult to test locally. `npm link` would get tricky quickly due to different versions of Node being involved, and the docs for changesets warns that using it for [prereleases is "very complicated"](https://github.com/changesets/changesets/blob/main/docs/prereleases.md). So my suggestion is that we merge and release this, test with Composer, and then fix forward if issues are found.

edit—Alternatives I have tried are `npm install --no-save ../../prosemirror-elements`, `npm link ../../prosemirror-elements` and `npm install ../../prosemirror-elements` (the latter creates a `file:` reference in the `package.json`) and all of these seem to have issues with Jest or Webpack.

Shipping this package in a state less likely to confuse people

There's an unlikely, but non-zero possibility that this will break something in Composer.

### DIFF
--- a/tsconfig-cjs.json
+++ b/tsconfig-cjs.json
@@ -1,8 +1,11 @@
 {
-  "extends": "./tsconfig-esm.json",
+  "extends": "./tsconfig-base.json",
   "compilerOptions": {
     "module": "commonjs",
     "target": "es5",
-    "outDir": "./dist/cjs"
+    "outDir": "./dist/cjs",
+    "sourceMap": true,
+    "declaration": true,
+    "declarationDir": "dist/declaration"
   }
 }
--- a/.changeset/shy-icons-relax.md
+++ b/.changeset/shy-icons-relax.md
@@ -1,0 +1,5 @@
+---
+"@guardian/prosemirror-elements": patch
+---
+
+Remove ESM support (that did not actually work)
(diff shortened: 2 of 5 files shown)


tsconfig-cjs.json
```diff
@@ -1,8 +1,11 @@
...
-  "extends": "./tsconfig-esm.json",
+  "extends": "./tsconfig-base.json",
   "compilerOptions": {
     "module": "commonjs",
     "target": "es5",
-    "outDir": "./dist/cjs"
+    "outDir": "./dist/cjs",
+    "sourceMap": true,
+    "declaration": true,
+    "declarationDir": "dist/declaration"
```
liamness, line 9: This is now the only tsconfig file that is used by the build command, so moved build-related configuration settings here.

davidfurey (approved):
Looks reasonable to me, and the strategy of fixing-forward if necessary seems appropriate.

